PR #7, opened 2016-08-16
Update README.md

omridevk:
Update README.md file with a reference to the playlist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ code or various creative assets, feel free to follow the standard Github workflo
 6. Make a "Pull Request" from your branch here on Github.
 
 For a video tutorial, please see:
+* <https://www.youtube.com/user/quill18creates/playlists>
 
 
 ## Contact


quill18:
Two people will a similar quick fix! Have to decline one of them.

omridevk:
yeah the other one is better :)

omridevk:
@quill18 
Mind if I do some refactoring?

quill18:
@omridevk It would be a pretty crummy open source project if I said no!

omridevk:
Have to say good job and love your videos. Information that is very difficult to acquire online. A job well done indeed.

Don42:
@quill18 Since people often disagree about formatting and style it would probably good to define those before there are conflicting pull requests.
I don't know if there are standard formatting guidelines for C#, but if there are, a link would probably be helpful.

Koljenka:
please dont do it like this: "method() {" do it like this: "method{" just to add to that formatting thing.

omridevk:
This is the C# convention:
https://msdn.microsoft.com/en-us/library/ff926074.aspx
@Koljenka one way or another does not really matter, as long as it is consistent across the entire code base. Choose one way, use it everywhere.

quill18:
Yeah, I use everything in the C# convention (AFAIK)...except that I like to put my brace on the same line as the if.  E.G.:

`if (blah) {`
`    DoSomething();`
`}`
`else {`
`    DoSomethingElse();`
`}`

If the majority of people are religiously opposed to this, I could force myself to change. :)

omridevk:
I would even go as far as saying that you should avoid using 
```
else
```
all together. 

As in  SmallTalk coding guide (only one way to avoid else statements):
http://c2.com/cgi/wiki?GuardClause
and (Martin Fowler refactoring book):
http://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html
Polymorphism is another great way to avoid elseif and switch cases.
Covered in a great Google Clean Code talk:
<https://www.youtube.com/watch?v=4F72VULWFvc>

PR #12, opened 2016-08-16
Added ability to dismiss build mode with right click

RodRitter:
A small change, but hopefully very useful.


NogginBops:
I think this should be done with the unity input system to be able to remap the controls at startup and run time.

RodRitter:
I am actually thinking about working on that now. If we are going to wait until we have a set standard for input, then we might consider adding this as a hotfix for now. I found it useful while debugging and playing around.

TylerVu:
Doge

svmnotn:
To use the unity input system would require a whole separate PR as it takes a bit more and as @RodRitter said this is more of a quick fix

NogginBops:
That is true. Added a issue for the input manager. #13

PR #19, opened 2016-08-16
Returned the circle cursor

kd7uiy:
This feature was lost after the transition to Streaming Assets. I am now specifying it manually, it's a bit hacky, someone feel free to make it better, but as it's the root cause of the problem, I thought I'd add it.


Dormanil:
You might want to change your MonoDevelop/Visual Studio settings for this project, as quill is using tabs that are **not** changed into spaces, while your additions use spaces as indentation.

kd7uiy:
Good call, I always forget silly things like make sure the tab style is consistent... Sigh. Thanks!

PR #23, opened 2016-08-17
Allow certain furniture to replace other furniture - ie. Doors into Walls

RodRitter:
I added the ability to inset furniture in place of other furniture. Details below for a `Door` into `Wall`example

For example, a `Door` needs to be able to replace a `Wall` so in the XML on the Door furniture, I'll add this line.

ie. `<CanReplaceFurniture objectName="Basic Wall" />`

Now validation will allow me to replace a wall block with a door.

**Note**: I apologise for all the unnecessary file changes. My MonoDevelop added auto-formatting to some of the methods.

### DIFF
--- a/Assets/Scripts/Controllers/BuildModeController.cs
+++ b/Assets/Scripts/Controllers/BuildModeController.cs
@@ -80,6 +80,13 @@ public class BuildModeController : MonoBehaviour {
 				t.pendingFurnitureJob == null
 			) {
 				// This tile position is valid for this furniture
+
+				// Check if there is existing furniture in this tile. If so delete it.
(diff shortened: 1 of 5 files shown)


quill18:
Good change! I'm okay with merging this as-is, but I would prefer if instead of instantly deconstructing the wall, the "build door" job would check to see if there's something that needs to be deconstructed first.

Do you want to add this to this pull request, or do we want to save this for later?

quill18:
Actually nevermind -- that would require extensive changes to Character.cs  -- better for us to merge this and then do a separate PR after.  Massive merges are bad -- especially since we're still dealing with code format standardization issues.

RodRitter:
Ah ok great. I'll look into that tomorrow definitely.

PR #28, opened 2016-08-17
Contribution Style Guidelines 

quill18:


SquareMan:
I may be misreading something, but shouldn't "Avoid using 'var', even with the type would be clear from context." say "Avoid using 'var', even **WHEN** the type would be clear from context. Verbose typing is best typing."?

PR #36, opened 2016-08-17
Replace hardcoded O2 parameters with xml versions

bjubes:


RodRitter:
Maybe you should have redundancy checks for those values. You never updated the xml with any values so it might throw an error when you're looking for non-existant values.

bjubes:
Good Point!

svmnotn:
@bjubes you might want to fix the conflicts by updating your branch to the latest version of the master in this repo so that it is easier for @quill18 to merge the PR.

bjubes:
I'm not at a computer right now but I will when I can. Should I merge master in or rebase on top, or does it not matter?

svmnotn:
it doesn't matter too much, but I normally do a rebase since it ends up a tad cleaner. Although that involves a force push (sometimes).

ComputerOverlord:
This was like, the first request. How has Quill not considered it?

bjubes:
@ComputerOverlord  because the line endings change from earlier created a merge conflict which git couldn't automatically resolve. This is very common in git and can happen whenever two people modify the same line in a file. I Wouldn't be surprised if quill eventually makes a "solving merge conflicts" tutorial

quill18:
Unfortunately this is conflicting again -- likely due to the change to room gas level calculations.

bjubes:
@quill18  Yep. incorporated the new pressure system into my changes

quill18:
Thanks!

PR #37, opened 2016-08-17
ModUtils: expose C# and Unity methods to LUA for more extensive modding

bjubes:
Furniture.lua needs to be able to access methods built into unity and c#, from the basic math functions like clamp to more unity centric functions that cannot be replicated in Lua.

Solution: ModUtils exposes methods in c# by encapsulating them in methods within the ModUtils class. These can then be called in Lua with `ModUtils.<MethodName>` and removes the need for a redundant "utilities" section in the Lua files, which was essentially for rewrites of existing c# functions.


Dormanil:
Only problem I see with this is, we have to tell our modders which Unity or C# methods are available for use in Lua because they were encapsulated in ModUtils.

I personally as a modder would not want to search the C# source-code for methods I could use in my Lua scripts, if I just simply want to write Lua.

BeefEX:
There can be some kind of documentation generated from that class. That would make easy.

lameox:
We could also do it both ways and expose those functions in modutil, but also autogenerate the wrapper functions in the lua file that basically just call into the modutils functions. This way the modders could see all available functions, and the c# programmers would be able to freely extend the functionality exposed to modders.

I propse an attribute we decorate the ModUtils Functions with (possibly with some metadata like a short description that appears in the lua wrapper), that allows us to inspect the class with reflection at startup, and generate the appropriate lua wrappers. This way the wrappers would always be in sync with the newest version of the c# code after the user/modder starts the program once. 

We should probably create an issue for that.

svmnotn:
@bjubes you might want to fix the conflicts by updating your branch to the latest version of the master in this repo so that it is easier for @quill18 to merge the PR.

bjubes:
@lameox do u think I should keep this PR for now and we can build that attribute meta data system in a future PR? I'm not great in Lua so I'm not sure if I could build what your thinking of cleanly

lameox:
@bjubes I think you should keep it open. There is no problem in having a PR open for a while. At worst somebody finds it and implements it ;)

bjubes:
okay well i clearly messed up that merge. This commit history is a mess and I don't think we need this feature at the moment so feel free to cherry pick this if you ever want modutils in the future. Closing for now.

PR #39, opened 2016-08-17
Console in game

kishtarn555:
This allow to do Lua instructions
Recive Text and show it in UI
But needs premade Lua Functions in Command.lua
And the information after try a Lua instruction is almost null


ComputerOverlord:
Good idea! I wonder if Thequill will add this!

quill18:
Please add more details about how this actually gets used in practice.

kishtarn555:
It's an UI element and has attach a c# derivate from dialogbox.
There tree functions: Log, LogWarning and LogError.
They recive two objects as arguments:
 First is optional, its who is telling the message, and the second its the message. Both are printed by object.ToString().
The UI also has an InputField, when you press enter and finish the edit take his text and is interpreted 100% as Lua code. You can add premade Lua functions in Commands.lua

kishtarn555:
image

kishtarn555:
I moved the console to prefab because the merge problems

svmnotn:
@kishtarn555 how would someone open/close(maybe not close as I do see the 'X') this? Is there a key binding? Like how paradox games have it in the ` key? or some in game button? does it just show up at the start and can't be opened afterwards?

kishtarn555:
I dont have any open because I wanted to leave to the future because need more stuff like have a Developer mode.
But you can just add a button o key and write this line code
DevConsole.Instance.ShowDialog();

kishtarn555:
I added a button to the Top Menu as temporal way to open it

ComputerOverlord:
Not quite what I expected... well maybe sort of. You should add Commands or at least more commands.

BeefEX:
@ComputerOverlord commands are defined in LUA, read please

ComputerOverlord:
Maybe ~ should open console like in most games?

Tranberry:
If tilde is added as the 'open console' key, will it be remapped to the key that is below (esc) on localized keyboards? On swedish keyboards (§) is the default console key (located below esc).

ComputerOverlord:
@Tranberry I only half-understood that haha. I wonder how the keybind will be set up in the end..

kishtarn555:
Well im not sure what to do...
Because if I add a key to openit This mean i would ad a Keyboard Controller And this part it's not about.
Like an temporally I will addit a father wich listen to a key

ComputerOverlord:
@kishtarn555  yeah, It could be added later..

Tranberry:
@ComputerOverlord 
delete

Do you get it now?

ComputerOverlord:
@Tranberry haha. I was confused about the enter... I get it now...

Mizipzor:
Off topic: You seem to merge a lot. Almost all the time. But thus far it looks like you never had to do it. You can probably save some time and effort. Like only merging after you create the pull request or push updates to it and Github says you have conflicts. What tool do you use btw? If I may ask. And is there a better place to discuss this rather than in this issue? :P

kishtarn555:
I used F12 key

kishtarn555:
Hey can someone tellme the correct translation for console and Type a command in Deustch 
I speak spanish and english

RodRitter:
@kishtarn555 are the language updates necessary for this merge?

Mizipzor:
I definitely think the localization changes should go in a separate pull request.

vogonistic:
Please open a new PR where the merge errors are fixed and where we can use it to trigger lua. Please keep language updates separate.

PR #40, opened 2016-08-17
Checking if saves directory exist before trying to read from it

QuiZr:
Before that DialogBoxLoadSaveGame.cs was spitting an exception when there was no saves folder (i.e. before making first save at fresh copy of the game).


quill18:
Good idea! 

Since DialogBoxSaveGame::SaveWorld also checks/creates the save folder (and duplication is bad), can you make a function in DialogBoxLoadSaveGame like...."EnsureDirectoryExists" or something have have both DialogBoxLoadSaveGame::ShowDialog and DialogBoxSaveGame::SaveWorld call that?

In theory, SaveWorld should never get called before ShowDialog, but just in case I think we should still check in both spots.

QuiZr:
Aaaand file endings.... Damn you git >.>

quill18:
Merged!

PR #41, opened 2016-08-17
Ensure that git sets text files to OS-native EOL

QuiZr:
Without this everyone who contributed could push his own line-endings to repository leading to destroying of the universe... Or some annoying Unity warnings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# https://help.github.com/articles/dealing-with-line-endings/
+# Declare files that will always have LF line endings on checkout.
+*.cs text eol=lf
+*.xml text eol=lf
+*.lua text eol=lf


quill18:
Our contribution guidelines encourage people to set their MonoDevelop code formatting to the Microsoft Visual Studios defaults, which uses "Native" line endings.

Would this not mean that Windows users would expect CRLF and would this proposed change not conflict with that?  (Also, if someone edits files in an editor that isn't flexible when it comes to line-endings, such as someone using Notepad to tweak the README.)

Would "text=auto" not be a better option (which should be the default in most Git installs).  That should auto-convert all files to native line-endings, which should make MonoDevelop happy -- and git should still handle everything as expected during merges.

This is very "AFAIK" -- I certainly haven't checked all situations personally.  Do feel free to clarify if your testing indicates otherwise.

PR #45, opened 2016-08-17
Code refactoring of Character's job handling functions

lameox:
Took the huge function body and divided it into a few logical functions. Also added/reworked/deleted some comments where appropriate and layed some foundadtions for further improvements to the job queue, i.e. Dumping of Inventories. I will also open an issue discussing the inventory-dumping further.

### DIFF
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -29,7 +29,7 @@ public class Character : IXmlSerializable, ISelectable
         {
             if (NextTile == null)
                 return CurrTile.X;
-			
+
             return Mathf.Lerp(CurrTile.X, NextTile.X, movementPercentage);
         }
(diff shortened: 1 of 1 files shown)


lameox:
Wrote up a proposal for future inventory handling system. See #47

quill18:
Excellent! That was definitely very high on my wishlist.  We could probably break things down even further, but at least it's not "literally the worst ever" anymore!

lameox:
@quill18  Well we are currently discussing some ways to address the current job code further in #63 and i also wrote a rather large proposal in #47 on how to proceed. However i did not want to change too much in a single pull request but rather wait for feedback. We should maybe also ask you, how true you want to stay to the Code shown on youtube, especially if we rewrite a rather large part like the job system. Thoughts?

Issue #43, opened 2016-08-17
Pathfinding issues


QuiZr:
I've found some issues with pathfinding and jobs systems (i.e. https://youtu.be/AIhuLfB-9F0 and https://youtu.be/W_M4i7eHz6w?t=32m36s) - will try to document and fix that soon.

emilkris33:
I think this is caused be a fairly fundamental flaw in the way stockpiles creates jobs. Currently every non-full stockpile tile creates a job asking for items. That is definitely not optimal. I think a more intelligent system for creating stockpile-jobs are necessary to fix the root of this problem.

QuiZr:
You are right - someone more experienced should rewrite this. For now I've fixed that weirdness that i was talking about.

PR #46, opened 2016-08-17
Fixed pathfinding issues

QuiZr:
#43 - Fixed end tile on old destination stockpile check


QuiZr:
# NOT SAFE TO MERGE
After merge from upstream there is a infinite loop somewhere but I can't work out where

Issue #49, opened 2016-08-17
Feature set: Building the base. (Constructing floor, needing rooms(air), airlocks)


emilkris33:
This is is an outline of what i view is the set of features that is logical for us to move development towards right now and a road map for how i think we should go about getting there. This should all be uncontroversial stuff i line with what @quill18 has already talked about in videos. All feedback is greatly appreciated so we can all move towards a common vision of at least the next few steps in development.

## Features to aim for
- Everything is constructed by a character (no magical floor)

- An enclosed with an atmosphere is required for normal functioning of most things

- Characters can go through an airlock to access outside areas (for construction new rooms)

- Game starts with a bare minimum construction (One small room  and airlock?) and some materials

- System for generating, maintain and requiring an atmosphere(For now just oxygen)

## Roadmap
What needs to be implemented in the engine to accomplish this in rough order that it has to be done. Some things might need to be broken up into several steps.

- [x] Character able to walk on empty tiles

- [x] Character build floor incl. job creation for floor building

- [ ] Airlocks, (different door from internal doors or a small room with two doors that we frequently depressurise?)

- [ ] Atmosphere requirement for some buildings

- [ ] Prevent deconstruction of outer walls and floor in a pressurised zone

- [ ] Path-finding logic to handle differentiating between inside and outside (Only exit the airlock when required)

- [ ] Job logic to handle bringing material for multiple constructions (Don't bring stuff for only one floor tile on each trip outside.

- [ ] Generate initial construction
- [ ] Expand atmosphere system. (Flow air between rooms)

TheRiven:
As a novice trying to work out what to do to help out, an approved task/roadmap list would definitely be helpful. I imagine I am not the only one scratching his head watching stuff popping up having never been involved in a project like this. (or any coding project for that matter!  :confused: )

I have not done any serious programming until Quills tutorials, only dabbled, and so I am lacking in any 'serious' experience. I have now gotten to the stage, thanks to Quill, that I can more or less work my way through most problems (if a bit slowly) and even see some possible solutions as Quill was running through a video (an awesome feeling!).

So any directions/suggestions on where best to contribute would be most appreciated! :smile:  

As a side note to contribute to your roadmap : do we also want to think about power systems for buildings? or is that to soon?

lameox:
Hey @TheRiven, I think one of the easier parts to work on will be the walking on empty tiles. 

[...]

PR #51, opened 2016-08-17
Floor is build by character

emilkris33:
Floor is now manually build by the character through the job system.


emilkris33:
Related to #49

quill18:
Seems to work well!  Will look even better when we add in the ghostly "pending job" image, add a better drag graphic, and maybe make people work jobs from adjacent tiles.  Merging!

PR #56, opened 2016-08-17
All Furnitures with linksToNeighbour are now properly deconstructed and their Jobs are canceled

BeefEX:
Added code that updates all tiles with same furniture around
deconstructed furniture.

### DIFF
--- a/Assets/Scripts/Models/Tile.cs
+++ b/Assets/Scripts/Models/Tile.cs
@@ -133,7 +133,7 @@ public class Tile :IXmlSerializable, ISelectable
             return false;
 
         Furniture f = furniture;
-
+        f.CancelJobs ();
         for (int x_off = X; x_off < (X + f.Width); x_off++)
         {
             for (int y_off = Y; y_off < (Y + f.Height); y_off++)
--- a/Assets/Scripts/Models/Tile.cs
+++ b/Assets/Scripts/Models/Tile.cs
@@ -123,6 +123,8 @@ public class Tile :IXmlSerializable, ISelectable
         cbTileChanged -= callback;
     }
 
+    private static int[][] offsets = new int[][] { new int[] { 1, 0 }, new int[] { 0, 1 }, new int[] { -1, 0 }, new int[] { 0, -1 } };
+
     public bool UnplaceFurniture()
     {
         // Just uninstalling.  FIXME:  What if we have a multi-tile furniture?
@@ -133,7 +135,9 @@ public class Tile :IXmlSerializable, ISelectable
             return false;
 
         Furniture f = furniture;
-        f.CancelJobs ();
+
+        f.CancelJobs (); // Canceles all jobs by this furniture.
+
         for (int x_off = X; x_off < (X + f.Width); x_off++)
         {
             for (int y_off = Y; y_off < (Y + f.Height); y_off++)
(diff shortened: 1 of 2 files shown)


BeefEX:
Sorry for messy comitts, i was messing with github :smile:

lameox:
You should probably squash your last 3 commits into one, so they dont pollute the history. Alternatively you can revert your branch by two commits which serves the same purpose.

BeefEX:
I dont know what is going on, i have huge mess in my github. How can i squash them?

quill18:
@lameox I have forced squashing for the actual merge into the main project -- so we shouldn't have to worry about things getting too messy.

BeefEX:
@quill18 thats nice for the noobies to git as i am :D thx

quill18:
@BeefEX Heya!  Can you add in a few comments please?  Just at the start of each major block should be sufficient.

BeefEX:
@quill18 why not wait a minute

BeefEX:
@quill18 done, and fixed the code too, but my commits are very messy :/

Assets/Scripts/Models/Tile.cs
```diff
@@ -124,25 +124,51 @@ public void UnregisterTileTypeChangedCallback(Action<Tile> callback)
...
+        f.CancelJobs (); // Canceles all jobs by this furniture.
+
         for (int x_off = X; x_off < (X + f.Width); x_off++)
         {
             for (int y_off = Y; y_off < (Y + f.Height); y_off++)
             {
 
                 Tile t = World.current.GetTileAt(x_off, y_off);
+                if (t.furniture.linksToNeighbour)
+                    affectedTiles.Add (t);
```
lameox: I think you can discard the affectedTiles list  and just call t.cbOnChanged(t.furniture) right here

BeefEX: there is another way let me try it

BeefEX:
i have some problems with commiting right now, but here is the final code:
`for (int x_off = X; x_off < (X + f.Width); x_off++)
        {
            for (int y_off = Y; y_off < (Y + f.Height); y_off++)
            {

                Tile t = World.current.GetTileAt(x_off, y_off);
				if (t.furniture.linksToNeighbour)
				{
					//This will update all neighbour tiles.
					foreach (int[] offset in offsets) // Using hardcoded offsets. Is there better way?
					{ 
						Tile tmpTile = World.current.GetTileAt (t.X + offset[0], t.Y + offset[1]);
						if (tmpTile.furniture != null && tmpTile.furniture.objectType == f.objectType)
						{
							tmpTile.furniture.cbOnChanged (tmpTile.furniture);
						}
					}
				}
				t.furniture = null;
            }
        }`

rfmsimoes:
Hello,

@BeefEX  in my opion I think you don't need (and probably it is not the best place) execute the function CancelJobs() inside the UnplaceFurniture function in Tile.cs but what you need to do is go to Furniture.cs and in the fucntion Deconstruct() before call the tile.UnplaceFurniture() call the CancelJobs fuctions.  

Because when you deconstruct furnitures, after the click or dragging the mouseController will call the tile.furniture.Deconstruct()

=)

BeefEX:
@rfmsimoes will update it when i get home thx for pointing that out!

BeefEX:
PR #92 is better. Closing this one.

PR #69, opened 2016-08-17
Fix the circle cursor so that its displayed when dragging floor

yonner:
Fix the circle cursor so that its displayed when dragging floor and did
a little bit of refacting of UpdateDrag() method.


quill18:
Sorry, another fix already added the circle cursor back in while dragging out the floor.

yonner:
I have seen the fix now, I think my fix was slightly better as the sprite it only set on start where the current fix sets the sprite on every update?

PR #74, opened 2016-08-17
Bugfix nullreference

JakubSzczerbinski:
Fixed problems with hovering above no tiles, that caused NullReferenceExceptions.


JakubSzczerbinski:
Sorry for the mess switched from basic git usage on command line to UI.

Assets/Scripts/UI/MouseOverFurnitureTypeText.cs
```diff
@@ -31,19 +31,26 @@ void Start()
...
         }
     }
-	
+
     // Update is called once per frame
     void Update()
     {
         Tile t = mouseController.GetMouseOverTile();
 
-        string s = "NULL";
```
AE40: Becouse Of This Edit, It Will Not Display Anything Instead Of The Intended "NULL"

Assets/Scripts/UI/MouseOverFurnitureTypeText.cs
```diff
@@ -31,19 +31,26 @@ void Start()
...
+                s = t.furniture.Name;
+            }
 
-        if (t.furniture != null)
+            myText.text = "Furniture: " + s;
+        }
+        else
         {
-            s = t.furniture.Name;
+            myText.text = "No Furniture";
```
AE40: Maby "No Tile" Instead?

JakubSzczerbinski: The function should return furniture type in text form. In that case i think "No Furniture" is more suitable. Its not that important anyway, becouse situation, where you hold mouse over no Tile rarely happens.

quill18:
This branch has conflicts that must be resolved.

Assets/Scripts/UI/MouseOverFurnitureTypeText.cs
```diff
@@ -31,19 +31,26 @@ void Start()
...
+        {
+            string s = "NULL";
+
+            if (t.furniture != null)
+            {
+                s = t.furniture.Name;
+            }
 
-        if (t.furniture != null)
+            myText.text = "Furniture: " + s;
```
OzieGamma: *I am not sure whether it works on Unity. This is just a suggestion*

```
if(t != null)
{
    myText.text = "Furniture: " + (t.furniture.Name ?? "NULL");
}
```

It does exactly the same thing than what you wrote but is shorter.
https://msdn.microsoft.com/en-us/library/ms173224.aspx

Assets/Scripts/UI/MouseOverRoomIndexText.cs
```diff
@@ -37,13 +37,20 @@ void Update()
...
+        {
+            string roomID = "N/A";
+
+            if (t.room != null)
+            {
+                roomID = t.room.ID.ToString();
+            }
 
-        if (t.room != null)
+            myText.text = "Room Index: " + roomID;
```
OzieGamma: Same here you can easily use `??`

vogonistic:
If this is still an issue, can you update so it merges cleanly?

Mizipzor:
I just want to add that what @OzieGamma is talking about is called the [null coalescing operator](https://msdn.microsoft.com/en-us/library/ms173224.aspx) and it super awesome, greatly increases readability.

vogonistic:
This has been fixed in other PRs.

PR #76, opened 2016-08-17
Update graphics on furniture deconstruction

SquareMan:
Graphics haven't been being updated when furniture gets deconstructed. I've made it so when a piece of furniture gets deconstructed it looks it if links to neighbors and if so call the OnChangedCallback for all its neighbors.


ComputerOverlord:
Something that I'm sure'll be merged!

BeefEX:
Already fixed in #56 . I suggest closing this one.

SquareMan:
I'm now closing this because #92 is a much better implementation that will work better for generic cases.

PR #77, opened 2016-08-17
Improving the atmosphere system

emilkris33:
The atmosphere system now considers and produces gas in terms of absolute amount. Meaning that it takes twice as long to fill a room with double the size. And it handles gas correctly when splitting and merging rooms.

This is inline with the goals outlined in #49


ComputerOverlord:
I thought this was a cool idea when Quill programmed the atmosSystem!
votes ++=

quill18:
Accepted!

PR #78, opened 2016-08-17
Basic power system

TheRiven:
Ok, Novice Riven reporting in!
I have created a very basic power system (that will most definitely need a rework, refactor and all that) BUT it works! 

This system allows furniture that have the power generator tag in the furnitures.xml to provide power to the currently invisible base power network. 
Furnitures that have a power tag will also attempt to register for a power supply so that they can run their update function. If no power is available then they will not run their update.

At the moment there is no visible feedback on this but i would guess a sprite or something to show its offline would be used eventually.

As this is my first contribution feedback is definitely appreciated!

EDIT: I forgot to mention that I added both a generator with the power generation tag on it and changed the oxygen generator to use power. I felt that was enough to test the system without interfering with anything else for the time being.


TheRiven:
Oh it also contains a quick sprite I made for the power gen, its not great but it does the job for now

ComputerOverlord:
I really like this! Maybe power cables should be added in?

emilkris33:
Looks very good. Though I do see a couple of problems with the system. 

Right now you are denying updates to objects that doesn't have power instead you should call the update telling it that is doesn't have power or add the possibility for a different update for un-powered frames. Why? A light might need to turn of if not powered. A door might be unable to open or close by it self, but might be able to be forced be a character i un-powered or might still leak air. And so on, objects might still have some update behaviour when not powered.

Also you should multiply power consumption and generation by deltaTime

ComputerOverlord:
This should require a resource 'Copper' to make.

TheRiven:
@ComputerOverlord - Thanks and yeah I was thinking in the future it should be possible to build a cable system somehow.

@emilkris33 - Yeah the update check was a pretty rough but quick way of getting furniture to do the check. As you say, it would be better to have it use the fact it has no power to determine the behaviour rather then terminate it altogether. That's what I will probably look at next but for now at least it shows that "powerless" items can be stopped which was the first step.

Could you clarify what you mean by having the power multiplied by deltaTime? Would that not just build up power overtime like a battery? My intention was that the generator would produce a constant amount of power when running and this would probably later be stored in some sort of capacitor. Unless I am getting the wrong end of the stick?

Assets/Scripts/Models/PowerSystem.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        foreach (Furniture furn in powerConsumers)
+        {
+            powerValues -= furn.powerValue;
+        }
+
+        currentPower = powerValues;
+
+        Debug.Log("Current Power level: " + currentPower);
```
alexisfcote: Maybe remove this log?

emilkris33:
@TheRiven I see now never mind about the deltaTime thing. I was thinking that power plant produced to a battery and consumers then drained the battery. But i think your implementation makes more sense.

quill18:
Accepted -- can't wait to see what we do with this!

TheRiven:
Awesome, I wake up to find my first contribution has been deemed worthy! :smiley: 

Now a quick Git question, do I now close this branch and merge everything back into "my" master branch and then create a new branch for future work on the power system, or should I continue using this branch? 

My guess would be the former?

emilkris33:
@TheRiven You delete the branch you did this on. Then you update your master from TeamPorcupine:master, which now has your changes. And finally you create a new branch for your next awesome addition.

TheRiven:
@emilkris33 Ok thanks for the help!

PR #83, opened 2016-08-17
BuildingJobs for every furniture.

SanteriHetekivi:
**BuildingJobs for every furniture so their furniture job prototypes get initialized.**

I noticed that every time I place any other furniture than Door or Wall I get an error like:
`There is no furniture job prototype for 'Mining Drone Station'.`

Reason for that was missing BuildingJob tags for other furniture.

So I added the BuildingJob tags with jobTime 0 in those furniture and now the error is gone.

### DIFF
--- a/Assets/StreamingAssets/Data/Furniture.xml
+++ b/Assets/StreamingAssets/Data/Furniture.xml
@@ -57,6 +57,9 @@
         <MovementCost>1</MovementCost>
         <LinksToNeighbours>true</LinksToNeighbours>
 
+        <BuildingJob jobTime="0">
+        </BuildingJob>
+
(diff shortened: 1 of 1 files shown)


NogginBops:
@lameox We have a solution for that in the works, see #86

SanteriHetekivi:
Anything that makes modder's life easier is great because it makes the game more appealing to the modding community and keeps the game alive longer after the main development has ended.

This is just quick fix for the current error. :)

quill18:
@lameox Technically we don't require a building job as-is.  Arguably, the warning in the log could be removed.  It's just there so that we get reminded that we didn't define a job for that piece of furniture and the default dummy job gets used.

PR #84, opened 2016-08-17
Fixed Null Reference Exception When Not Hovering Over A Tile

AE40:
Simple And Propably Without Point As We Will Propably Fix This By Making So You Cant Zoom Out Of The Tiled Area, But A Simple Quick Fix.


AE40:
Closing Due To Duplicate

PR #90, opened 2016-08-17
Fixing doors part 1: Update surrounding furniture when create and removing furniture

Raketfart:
When placing or removing furniture, all surrounding furniture gets
updated. Regardless if they are same type or connects to neighbours.
This is first step in fixing doors not connecting correctly to walls if door is placed before walls.


Raketfart:
Missing a file, will try again

PR #92, opened 2016-08-17
Fixing doors: Update surrounding furniture when create and removing furniture

Raketfart:
When placing or removing furniture, all surrounding furniture gets
updated. Regardless if they are same type or connects to neighbours.
This also works for furniture larger than 1x1.

So when placing a wall north/south of a door, will now make the door rotate correctly.
Still not a perfect solution for rotating furniture like this.

### DIFF
--- a/Assets/Scripts/Controllers/FurnitureSpriteController.cs
+++ b/Assets/Scripts/Controllers/FurnitureSpriteController.cs
@@ -120,11 +120,18 @@ public class FurnitureSpriteController : MonoBehaviour
 
             Tile northTile = world.GetTileAt(furn.tile.X, furn.tile.Y + 1);
             Tile southTile = world.GetTileAt(furn.tile.X, furn.tile.Y - 1);
+            Tile eastTile = world.GetTileAt(furn.tile.X+1, furn.tile.Y);
+            Tile westTile = world.GetTileAt(furn.tile.X-1, furn.tile.Y);
 
             if (northTile != null && southTile != null && northTile.furniture != null && southTile.furniture != null &&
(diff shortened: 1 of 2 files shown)


BeefEX:
Please call cancelJobs() on deconstructed furniture as well. This is much better than mine.

Raketfart:
Thanks, I added cancelJobs along with a check for furniture null (not sure if that's necessary)

rfmsimoes:
Hi,

Your change in FurnitureSpriteController is ulesse ,because the IF you paste is executed in here:
          -> furn_go.GetComponent<SpriteRenderer>().sprite = GetSpriteForFurniture(furn);

The GetSpriteForFurnite will do that IF !!

In Furniture.cs I think you shouldn't remove the validation  if (obj.linksToNeighbour) because is if you have one furniture that doesn't have same behaviour like wall...you don't need to check the neighbours, for example 2 chairs....they don't link each other because they are distinct furnitures

Raketfart:
I guess the linksToNeighbour check can stay in, because it's the wall that is placed last. 
I agree that the furn.objectType == "Door" is a temporary fix. We'll have to see how we fix rotatable furniture like doors properly.

BeefEX:
@Raketfart I wanted you to add it because it was the other change i did in my fix, and this is better so i will close mine, but then cancelJobs() fix wouldnt be done at all :smile:

Raketfart:
@BeefEX Yes, I understand, and thanks for the comment. I call cancelJobs() in Deconstruct(), so I hope that's alright.

Raketfart:
Weird conflict. I'll try to add a new PR

Issue #94, opened 2016-08-17
Log Spam and Practices


bjubes:
While it is very much encouraged to spam the log to your hearts content while testing, there should be some general guidelines on what debug messages should be allowed to be kept in a final PR. I am currently removing the more spammy ones to prevent lag and to find my own messages in the mix, but there should definitely be some written rules about this in either the CONTRIBUTING file or somewhere more fitting.

NogginBops:
I agree we need some kind of standard for this!

emilkris33:
I general i think that all debug statements that has a chance of running fine should be commented out or deleted. And you should only leave debug statements that definitively indicate something has gone wrong in.

bjubes:
Agree @emilkris33. Maybe in the future we could have a custom log with log channels that you could toggle on and off, so that if i'm working with the job system i can turn the job channel on and get only the relevant debug info i need. However that would be beyond impractical now, so a "delete when you don't need it anymore" policy the easiest solution IMO

emilkris33:
@bjubes "delete when you don't need it anymore" and more importantly "delete before making a pull request". You might still need it for working on your next thing, but that does not mean everybody else needs it when working on their thing.

lameox:
I suggest we facilitate some Compiler Flags. Something like a 
'''
#if VerboseOutput
            Debug.Log("Some Log Message");
#endif
'''
snippet would be great. This way we could just turn those Debug Messages off and on with a compilation Flag. It might also be useful to define our own dedicated logging class that is used instead of Debug.Log. That way we can easily extend our logging mechanisms in the Future. For now it could just mirror the Debug class' Logging methods and call them, wrapped in the above mentioned compiler flag, but in the future we might want to log to a file, sqliteDB or something else along those lines.

Mizipzor:
A much better way to do that is to create our own logger class that wraps the unity logging. You already log to a "level" (error, warning, verbose), so we could add a setting and only show that level and above. Sometimes you need to readd the verbose logging when you debug something which is difficult if the lines have been removed. Adding a setting scales much better.

lameox:
We should keep the levels for sure. But i think in a release build we should not have any calls to Debug.Log at all. This is best accomplished with a Preprocessor flag around our Debug.Log call(s).

In doing so we will also have a slightly smaller, slightly more performant binary as the Debug.Log calls wont even be compiled.

Mizipzor:
[...]

PR #95, opened 2016-08-17
Remove log spam

bjubes:
Some leftover debug code from testing features is filling the debug log with hundreds of messages. This addresses #94

### DIFF
--- a/Assets/Scripts/Controllers/MouseController.cs
+++ b/Assets/Scripts/Controllers/MouseController.cs
@@ -165,7 +165,6 @@ public class MouseController : MonoBehaviour
                     mySelection.subSelection = (mySelection.subSelection + 1) % mySelection.stuffInTile.Length;
                 } while(mySelection.stuffInTile[mySelection.subSelection] == null);
             }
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Controllers/MouseController.cs
```diff
@@ -165,7 +165,6 @@ void UpdateSelection()
                     mySelection.subSelection = (mySelection.subSelection + 1) % mySelection.stuffInTile.Length;
                 } while(mySelection.stuffInTile[mySelection.subSelection] == null);
             }
-            Debug.Log(mySelection.subSelection);
```
emilkris33: Maybe commenting the statement out would be better so that it is ready for the next person who has to work with this code. Then later on we can clean out debug statements that are commented out i necessary.

bjubes:
@emilkris33 I agree with your point of commenting since its less destructive. However im 90% sure that line was from when quill was creating the selection code back in ep 67, so its not really needed anymore.

The other log message I removed was in a job called "waiting", so the debug call that repeated this information is not really necessary anymore. It also called 2 - 10 times a second.

PR #97, opened 2016-08-17
Use proper events

NogginBops:
This branch has changed all the callbacks to proper events.


bjubes:
I'm confused as to why you deleted the `Register_____Changed(action<T>)` and `Unregister_____Changed(action<T>)` . I was under the impression that those were supposed to be public encapsulations of `on___Callback `events which would eventually be made private.

QuiZr:
@bjubes You don't need to wrap them in seperate functions because event keyword prevents classes not inhareted from the class that this event is declared to remove/call that event.

vogonistic:
I'm curious what @quill18 thinks of this.

This PR seems to make it follow a more recognizable C# pattern, but Quill seems to prefer obviousness over programming patterns. I think to someone who is new to programming, `world.RegisterCharacterCreated(OnCharacterCreated)` is probably more obvious how and what it does than `world.cbCharacterCreated += OnCharacterCreated`, but only marginally.

@NogginBops Correct me if I'm wrong, but my limited understanding of `event` is that it causes calling the callbacks to be `protected`, so the privacy model is unchanged.

I have no opinion either way, just curious as I'm trying to understand Quills style.

NogginBops:
@vogonistic From the msdn event page:
"From outside the class that declared it, an event looks like a field, but access to that field is very restricted. The only things that can be done are:
- Compose a new delegate onto that field.
- Remove a delegate from a (possibly composite) field.
"

quill18:
For me, the only question is whether or not the Register/Unregister functions will ever need to do any other bookkeeping & cleanup. I suspect the answer is no and that this is indeed a big win for cleaner code.

Approved.

PR #98, opened 2016-08-17
Placing multi tile items by their middle or workspot

emilkris33:
Mouse is now bound to the middle of multi tile instead of bottom left corner as before.
EDIT: Mouse is bound to job spot when ever that has been defined.

This changes only the mouseController, affects all multitile furniture and only multitile furniture.  

No mechanical changes only user interface. Bottom left corner is stil used internally for many things and when character constucts building.


NogginBops:
@emilkris33 I think this might add uncertainty/unaccuracy to where exactly the furniture is going to be placed. I would suggest the upper right corner.

emilkris33:
@NogginBops Since it is only for furniture placing you always have a good preview which should eliminate any confusion.

Also i don't see how this could be done internally, unless you are talking about switching from lower left to upper right. In that case though i would argue that the internal implementation does not matter and we should instead make sure the player doesn't experience that the object exists more in one tile than in the others.

NogginBops:
@emilkris33 I can understand what you mean, but what I would suggest then is to position the furniture by work offset (work spot). This would add some kind of useful gameplay hint to the player if all multi tile furniture is placed this way.

quill18:
I agree that work offset is probably the ideal spot -- especially once object rotation is in.

emilkris33:
@NogginBops @quill18 I have now modified the code so that it uses the job offset for objects that has it defined. It instantly fells a lot more natural to place the Oxygen Generator.

PR #107, opened 2016-08-17
Add preview for building floors

SquareMan:
Since floors require the player to build them I decided to add in the preview sprite. Unfortunately since the job system isn't fleshed out yet, the approach is a bit hackish. However, it works perfectly fine and can easily be updated once the job system is redone (specifically the job sprite system)

### DIFF
--- a/Assets/Scripts/Controllers/JobSpriteController.cs
+++ b/Assets/Scripts/Controllers/JobSpriteController.cs
@@ -24,7 +24,7 @@ public class JobSpriteController : MonoBehaviour
     void OnJobCreated(Job job)
     {
 
-        if (job.jobObjectType == null)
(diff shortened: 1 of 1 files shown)


vogonistic:
Can you add a screenshot?

SquareMan:
It's just normal "greenified" version of the normal sprite. Give me a minute and I'll get a screenshot.

SquareMan:
floor sprite preview

Tranberry:
Might be a number of different reasons (nightmode, laptop screen etc..) but I can't see any difference.

vogonistic:
I assume it's more visible while running it locally. Definitely agree that it's better than no sprite at all. 👍

PR #116, opened 2016-08-18
Added double Click to Load and Save Dialog

alexisfcote:
Invoke the OnClick of the displayed okay button when double clicking on a item in the list


svmnotn:
@alexisfcote you might want to fix the conflicts

quill18:
It would likely be better if you found the button through means other than a tag. Your good, but very minor feature is causing more files than necessary to be changed -- and the changes are undesirable since it modifies the Scene (which should be avoided when possible) and adds clutter to the global list of tags.  Additionally, there could potentially be more than one active GameObject with the "OkayButton" tag in the future, which could cause your code to fail if it finds the wrong one.

PR #119, opened 2016-08-18
Ignoring .db files

frankitox16:
For some bizarre reason, Unity was trying to load "thumbs.db" file like
it was an image... added this bit to just ignore .db files (guess we
won't use them anyway)


svmnotn:
@frankitox16 you might want to fix the confilcts

quill18:
This branch has a conflict and it also appears that your whitespace rules differ. Please check CONTRIBUTING.md

PR #122, opened 2016-08-18
Hotfix - Cancelling build mode while dragging would not reset drag mode

RodRitter:
This bug didn't reset isDragging to false when going out of build mode, so the result would be that when selecting another furniture, it would begin dragging it before you even clicked.


SquareMan:
Nice fix. I noticed this earlier but couldn't figure out the exact process to reproduce it.

quill18:
Nice!

PR #124, opened 2016-08-18
create first es_ES translation

svmnotn:
Currently Mining Drone is not translated because I can not come up with an adequate translation for it. @quill18 I would like to know if you would prefer translations to be direct or based on the meaning?


ComputerOverlord:
I think based on meaning.

vogonistic:
@svmnotn I would create a translation that you think might be a good value and then if you or someone else figures out something better, there can be another PR. It's better to get an initial value than a missing one.

svmnotn:
I think based on meaning is definitely the way to go as it leads to less strange translations but I wanted @quill18's input so as to make that the "official" strategy.

frankitox16:
@svmnotn Maybe "Estación de minería robótica" for the mining drone station? drones are robots after all :) (although it seems a bit long, I guess).

quill18:
Translations based on meaning (i.e. what actually looks most sensible in the language) is probably the way to go.

svmnotn:
@frankitox16 I totally missed that link... Thanks I'll make a quick fix now

PR #125, opened 2016-08-18
Quick fix for localization

relinu:
Quick fix for localization probelms. Added format values after key.


relinu:
screenshot03

relinu:
Also added LocalizationLoader to Controllers object

vogonistic:
You should probably remove the font files from this PR.

PR #127, opened 2016-08-18
Localization quick fix

relinu:
Adds localizationLoader to controllers-object and one additionalValue to the key.

PR #134, opened 2016-08-18
Fixing issues with Localization

UnknownSoldier1:
In my repo it seems that the scene was missing a LocalizationLoader, with meant that none of the localizations were working as intended. Fixed that.


UnknownSoldier1:
Closing Pull Request as it's fixed in #127

PR #135, opened 2016-08-18
Added Damage system

BeefEX:
This is first implementation of Damage system using interface, characters and furniture can be killed (destroyed). Added way to kill character. It Destroyes character GameObject cancles his Job and dump his Inventory. Alse updated GetHitPointString() of Furniture and Character to work with this system.

### DIFF
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -123,6 +123,7 @@ public class Character : IXmlSerializable, ISelectable, IDamageable
         CurrTile = DestTile = NextTile = tile;
 		this.maxHitpoints = 100;
 		this.hitpoints = maxHitpoints;
+		World.current.cbCharacterRemoved += onCharacterRemoved;
     }
 
 
@@ -447,6 +448,12 @@ public class Character : IXmlSerializable, ISelectable, IDamageable
         if (cbCharacterChanged != null)
             cbCharacterChanged(this);
     }
+
+	void onCharacterRemoved (Character c)
+	{
+		this.AbandonJob ();
+		this.DumpExcessInventory ();
+	}
     
     void OnJobStopped(Job j)
     {
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -16,7 +16,7 @@ using System.Xml.Serialization;
 /// Later, the Character class will likely be refactored (possibly with
 /// sub-classes or interfaces) to support friendly workers, enemies, etc...
 /// </summary>
-public class Character : IXmlSerializable, ISelectable
+public class Character : IXmlSerializable, ISelectable, IDamageable
 {
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


BeefEX:
There is no way to deal demage for now, this system is here just to be prepared.

BeefEX:
Also here is video of how furniture is removed: https://youtu.be/Lk9asECy-94
This was just to test, dealing 1 demage each frame.

ComputerOverlord:
Sounds _**GREAT**_

emilkris33:
Great addition. A suggestion: You have a lot of code that is copied between the different files. That is the kind of thing that should be defined in the interface. All the functions:
```
void takeDemage (int amount);
bool heal (int amount);
int getHitPoints ();
getMaximumHitPoints ();
```
Could exist entirely or mostly in the interface file.

BeefEX:
@emilkris33 they are in interface

emilkris33:
@BeefEX The interface declare that they must exist, but you can also implement the functionality directly in the interface file so you don't have to copy paste. CopyPasted code is always a bad sign.

BeefEX:
@emilkris33 i am used to Java, so when i saw there is no default i wasnt even trying to find other way :D will look into it thanks

BeefEX:
@emilkris33 can you please point me some resources about it, only way to do this is extension methods, is that what you meant?

emilkris33:
@BeefEX Sorry i thought that should be possible but looking into it i can find the way. But i does seem there should be a better way to include the exact same code in multiple classes using an interface, without copypaste.

Assets/Scripts/Models/Character.cs
```diff
@@ -496,10 +504,56 @@ public string GetDescription()
...
+			return true;
+		}
+		else
+		{
+			this.hitpoints += amount;
+			return true;
+		}
+	}
+
+	public int getHitPoints ()
```
OzieGamma: Why not use a property for this ?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -683,10 +698,56 @@ public string GetDescription()
...
+			return true;
+		}
+		else
+		{
+			this.hitpoints += amount;
+			return true;
+		}
+	}
+
+	public int getHitPoints ()
```
OzieGamma: Try using properties:

https://msdn.microsoft.com/en-us/library/x9fsa0sw.aspx

BeefEX: i like it more with the user friendly function names if you dont mind

vogonistic:
Can you update this PR so it merges? I also spotted some tabs that needs to be converted to spaces. There has been a lot of updates to the CONTRIBUTING.md file the last days, so make sure you look through it for MonoDevelop settings. Thanks!

BeefEX:
@vogonistic i use VS but i had wrong setting. I will rewrite this to work with newest verion of master

vogonistic:
Make sure you add something simple that can cause damage as well, even if it's something simple.

vogonistic:
Are you planning on creating a new pr or keeping this?

FilippoLeon:
@BeefEX I was thinking about implementing a comprehensive health system, where each character has attached an health class, with health conditions, possibility of treatment, and more.

What do you think? This could be a very heavy task, with xml and lua configuration. And could be developed separately from the character class.

vogonistic:
I'm closing this PR for now. @FilippoLeon and @BeefEX please coordinate on creating a health system.

PR #137, opened 2016-08-18
Added check to stop deconstruction of furniture if it links to the outside

mattpott:


Raketfart:
Why should this not be allowed?
When constructing an addition to the base, new rooms could be temporary, while working on new parts of the base. If you need to make a room smaller and cut off a part of your base, this wouldn't be possible.
Why not just let the player "mess up", and the room looses oxygen if it connects to the outside?

mattpott:
It is not that you wouldn't be able to de-construct outside walls but it would require a procedure to try and dismantle outside walls. i.e de-pressurise the room first so as not to get explosive decompression.
It could be argued that this is a bit too much hand holding for the player.

emilkris33:
@mattpott I think this is something for a more mature pressure system

mattpott:
@emilkris33 Agreed. I have changed it now to just spit out an debug message instead of just preventing the deconstruction

PR #139, opened 2016-08-18
Optimising pathfinding

### LINKED ISSUE #64: Game is slowed significantly be current path-finding implementation
As it is currently the game has to generate the entire A*graph every time a floor tile or a piece of furniture is constructed. On my computer this causes very visible slowdown. As shown below my computer uses almost 3 full seconds regenerating the graph.
image

I am not really sure how to go about improving this so any suggestions would be greatly appreciated

emilkris33:
Regenerates the pathfinding graph around the point that has been modified rather than deleting the entire graph. Makes pathfinding after having constructed take around 40 ms rather than 3000 ms on my machine.

Fixes #64


Mizipzor:
Why so many merges? You should rebase before you push or at the very least only merge once. Makes for much cleaner pull requests. :)

Tranberry:
@Mizipzor my guess is that like me emilkris23 is new to git. So maybe tell him/her how to do that with the GitHub Client and shell would be a nice gesture. Now I'm off to reading up on what rebasing is.

emilkris33:
@Mizipzor If you can tell me how to clean this up it would be very welcome. I seem to generate an empty commit when pulling from TeamPorcupine/Master

Though for now it is not really a problem as quill merges the commits then accepting the PR.

Mizipzor:
@Tranberry Good idea, rebase could actually be the solution here as well. :)

Mizipzor:
@emilkris33 I'm nitpicking because the point of this project is, as far as I'm concerned, to learn. Not just C# and Unity but also Git. Without a real world application Git is unusually awkward to learn. Here we have an actual specific problem you want to solve. Let's take the chance to do it properly. 

There are a few ways to go about this. By far the easiest is to start over, close this pull request and start a new branch and copy paste your changes. Don't go so crazy on the merges unless you have to. Create a new pull request, if Github later says it's out of date that's completely fine. You should only care and start merging/fixing if Github says it has conflicts that needs to be resolved. 

There are ways of fixing your current commits and this whole branch with `git rebase`, essentially rewriting it's history. But I would advise you to take the easier approach.

QuiZr:
@Mizipzor  There is also `git reset --hard upstream/master` - this should overwrite whole history, then he could manually add those changes again and whole pull request would be clean. Is it also a valid solution or this make some kind of mess somewhere else? I've never used git in projects larger than 2-3 people. 

Example usage in #46 - after many hours I've reseted my branch to upstream master to remove about ten old commits and pushed my changes in one, clean commit (also there was that merge but it's unrelated to this topic).

Mizipzor:
It doesnt make a mess unless someone has pulled your pull request into their fork before its merged and started making changes on top of it. This shouldnt happen. I assume you realize theres a myriad of problems that could arise from that scenario.

A common comment in Git docs is that you shouldnt "rewrite history", but I think you should when its feasible. And a pull request is not set in stone and should be open to (history) changes. Especially if it makes for a cleaner merge.

PR #157, opened 2016-08-18
fix mining drone station es translation

svmnotn:
As @frankitox16 mentioned in #124 this is a better translation. Don't translate while asleep kids :D

### DIFF
--- a/Assets/StreamingAssets/Localization/es_ES.lang
+++ b/Assets/StreamingAssets/Localization/es_ES.lang
@@ -21,5 +21,5 @@ furn_stockpile=Reservas
 furn_stockpile_desc=Una lugar para acumular tus materiales.
 furn_oxygen_generator=Generador de Oxígeno
 furn_oxygen_generator_desc=Generando oxígeno para tus astronautas.
-furn_mining_drone_station=Estación de extracción de Materiales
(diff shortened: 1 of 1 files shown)


beto2912:
I'm spanish, and a more accurate translation would be "dron de mineria"

svmnotn:
@beto2912 I was not going for a literal translation just whatever felt right given the meaning it has in english, but you are welcome to send a PR fixing it. Honestly I should of named it es_US o es_DO since I'm not from Spain but I figured that is the one that is used the most... so... feel free to send a pr using proper spanish (as in from spain) idioms.

frankitox16:
@beto2912 I suggested that translation because it sounded better imo... My native language is spanish, and I do know that translation that you gave is the literal one, however if you read "estacion de minería robótica", I guess you'd be expecting what the sprite looks like, a big monitor that says mining drone. 

It's also worth noticing that "Salvar" isn't the correct translation for "Save" (in this context, at least. The right one would be "Guardar"). But that's something for another PR

svmnotn:
heh yeah... my native language is also spanish but living in the US has made translating more difficult, so sorry for the anything that looks off.

frankitox16:
Nothing to worry, you actually did a great job!

PR #158, opened 2016-08-18
Added StyleCop settings.

Mizipzor:
These settings files wont affect anyone not using StyleCop.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,7 @@ sysinfo.txt
 # Builds
 *.apk
 *.unitypackage
+
+# StyleCop
+/Assets/StyleCop.Cache
+/Assets/StyleCop.Cache.meta
--- a/Assets/Settings.StyleCop
+++ b/Assets/Settings.StyleCop
@@ -1,0 +1,269 @@
+<StyleCopSettings Version="105">
+  <Analyzers>
+    <Analyzer AnalyzerId="StyleCop.CSharp.DocumentationRules">
+      <Rules>
+        <Rule Name="FileMustHaveHeader">
+          <RuleSettings>
+            <BooleanProperty Name="Enabled">False</BooleanProperty>
+          </RuleSettings>
+        </Rule>
+        <Rule Name="ElementsMustBeDocumented">
+          <RuleSettings>
+            <BooleanProperty Name="Enabled">False</BooleanProperty>
+          </RuleSettings>
+        </Rule>
+        <Rule Name="PartialElementsMustBeDocumented">
+          <RuleSettings>
+            <BooleanProperty Name="Enabled">False</BooleanProperty>
+          </RuleSettings>
+        </Rule>
+        <Rule Name="EnumerationItemsMustBeDocumented">
+          <RuleSettings>
+            <BooleanProperty Name="Enabled">False</BooleanProperty>
+          </RuleSettings>
+        </Rule>
+        <Rule Name="DocumentationMustContainValidXml">
+          <RuleSettings>
+            <BooleanProperty Name="Enabled">False</BooleanProperty>
+          </RuleSettings>
+        </Rule>
+        <Rule Name="ElementDocumentationMustHaveSummary">
+          <RuleSettings>
+            <BooleanProperty Name="Enabled">False</BooleanProperty>
+          </RuleSettings>
+        </Rule>
+        <Rule Name="PartialElementDocumentationMustHaveSummary">
+          <RuleSettings>
+            <BooleanProperty Name="Enabled">False</BooleanProperty>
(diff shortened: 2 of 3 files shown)


quill18:
Do the StyleCop files have to be inside of Assets or could they be in a separate folder in the project root?

Mizipzor:
I'm afraid not. The convention is that the setting file applies to all code below the setting file in the folder hierarchy. So if we want it to apply to both `Scripts` and `UI ` (and any new folders) it has to be there. If we would put it in a sibling folder it wouldnt have any affect, or rater we would run StyleCop with default rules.

Is there a problem with having them there? I'm willing to fix any file traversing code that accidentally picks them up.

lameox:
@quill18 @Mizipzor i dont think there is a problem with keeping the file in Assets. If push comes to shove we could still have to identical files in the subfolders but i dont see the problem with having it in the root folder. 
@Mizipzor do you know if StyleCop would pick up on the file if it was hidden like the .git folder and others? That way we would no need to roll code to exclude it.

Mizipzor:
@lameox the "hidden" attribute in windows is probably meaningless in most tools except windows explorer. And the way to hide files in unix is to prepend a period, and we cant change a name. I still dont know the reason why its preferred for the file not to be there. Maybe just clutter?

Mizipzor:
The wiki button just redirects back to the project main page. So I'll just put the documentation here in the meantime.

-

To install:

1. In MonoDevelop in the menubar click **Tools** and then **Add-in Manager**.
2. Click **Gallery**.
3. In the search box in the upper right type **StyleCop**.
4. Should be one hit, select it and click **Install**.

To use:

1. In MonoDevelop in the menubar click **Tools** and then **Run StyleCop**.
2. You should see errors highlighted in your current file, if any.
3. Errors are also presented as compiler warnings.
4. Run like this it scans the entire project, you can also start it by right clicking a file tab to run on just that file.

lameox:
@Mizipzor I know that the standard is for programs and tools to not care about visibility but i did not know if StyleCop behaved the same. I had guessed that quill just had a problem with the file in the root folder for those reasons, as for example the .gitignore or .vs folders reside their too but are invisible. Other than that i couldn't think about a reason why quill would not want it in the root folder.

However [this](https://stylecop.codeplex.com/wikipage?title=Sharing%20StyleCop%20Settings%20Across%20Projects) documentation page suggests under "Linking To A Settings" File that we could have it in another folder and the users would then need to do an extra step in setting up style cop. Again, i don't really see the benefit of moving the file though

Mizipzor:
I tried moving the settings file to a ´.stylecop´ folder and using the link settings trick. The good news is that it worked. The bad news is that that file is implicitly created every time you run StyleCop.

lameox:
@Mizipzor So now we have one in .stylecop and an autogenerated one in assets if i understand you correctly?

Mizipzor:
No, it was just a test. It didnt work out so its not committed anywhere. I think what this PR currently contains is the proper way to do it.

lameox:
@Mizipzor i got that. i just meant that that would be the case if we had a separate folder.

Mizipzor:
@lameox it would indeed, yes. I see no way of getting rid of this file right now.

lameox:
Well i guess than we have to live with a wile in the assets folder

Mizipzor:
I'm gonna keep rambling here. :)

I realized that Visual Studio Express doesn't allow the installation of add-ins. So you can only run it from MonoDevelop or not-express Visual Studio. However, there is [a way](http://headsigned.com/article/intergrating-stylecop-with-visual-studio-express) by manually changing the `.csproj` file, which I dont really like. There's also a [small tool](https://github.com/Nylle/StyleCop.Console) that makes StyleCop a command line application.

Again, this is not meant as something everyone is doing all the time. Especially not if there's a bit of a setup. It's enough if its run every once in a while by someone. I'm one of those that consider concise and enforced code style is important, so I might do it anway.

Mizipzor:
Tested now and it **is** possible to put the settings file in the project root right next to the solution folder (but not in another subfolder). This way the settings are picked up if you run StyleCop on the solution. The main drawback is that you, for some reason, cant change those settings by right clicking on the solution. You can only change settings on a project.

I just realized how long this became, it was supposed to be something minor. Hope you have the stamina to read through it. :/

Mizipzor:
I was wrong! 

Apparently you can run StyleCop from within Visual Studio Express. Because its not an addin. Its included by default! I wonder if this means Microsoft will push harder for StyleCop going forward. :)

quill18:
Merging as-is, but if you can find the best way to minimize the confusion that people might have about seeing the file in the Unity hierarchy, the better!

PR #163, opened 2016-08-18
Save/Load Dialogox localization

relinu:
Adds Save & Load Dialogox localization


Mizipzor:
Now this is a nice clean PR. :)

PR #165, opened 2016-08-18
Metal Smelter

UnknownSoldier1:
I added a new type of furniture, the metal smelter. It will smelt Raw Iron (which is now produced by the Mining Drone) into Steel Plates.
Since I'm a terrible artist, the two textures i made are only dummies and should be replaced by somebody who has more tallent than me.

Also, I'm sorry for all these additional commits. All of these changes that GitHub says I did are wrong (I didn't change these files at all). And since I'm new to GitHub, I wasn't able to clear this up. I hope it will work anyways.


ComputerOverlord:
Oooh! Should be added!

ComputerOverlord:
It... CONFLICTS AGAIN

quill18:
Closing because this modifies way too many unrelated (AFAIK) files.

quill18:
This is a GOOD feature and I want to include it, but the it'll forever conflict (and probably change things we don't want changed) as-is.

PR #166, opened 2016-08-18
Adding airlocks and making normal doors leak air

emilkris33:
This is part of the goals outlined in #49

Makes doors leak air between rooms on either side. Is generalised so it can easily be controlled and applied to any furniture through lua. 
This has a dual purpose. 1 air can leak into space. 2 O2 generators are not needed in every room.

Also ads airlock. More ekspensive differently colored door that only leaks air when open.


ComputerOverlord:
Great Idea. +1

emilkris33:
@ComputerOverlord Leak speed is modified by pressure difference in this implementation

ComputerOverlord:
@emilkris33 Oh, I didn't look at it in-depth :P I just understood the idea.

fre-ber:
I made a few attempts of getting this right in my own version of this game, as made during the video series. Unfortunately I never managed to get anything that felt natural enough. I wonder if the mechanic presented here actually does that trick. I haven't tried this code, but as the pressure depends on the volume of the room the mathematice feels off to me. 

In my attempts I changed the gas dictionary to actually hold the mass of the gas as an absolute value and tried to figure out a formula for how much mass is actually transferred though a vent/door/deconstructed wall/etc. Googling this topic, however showed me just how deep this rabbit hole can go - the flow depends on the shape of the hole, roundness of the edges of the hole, smoothness of the inner sides of the hole, turbulence, temperature, density of the gas, pressure, etc. 

How closely have you watched the behaviour of this code? Does it feel natural to you? I might implement the equivalent code in my own project and test for myself, but I would be happy to know how realistic you think this feels.

Also, have you thought about how to handle pressure equalisation when rooms are merged?

GamerInspired:
What if there was some sort of valve that you could use the let O2 flow between room but also allowed you to close them if required?

emilkris33:
@fre-ber I have to say it myself then the gas handling is very thought through. Most of the basic gas mechanics are from an earlier PR #77 . Gas is stored internally as pressure times volume(Units are atm times tiles). The O2 generator produces in terms of this value, but the tooltip shows it divided by number of tiles. How much gas is transferred is dependent on the pressure difference, deltaTime, and a leakFactor for the object. LeakFactor for door is greater the more open the door is. The leakfactor takes care of all the weird variables and just has to be tuned to something that feels right. Tweaking of the leak values is still needed though i would like to have a more final version of the system before spending time on that.

Handling of merging and splitting rooms is also good. (implemented in #77 ) When spitting a room the atmosphere so the pressure in both rooms are as it where before, though the internal value different. When merging the total gas in the previous rooms get added.

@GamerInspired That can be implemented using this code without touching the C#. Only lua and xml and a new graphic is needed to add that. It can therefore also be modded in. Though for now you can not open for only one gas.(Though for now there is only one gas)

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -40,12 +40,19 @@ function OnUpdate_Door( furniture, deltaTime )
...
 	end
 
-
 	furniture.SetParameter("openness", Clamp01(furniture.GetParameter("openness")) )
 
 	furniture.UpdateOnChanged(furniture);
 end
 
+function OnUpdate_Leak_Door( furniture, deltaTime )
+	furniture.tile.EqualiseGas(deltaTime * 10.0 * (furniture.GetParameter("openness") + 0.1))
```
vogonistic: Should the `0.1` leak in normal doors be a parameter so that it'd be easy for other people creating doors to add more or less of it?

emilkris33: Yes, I might look at doing that tomorrow. I this PR gets merged before that I will do it as part of my next changed to the atmosphere system(Unless someone beats me to it)

fre-ber:
Cool, I will test something similar in my own project. Does the equalisation on room merge happen over time or is it instantaneous?

________________________________

quill18:
Merged!

PR #170, opened 2016-08-18
Cleaning World

simonlary:
Deleted lots of old commented code, for example the old _CreateFurniturePrototypes_, and deleted some unecessary _Debug.Log_.


Assets/Scripts/Models/World.cs
```diff
@@ -244,120 +236,8 @@ void CreateFurniturePrototypes()
             Debug.LogError("Did not find a 'Furnitures' element in the prototype definition file.");
         }
 
-        Debug.Log("Furniture prototypes read: " + furnCount.ToString());
```
OzieGamma: Why do you want to remove logging ?

Assets/Scripts/Models/World.cs
```diff
@@ -581,20 +457,11 @@ public void WriteXml(XmlWriter writer)
...
-*/
-
-        //Debug.Log(writer.ToString());
-	
+        writer.WriteEndElement();	
     }
 
     public void ReadXml(XmlReader reader)
     {
-        Debug.Log("World::ReadXml");
```
OzieGamma: Why remove logging ?

NogginBops:
@OzieGamma These lines are not needed. The features work and are unlikely to provide any useful debug data in the future. There is a discussion about this in #94 if you want to participate.

OzieGamma:
@NogginBops thanks for the clarification !
I have been going through PRs and Issues. I unfortunatly read them in the wrong order ;)

simonlary:
Doesn't look like I need to explain more, but as @NogginBops mentionned, these _Debug.Log_ didn't give any useful information and were just filling the console, making debugging harder.

Mizipzor:
Instead of just deleting the noisy logging, maybe consider using the `Verbose` level in the custom logger in #185? Assuming its merged. Disclaimer: I made it.

simonlary:
Some of these could definitely be in the `Verbose` level of the custom logger, but it didn't exist when I made these changes. Also, some of these logs were only there to make sure the functions were getting called. I don't think we need to log every function call, even in verbose. For example : `Debug.Log("CreateCharacter")` doesn't really give any information.

WardBenjamin:
Agreed, these logs are pretty unnecessary. +1

PR #169, opened 2016-08-18
Added text for power generator.

SanteriHetekivi:
I noticed that Power Generator (#78) does not have its name set in a button.

**Before**
image

**After**
image

This request fixes that text and also adds missing BuildingJob tags for Power Generator.

The name barely fits the button and I have already opened issue (#164) for button resizing.

Issue #153, opened 2016-08-18
Discussion: Localization, how to handle it?


emilkris33:
We have already implemented a localization system to the code which look great. But this means that as we add text to the game we are going to get a lot of PRs like #146 with a lot of discussion most of us can't follow. Also this could be quite a lot of PR once we start adding strings and get people working on more languages. In general i think git is a lot better for handling code collaboration than localization collaboration that tells me that we should probably have translation happen somewhere else.

The way this would work is that we add only english strings while working in unity. Then when a new string gets to the TeamPorcupine/Master through PR the string list gets pulled into another tool where people(even those who don't want to touch unity or git) can translate. And then from that tool we export the translation files to unity, either at regular intervals or when ever significant changes exists. For this we could use one of the tools listed at https://github.com/integrations/feature/localization that have build in git integration or another tool. I have some experience using https://translate.google.com/toolkit/ which while not git integrated has the plus of being completely free.

What are everybody's thoughts on this?

bjubes:
I agree that translating should occur outside of the git repo, especially because the subset of people who translate is different that those who code even if some people are doing both. 

At this point in the project, many names and ui elements with text are very subject to change. while having code to handle localizations is nessicary, right now any language other than English may very well be a a waste of time when names of items are changed. 

Tldr; implement localization code now, translate and set up translation integration when the project's names and text are final

SanteriHetekivi:
I also agree.

Made Finnish translation, just for a test of the translation system.
That got me thinking how it would be pain to update every time something changes in English version.

I would think that mostly everyone who are working on this project can at least understand English.

Also before UI resizing is implemented some of the translations are useless because of text length. -#164

Tranberry:
What does Quill think, have he said anything on this topic somewhere?

emilkris33:
@Tranberry I don't think @quill18 has addressed this. But he might see now that he has been tagged

Fogsight:
It is a great idea to have more accessible area for people to contribute their translations, but languages should be represented right from the start for localization to work properly. They just could be updated properly later.

Because in Russian, for example, you'll need to mutate nouns depending on situation, it's less flexible than English. (i.e. This is "Steel Wall", but adding "Build" would change its form, and has to be a separate string, which can be avoided if you don't add "Build" in the first place.)

I'm not a linguist BTW, I'm sure there are many more pitfalls and in other languages as well.

Aiwha:
I agree with Fogsight.
[...]

PR #173, opened 2016-08-18
Add Russian language localization

Fogsight:


emilkris33:
@Fogsight Good, please have a look at #153 . We need input from people who want to localize.

SanteriHetekivi:
Could you add translations to Russian for:
furn_power_generator=Power Generator
furn_power_generator_desc=Generates power out of nothing.

They are missing in current master, but pull request #169 should add them in.

PR #175, opened 2016-08-18
Asteroids/WorldGenerator

relinu:
Adds a WorldGenerator, cann add asteroids or a starting area.

If this gets to many thumbs down I'll remove it ^^'

### DIFF
--- a/Assets/Scripts/Models/WorldGenerator.cs
+++ b/Assets/Scripts/Models/WorldGenerator.cs
@@ -3,14 +3,14 @@ using System.Collections;
 
 public class WorldGenerator {
 
-    public static int startAreaSize = 3;
+    public const int startAreaSize = 3;
 
-    public static TileType asteroid_floor_type = TileType.Floor;
-    public static float asteroid_noise_scale = 0.2f;
-    public static float asteroid_noise_threshhold = 0.75f;
-    public static float asteroid_ressource_chance = 0.85f;
-    public static int asteroid_ressource_min = 5;
-    public static int asteroid_ressource_max = 15;
+    public const TileType asteroid_floor_type = TileType.Floor;
+    public const float asteroid_noise_scale = 0.2f;
+    public const float asteroid_noise_threshhold = 0.75f;
+    public const float asteroid_ressource_chance = 0.85f;
+    public const int asteroid_ressource_min = 5;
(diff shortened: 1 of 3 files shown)


relinu:
screenshot04

vogonistic:
Nice! Maybe add a higher threshold where, if it's reached you can put a scaled stack of steel plates on there, as a placeholder for better astroid resources later. That way they aren't just pretty, they are also useful.

ComputerOverlord:
It's great! You should add some sort of asteroid furniture (that can't be built)

emilkris33:
@relinu Maybe you should start people out with some steelplates so they can get building

ComputerOverlord:
@emilkris33 I think you should have to mine asteroids first.

vogonistic:
@ComputerOverlord Mining hasn't been built yet. Plates exists and fit well as a placeholder.

relinu:
screenshot05

So many steel plates *~*

ComputerOverlord:
@relinu You should limit how many plates there are. 1; Maybe stacks could be smaller, 2: Halving or even dividing the amount by 4 _might_ be good...

relinu:
Someone can adjust the values later, if you look at the code you can see that there is a min, max value of stacksize and a plate spawnchance.

ComputerOverlord:
@relinu I figured that was the case.

RodRitter:
I love it!

Assets/Scripts/Models/WorldGenerator.cs
```diff
@@ -0,0 +1,66 @@
...
+public class WorldGenerator {
+
+    public static int startAreaSize = 3;
+
+    public static TileType asteroid_floor_type = TileType.Floor;
+    public static float asteroid_noise_scale = 0.2f;
+    public static float asteroid_noise_threshhold = 0.75f;
+    public static float asteroid_ressource_chance = 0.85f;
+    public static int asteroid_ressource_min = 5;
+    public static int asteroid_ressource_max = 15;
```
OzieGamma: If these are constants, I suggest using `public const int ...`

ComputerOverlord:
@quill18 should get to this.

quill18:
An excellent start to world generation. Merged!

PR #184, opened 2016-08-18
Added missing localization and "fixed" some german

Ceddini:
I added the missing localization for the power generator, and while I was there, I changed some german words to make it sound "better".


Ceddini:
Somehow the changed files didn't show up... (I'm completely new to this..)

PR #185, opened 2016-08-18
Added custom logging class wrapping unity logger.

Mizipzor:
This was discussed in #94.

I've made the methods look like the ones already in Unity so that you can simply swap anywhere it says `Debug.LogError()` to say `Logger.LogError()` and all its variants. 

It has a minimum level that defaults to `Info`, only that level and more severe levels are logged. This means by default the level `Verbose` is hidden.

Verbose is the only new "level" compared to what Unity has. It's intended for everyone to go completely nuts with logging here. Since its hidden by default it wont affect anyone and should only be enabled when developing or debugging. This addresses the concerns with how much you are allowed to log, and how much of your debugging logging you should strip out when you make a pull request.

There is a neat trick to using this that might not be immediately obvious. What you do is set the `minimumLevel` to `Exception`, essentially hiding _all_ logging. Then you **don't** use the logging class but instead use the regular Unity `Debug.Log` thereby bypassing the filter. When you are done, you change all your logging lines to use `Logger`.

### DIFF
--- a/Assets/Scripts/Utilities/Logger.cs
+++ b/Assets/Scripts/Utilities/Logger.cs
@@ -1,0 +1,134 @@
+﻿using System;
+using UnityEngine;
+
+public static class Logger
+{
+    public enum Level
+    {
+        Exception,
+        Assertion,
+        Error,
+        Warning,
+        Info,
+        Verbose
+    }
+
+    // TODO: Not used yet but it's the logical next step.
+    public enum Channel
+    {
+        Default
+    }
+
+    static Level minimumLevel = Level.Info;
+
+    public static void LogException(Exception exception)
+    {
+        // Exceptions are special as they support neither formatting nor message
+        if (minimumLevel <= Level.Exception)
+        {
+            Debug.LogException(exception);
+        }
+    }
+
+    public static void LogAssertion(string message)
+    {
+        Log(Level.Assertion, message);
+    }
+
+    public static void LogAssertionFormat(string format, params object[] args)
+    {
+        Log(Level.Assertion, format, args);
+    }
+
+    public static void LogError(string message)
+    {
+        Log(Level.Error, message);
+    }
+
+    public static void LogErrorFormat(string format, params object[] args)
+    {
+        Log(Level.Error, format, args);
+    }
+
+    public static void LogWarning(string message)
+    {
+        Log(Level.Warning, message);
+    }
+
+    public static void LogWarningFormat(string format, params object[] args)
+    {
+        Log(Level.Warning, format, args);
+    }
+
+    public static void LogInfo(string message)
+    {
+        Log(Level.Info, message);
+    }
+
+    public static void LogInfoFormat(string format, params object[] args)
+    {
+        Log(Level.Info, format, args);
+    }
+
+    public static void LogVerbose(string message)
+    {
+        Log(Level.Verbose, message);
+    }
+
+    public static void LogVerboseFormat(string format, params object[] args)
+    {
+        Log(Level.Verbose, format, args);
+    }
+
+    /// <summary>
+    /// Alias for info.
+    /// </summary>
+    public static void Log(string message)
+    {
+        Log(Level.Info, message);
+    }
+
+    /// <summary>
+    /// Alias for info.
+    /// </summary>
+    public static void LogFormat(string message, params object[] args)
+    {
+        Log(Level.Info, message, args);
+    }
+
+    public static void Log(Level level, string message, params object[] formatArgs)
+    {
+        if (level > minimumLevel)
+        {
+            return;
+        }
+
+        if (formatArgs.Length > 0)
+        {
+            message = string.Format(message, formatArgs);
+        }
+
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Utilities/Logger.cs
```diff
@@ -0,0 +1,134 @@
...
+        Verbose
+    }
+
+    // TODO: Not used yet but it's the logical next step.
+    public enum Channel
+    {
+        Default
+    }
+
+    static Level minimumLevel = Level.Info;
```
OzieGamma: Wouldn't `const Level defaultMinimumLevel = Level.Info;` be more redable ?

Assets/Scripts/Utilities/Logger.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        if (formatArgs.Length > 0)
+        {
+            message = string.Format(message, formatArgs);
+        }
+
+        switch (level)
+        {
+            case Level.Exception:
+                throw new InvalidOperationException("Exceptions are logged in a special way. You shouldnt see this message");
```
OzieGamma: This seems very unexpected behavior. Is that what Unity does ?
I mean the last thing I want as a dev is the logger to throw exceptions ... Better to log the fact that this is not proper use of the function no ?

I would simply do

```csharp
Debug.LogException(new InvalidOperationException("Exceptions should be logged using LogException(Exception). Log message: " + message));
```

Assets/Scripts/Utilities/Logger.cs
```diff
@@ -0,0 +1,134 @@
...
+                Debug.LogWarning(message);
+                break;
+            case Level.Info:
+                Debug.Log(message);
+                break;
+            case Level.Verbose:
+                Debug.Log(message);
+                break;
+            default:
+                throw new ArgumentException("Unhandled level: " + level, "level");
```
OzieGamma: Again, I would log this instead of throwing. Why make the game crash to the player because a Log was not properly configured ?

Assets/Scripts/Utilities/Logger.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        Log(Level.Info, message);
+    }
+
+    public static void LogInfoFormat(string format, params object[] args)
+    {
+        Log(Level.Info, format, args);
+    }
+
+    public static void LogVerbose(string message)
```
OzieGamma: Add [ConditionalAttribute("DEBUG")] ?

So that all the LogVerbose functions don't even make it to a release build ? (Not sure if supported by Unity)

http://stackoverflow.com/questions/5080477/debug-only-code-that-should-run-only-when-turned-on

Assets/Scripts/Utilities/Logger.cs
Mizipzor: You're absolutely right, I forgot that this method was public and someone might actually call it with the Exception level manually. I will probably change this method to private. Good review.

Assets/Scripts/Utilities/Logger.cs
Mizipzor: No. I believe in crashing early and crashing often. This default forces a crash rather than silently ignoring the message. This can currently not happen. This is only meant for when someone adds a new level but forgets to add the corresponding switch statement. If this "crash" gets out, it means that someone added a level but never bothered to test logging with that level.

Assets/Scripts/Utilities/Logger.cs
Mizipzor: The thing is I _want_ this to be available in a release build. If someone has a bug, we ask to have them reproduce the bug with verbose logging enabled.

Assets/Scripts/Utilities/Logger.cs
OzieGamma: I understand and the compiler will probably issue a warning. 

But should the game crash ? For a wrong log level ? Can't we log the fact the we used the log wrong ? (Meta-logging !)

Assets/Scripts/Utilities/Logger.cs
OzieGamma: Good point ! People were saying that there is too much logging and it's a performance issue. That's were my comment comes from

Assets/Scripts/Utilities/Logger.cs
Mizipzor: Crashing is not bad. As long as it happens at the developer. It's a mistake you can't miss. A logged error message you can miss.

Crashes at the end user is bad. And for this case, it means that code has reached the end user without ever being run. Which is really bad. And for that case I'm willing to risk a crash.

But to be fair, "to crash or not to crash" is one of the holy wars at work. I say crash early and often, some say crash never.

OzieGamma: Indeed. I say: try to recover, if you can’t crash early and often ;)

quill18:
We definitely wanted a custom logger at some point so that we could also redirect messages to log files on the end user's computer to aid in debug. Merged!

PR #186, opened 2016-08-18
Added missing localization (en_US, es_ES & de_DE) and "fixed" german

Ceddini:
I added the missing translations for the power generator (de_DE & en_US).

And while I was there I also "fixed" some german translations to make it sound better.


Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -119,12 +119,18 @@
...
-    <Height>2</Height>
+		<Name>Power Generator</Name>
+		<MovementCost>10</MovementCost>
+		<Width>2</Width>
+		<Height>2</Height>
 
-    <PowerGenerator supply="5"/>
+        <BuildingJob jobTime="0">
+        </BuildingJob>
+
```
Dormanil: You might want to change these tabs to spaces, since we decided to indent with four spaces equalling a tab.

Tranberry:
Makes me wonder is the default language/localization en_CA?

Dormanil:
It is en_US, @Tranberry.

Tranberry:
Funny that @Dormanil .

Dormanil:
I mean, that wasn't my decision. It just happened when the localization controller was implemented, even though this project originated in Canada and technically should be en_CA instead. 

Wait, quill18 lives in Northern Ontario, shouldn't it be fr_CA instead? ;)

SanteriHetekivi:
Power Generator translation implementation is already in #169, but that has no German translation.

Koljenka:
Pfadfindungs-Test can be one word (Pfadfindungstest) and i think that that looks alot nicer.

Koljenka:
and Einrichtung abbauen should something like Einrichtung entfernen

Ceddini:
@Koljenka I did that to emphasize the Test part, so that it's a lot more clearer, that this is just a Test.
But I can change those things. And I was unsure if I should change the part with "Einrichtung abbauen" because "abbauen" for me means regaining materials, while "entfernen" means just removing something. But I will change those things to make it better :D

Koljenka:
@Ceddini youve got a point with the materials but "Einrichtung abbauen" just sounds strange to me. :smiley:

Ceddini:
@Koljenka Same for me :D I will now change both things to make it better!

svmnotn:
@Ceddini here is the fix for the es_ES translation:
```
furn_power_generator=Generador eléctrico
furn_power_generator_desc=Genera electricidad de la nada. 
```

Ceddini:
@svmnotn Thanks, I will add it 😄

Ceddini:
Hey @svmnotn, could you possibly also translate these new things? Would be amazing 😄 

    furn_airlock=
    furn_airlock_desc=
    tile_Empty=
    tile_Empty_desc=
    tile_Floor=
    tile_Floor_desc=

The only way I could do it is google translator, but that would be a bad idea... 😃

Xeni91:
Well can approve of this german translation will close #256 for this.

SanteriHetekivi:
@Ceddini Now this PR is included in #169.

Ceddini:
Incorporated these changes into #169 as discussed in issue #299

PR #187, opened 2016-08-18
Fixed forgotten txt & lang file extension in .gitattributes and normalized line endings.

QuiZr:
This leaded to problems in committing same .txt files but with different line endings like in #165.
Never ending line endings story... >.>


WardBenjamin:
Looks good! I would also add .md.

henninglive:
This does not fix the end-of-line problem for me, but i think i have i a theory of what's happening. The text files in the font directory contain CRLF line endings since they were added to the repository before end-of-line normalization was enable through #41. Normalization ensures that all files that Git considers to be text will have normalized LF line endings in the repository, but when checking files in and out the line endings will be converted to the native line ending for your platform.

When @quill18 enabled normalization he forgot to normalize the existing files in the repository. The fix for this is to allow someone to commit the normalized files.

From [git documentation](https://git-scm.com/docs/gitattributes):
> When text=auto normalization is enabled in an existing repository, any text files containing CRLFs should be normalized. If they are not they will be normalized the next time someone tries to change them, causing unfortunate misattribution.

vogonistic:
I vote we add the pesky .txt files in the Fonts and Music directories to this story and get it sorted out once and for all. It's sad to see so many PR's be rejected.

QuiZr:
@henninglive Yeah, I think that this is our problem.
@vogonistic I've added those .txt files - this should be now resolved once and for all (as long as this request get's accepted)

Raketfart:
Yes please! This hopefully would solve all my git problems

WardBenjamin:
This definitely needs merged @quill18. Normally .txt files would be recognized as text by git, but since we're using a custom .gitattributes file they're being recognized as binary, and the line endings are being changed per-PC, which is leading to a lot of extraneous line changes in pull request.

Mizipzor:
Yes, this is a problem in many pull requests. It should be a priority.

vogonistic:
@QuiZr Can you fix the conflict so we can get this merged?

QuiZr:
Done

vogonistic:
@quill18 This one needs merging for 🎉 😻 .

PR #189, opened 2016-08-18
French Translation

Aiwha:
Added French.


Aiwha:
I fell like I fucked up something. Only wanted to change fr_FR.lang.
And I realised I didn't add the power generator translation. Is there a way to un-request the pull request?

Tranberry:
Why are the font licenses changed? /sorry for n00b question

NogginBops:
@Aiwha You could close your current request and open a new one.

emilkris33:
@Aiwha You need to open up unity with your translation added so it can generate .meta files. Else everybody that gets your code will generate .meta files and we will have conflicts:

Aiwha:
@NogginBops Thanks!
@Tranberry Well… no idea.
@emilkris33 Thanks! I have not Unity. I think I'll wait until a solution is found in #153, that'll be easier.

Issue #182, opened 2016-08-18
Standards for variable naming/formatting


behindcurtain3:
Has there been any discussion on standards for this? Looking through the code it is a complete mix of lowercase, uppercase, public and private variables (some properties also). Does anyone have any suggestions on how this could be handled the best (and consistently) going forward?

ComputerOverlord:
For when quill finds this...
testVar
TestVar
or testvar?

BeefEX:
@behindcurtain3 most of this mess if by @quill18 himself :smile:

OzieGamma:
Related to #151

lameox:
I suggest we use the Microsoft Coding Conventions for everything C# Related
which means:
```
//camelCasing everywhere
//descriptive names for variables/methods
//fields, properties and methods should always specify their scope, aka private/protected/internal/public

interface ISomeInterface{ } //Interfaces starting with a captial I

class Class { //class names Uppercase

   private int _memberField; //fields start with an underscore and should be private if possible

   public int MemberField { get { return _memberField; } } //Properties should be captialised

  // Functions should be capticalised
  // Function parameters should be camelCase
   public void SomeFunction( int functionParameter ) 
   {
      int myLocalVariable = 0; //local variables should also be camelCased
   } 

   public event SomeEvent; //Events should be capitalised as well 
}
```

I think this would be the easiest as we could just link the MSDN Style Guide for our Contributors.
Also i think its the prettiest way :D

OzieGamma:
@lameox MS code style has braces on the next line:

```csharp
class Class
{ // On this line
}
```

behindcurtain3:
@BeefEX I don't want to point fingers :D @lameox I like those conventions as that is what I use in personal projects.

lameox:
@OzieGamma oh shit yeah. Will edit

BeefEX:
@behindcurtain3 i dont like them but still use them @D

Mizipzor:
StyleCop, in #158, complains if you name something incorrectly.

satch3l:
Isn't code standars mentioned in the readme or contribution files? 

Differentiations may be leftovers from earlier in the development. However I'm sure those will be changed in the future.

lameox:
Well it says 
> We have standardized on Microsoft's C# Coding Conventions and General Naming Conventions, with a few exceptions.

>Avoid using 'var', even when the type would be clear from context. Verbose typing is best typing.

[...]

PR #192, opened 2016-08-18
Specify Coding Style for Contributions according to #182

Dormanil:
Fixed some ambiguity in the Style Guidelines and added @lameox's TL;DR for our coding style.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,26 +35,33 @@ We have standardized on Microsoft's [C# Coding Conventions](https://msdn.microso
 As a TL;DR on our coding practises, adhere to the following example:
 
 ```c#
-//camelCasing everywhere
-//descriptive names for variables/methods
-//fields, properties and methods should always specify their scope, aka private/protected/internal/public
+// Use camelCasing everywhere.
+// Descriptive names for variables/methods should be used.
+// Fields, properties and methods should always specify their scope, aka private/protected/internal/public
 
-interface IInterfaceable { } //Interfaces start with an I and should use PascalCasing
+// Interfaces start with an I and should use PascalCasing.
+interface IInterfaceable { } 
 
-class Class //class names should use PascalCasing
-{ // braces on new line ;)
-   private int _memberField; //fields start with an underscore and should be private if possible
-
-   public int MemberField { get { return _memberField; } } //Properties should use PascalCasing
-
-  // Methods should use PascalCasing
-  // Method parameters should be camelCased
-   public void SomeMethod( int functionParameter ) 
-   {
-      int myLocalVariable = 0; //local variables should also be camelCased
-   } 
-
-   public event SomeEvent; //Events should use PascalCasing as well 
+// Class names should use PascalCasing.
+// Braces are on a new line. ;)
+class Class 
+{
+    // Fields start with an underscore and should be private if possible.
+    private int _memberField; 
+    
+    // Properties should use PascalCasing.
+    public int MemberField { get { return _memberField; } } 
+    
+    // Methods should use PascalCasing.
+    // Method parameters should be camelCased.
+    public void SomeMethod( int functionParameter ) 
+    {
+        // Local variables should also be camelCased
+        int myLocalVariable = 0; 
+    } 
+    
+    // Events should use PascalCasing as well 
+    public event SomeEvent; 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


lameox:
I like the new wording

QuiZr:
There is `this` keyword, should we really use underscores in private fields?

lameox:
I think it really helps when you read a function body because nobody will specify a this for every variable they use and with the underscore you can easily see what modifies object state and what just changes local variables.

behindcurtain3:
Looks good!

Mizipzor:
What's the reason for using underscores as a prefix for fields? I was under the impression that we were using the default [Microsoft style](https://msdn.microsoft.com/en-us/library/ms229002(v=vs.110).aspx), specifically for this would be the [Naming of type members](https://msdn.microsoft.com/en-us/library/ms229002(v=vs.110).aspx):

> X DO NOT use a prefix for field names.

I would strongly advice against coming up with a custom style. Writing an example class is a good idea. But link to the Microsoft style and follow that. This is one of the holy wars we dont want to start here.

vogonistic:
Hmm, can you update the styles to indicate that this is `lowerCamelCase` and this is `CamelCase`

NogginBops:
@vogonistic Actually this is `camelCase` and this is `PascalCase`.

lameox:
Actually its PascalCase and camelCase but i guess that is not really important to the topic :D

@Mizipzor oh i had thought microsoft would advise doing the same as their .Net Core people advise [here](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md). I found that coding style the easiest to understand quickly. 

But i guess we can also do it like in the msdn documentation as long as we reach a consensus and have a unified style.

Dormanil:
@Mizipzor Microsoft also states in the same style document:

> The field-naming guidelines apply to static public and protected fields. Internal and private fields are not covered by guidelines, and public or protected instance fields are not allowed by the member design guidelines.

Since only private fields start with an underscore, we still comply with.

lameox:
@Dormanil thank you for going over the guidelines and looking for that. Would have been weird if they contradicted themselves. I propose we use their underscore policy for private fields so we get a ruleset with less ambiguity

vogonistic:
I stand corrected.

Mizipzor:
@lameox what you linked is good. My main concern here is linking to something existing and not to try and author our own standard. It's not worth the effort. The style is not that important. But it is important that is concise and very well documented. 

How about using the default settings for StyleCop? I think they are close to that.

lameox:
@Mizipzor i just wanted to suggest a more or less official approach from microsoft that i can attest to. I think the .Net Core team also provides a custom Setting file for StyleCop. Will have to look into that though. 

But if the standard Settings are similar then why bother. I think your idea is simpler. Would running Stylecop with defaults still autogenerate the file in the assets folder?

Dormanil:
Sooo... what would the final style be then, so that I can fix the TL;DR example code in this PR?

Mizipzor:
@lameox I'm not sure I understand, what file do you expect to be generated? StyleCop runs some checks and gives you what looks like compiler warnings, lines the violate the coding standard.

Dormanil:
@Mizipzor I guess he means the StyleCop Settings in the Assets folder you are currently requesting to be pulled in #158.

lameox:
@Mizipzor yup. i referred to the file @Dormanil mentions. Is this file generated with standard settings? if not we can solve 2 issues at once...

Mizipzor:
The file is generated, and default settings would be an empty file. You specify overrides from the defaults. In my pull request I've added some overrides based on what the code currently looks like. For example, we don't appear to require methods to be documented.

NogginBops:
Shouldn't it be ``//Use camelCase unless otherwise stated`?

Dormanil:
Yes, yes it should. 

I also still need a concise answer as to the question of the underscore as a prefix to private fields.

NogginBops:
To me underscores before fields are just ugly and don't really make anything more understandable, though some people might disagree. I tend to go for a camel case approach.

koosemose:
I also agree that the leading underscore doesn't make things any more clear, with the exception of how quill has used a leading underscore so far (though I'm not absolutely sure he has used it completely consistently), where the field is backing a property (such as with the _Name field and Name property in Furniture.cs, if used consistently, I think that would provide a pretty clear signal that you shouldn't be using that field directly (even from within the class), and rather use the associated property.

Dormanil:
I liked it that way as well, when a field has backed a property. I think this would be a good compromise between underscores nowhere and underscores everywhere -- underscores where they make the most sense.

quill18:
Looks good!

And yes, I feel that the underscores are really best saved as indicators of "don't touch this directly" (i.e. backing a property), rather than used for all privates as some standards do.

quill18:
Merged -- though I always preferred the term CamelCase to PascalCase.

PR #194, opened 2016-08-18
Delete button

Koljenka:
so this is the redone version of #89 so it looks exactly the same. I hope that this will be merged soon so that there wont be any conflict issues. :wink:

PR #201, opened 2016-08-18
added double click to load and save only in code

alexisfcote:
Double clicking on a listitem now register like an okay was clicked.
Changed from the method with tags in scene.


Koljenka:
this conflicts with #194 , just saying.

WardBenjamin:
This should probably be closed now that #194 is merged.

quill18:
This conflicts now. Please update from master, merge, and push again.

Sorry you lost the race. :(

quill18:
@WardBenjamin I don't think it has to be closed -- as long as an update is made to remove the conflict it should be okay (though closing & redoing the PR also works.)

WardBenjamin:
Fair enough @quill18. I don't think it's really a necessary feature now that #194 provides an easier and more obvious way to do it, but up to you!

Koljenka:
#194 and #201 do completely different things

quill18:
@WardBenjamin I'm confused -- I didn't think #194 did the same thing as this.

Koljenka:
@quill18 #194 does not add the double click thing

WardBenjamin:
Looks like I was incorrect, thanks for the.... correction :P

I was under the impression that this just would delete on double-click, but that's not what it's for.

PR #208, opened 2016-08-18
Power generator Info (XML)

aleserena:
Added Name, Desc and Localization(ES) to Power generator


WardBenjamin:
This conflicts with the current master, due to the extraneous merges. I would open a new pull request based on master, basically by doing `git reset --hard origin/master`, assuming origin is the TeamPorcupine repo, and then re-adding the XML changes that you made. FYI, this WILL delete any other changes if you aren't careful.

aleserena:
Redone and Submited Again, Thanks WardBenjamin :)

PR #225, opened 2016-08-19
pt_BR Translation

notnanomercy:


quill18:
This PR includes files that shouldn't be included as part of the merge.

PR #243, opened 2016-08-19
Added Cloning Pod

MANDAL0R3:
Cloning pod is used to create additional characters.
Once a cloning pod is done being worked, it is immediately destroyed.

NOTE: 4+ characters all working seems to create some hiccups.
screenshot1
screenshot2


SquareMan:
Thats a really neat way to implement new characters. Probably wouldn't survive deeper into development, but I do hope this gets merged.

ComputerOverlord:
Neat!

koosemose:
With more complexity (at least some of which requires further development elsewhere), I could potentially see something like this staying in, just not as the main way to get new crewmembers. Something like an extended cooktime, beyond the donor's worktime, perhaps making use of skill (perhaps of a different character working a cloning station "overseeing" the growth of the clone), and chances for failures and mishaps, things from simply failing to produce a clone, producing a flawed clone (either just with simply reduced traits from the original and/or something from a pool of flawed clone traits), or even a completely evil clone,this could perhaps be one of the flawed clone traits, or something separate (for extra added evilness, if there was a way for it to "lie" about its traits, so it's not immediately obvious you've got an evil clone, that'd be potentially fun).

MANDAL0R3:
Thanks for all of the feedback. I originally did this as way to get new crew but at a cost (very expensive and long build/work times) with also other future implementations for other developers. I also did this as way for anyone to be able to create more characters without having to do much coding. As it stands with the AI in this projects, after having 5 AI building walls it kinda lagged ever time someone placed a new wall segment.

koosemose:
There's a few issues that are likely causing that, or at least making it worse, I know that one of them is simply that there's currently a bug where an AI will continue looking for a new path to an inventory for it's job on every tile it reaches (which may end up being once every frame if the tile is far enough.), I'm working on fixing that up as is. I'm guessing most of the other problems also relate to the pathfinding in some way or another.

Also, you have conflicts, you may want to look into resolving them.

henninglive:
You should remove the font files your commit.

MANDAL0R3:
I am still very new to using Github. How would I go about removing conflicts?

MANDAL0R3:
How would i remove the font files? I honestly don't know why those are even there. I never touched the resource folder.

NogginBops:
I think this is being fixed in #187

henninglive:
@NogginBops but there is going to a conflict with #187, if the files are not removed from this commit. This should be avoided if possible.

Issue #85, opened 2016-08-17
Doors are not updating if walls are placed after doors


Raketfart:
If you place the door first, then the walls, the door is never properly rotated.
I think placing a piece of furniture should always call update on surrounding furnite, regardless if the furniture has LinksToNeighbours set to true.
That way, a door can re-orient, if walls are placed adjacent.
I've started working on a pull request on for this.

Raketfart:
Put a pull request #92 that should fix it for now.
I know others have fiddled with updating surrounding furniture, but none of those will work with furniture larger than 1x1, and none will update a door because it doesnt connect to neighbours and is a different type than the walls.

Raketfart:
The issue is hot-fixed for now, but i wonder about the correct way to implement doors and similar furniture.
1. Should we move this logic to Lua and have a OnChange Lua function on all furniture? Then "wall connect logic", "door connect logic" and eventually things like "shelves orienting to walls"/"miningdrill aligning to adjacent rock". 
2. Is it better to have the user rotate objects when placing them, and then doors simply won't close off rooms if they are placed wrong?

Raketfart:
Fixed again in PR #442.

vogonistic:
Closing this issue since it's fixed.

PR #244, opened 2016-08-19
Fixed door rotation when walls are placed after doors

Raketfart:
Updates all surrounding furniture when deconstructing as mentioned in #85

Hotfixed door rotation if walls are placed after door placement. Still hardcoded for now, but fix is underway


Raketfart:
I still dont know how to omit the txt files from being part of my commit. 
I made a brand new fork, but these files still showed up as changed.

Mizipzor:
This is the issue discussed in #143 and #187. You could wait until it's properly fixed and merged. But there's an easier way. I don't know if it's obvious but in Git you dont have to commit every changed file. Yes, due to the problems linked above the files (four, was it?) would appear changed in addition to the files you actually changed. Commit only the files you want and create a new pull request.

Issue #209, opened 2016-08-18
Translation for (something) failed: Key not in dictionary warning is spammed in console


jakebacker:
![temp](https://cloud.githubusercontent.com/assets/14339279/17792775/d006a1ce-6570-11e6-9b70-7c9cc043796a.PNG)
This picture should explain it.
Notice the 2086s on the right

GamerInspired:
I got that earlier however when trying to recreate it later, I couldn't. If I remember correctly it was the "floor".

jakebacker:
If I just click on the game screen anywhere, it starts happening

NogginBops:
The line is defined here:
![image](https://cloud.githubusercontent.com/assets/3778357/17793059/ef5c1afc-65a4-11e6-9212-b9072845c499.png)

It's nothing serious, it just spams the crap out of the log...

jakebacker:
It _could_ cause slowness in the editor but it is not going to be much slower. Milliseconds if even.

svmnotn:
@jakebacker there is no translation for floor but the CurrentSelection script looks for one

quill18:
My #239 PR added localization for tiles (en_US anyway), which resolves the spam for now. 

I wouldn't just outright remove the logging message. Most likely what should happen is that the warning should be changed to use the new logging ( #185  ) on a low-priority level that won't result in console spam except on the most verbose level.

vogonistic:
Closing in favor of #745

PR #247, opened 2016-08-19
Quieter logging for less log spam

brlodi:
A fix for the general issue in #209, based on the suggestion from @quill18 to use the new Logger's "verbose" output for the failed localization message so it only appears if verbose logging is enabled.

I also took the opportunity to apply a similar treatment to the other messages commonly filling up the console, which come from method-tracing debit messages in `Character.cs`. I think this is a safer short-term solution to those, rather than commenting them out or removing them, so they will still appear if verbose logging is enabled but won't clutter the output during normal game operation. If that's not desired I'm happy to fully disable them instead.

Debug messages in the Character class that were clearly *errors* remain unchanged. I'm willing to tackle converting all the Debug calls in the project to use the new Logger and appropriate levels but I wanted to quiet down the spam quickly first.

### DIFF
--- a/Assets/Scripts/Localization/LocalizationTable.cs
+++ b/Assets/Scripts/Localization/LocalizationTable.cs
@@ -149,7 +149,7 @@ namespace ProjectPorcupine.Localization
             {
 #if UNITY_EDITOR //TODO: Think if #if is a good idea or not.
                 //Log a warning into the console if this operation fails.
-                Debug.LogWarning("Translation for " + key + " failed: Key not in dictionary.");
+                Logger.LogVerbose("Translation for " + key + " failed: Key not in dictionary.");
 #endif
 
                 //Switch the fallback mode.
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -234,13 +234,13 @@ public class Character : IXmlSerializable, ISelectable
             // At this point, the job still requires inventory, but we aren't carrying it!
 
             // Are we standing on a tile with goods that are desired by the job?
-            Debug.Log("Standing on Tile check");
+            Logger.LogVerbose("Standing on Tile check");
             if (CurrTile.inventory != null &&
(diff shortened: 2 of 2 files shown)


Mizipzor:
I really think we should keep the localization logs as warnings. The norm is to never report problems in the `info` or `verbose` level. Problems that the application cant recover from are `error`s. Problems that wont affect it or does so in minor ways are `warning`s, and the localization problems are clearly such a case.

WardBenjamin:
Since #255 replaces all of the calls to Debug.Log with Logger.Log, this probably conflicts. Other than that, looks good. I appreciate seeing much less of the spam :+1:

Mizipzor:
Ah, yes, it will be conflicted, forgot about this one. Its trivial to fix, we'll just have to see which one is merged first (if any).

brlodi:
@Mizipzor in retrospect I think you're right that they should be warnings, but doing that well will require adjusting quite a bit of code. We need to only log a warning the first time instead of *every time* we try to get a particular string lacking localization. I can look into that, and revert this change if people would feel more comfortable.

Mizipzor:
@brlodi I'm willing to fix that. Maybe add it as an issue? I will wait though until the other PR is merged as the fix will conflict with it.

kd7uiy:
Please fix this ASAP, so it can get pulled in to the main build! Thanks!

Mizipzor:
@kd7uiy which part? The thing where we change so that it only logs once for a missing key, rather than every time its accessed?

kd7uiy:
The conflict part. It can't go in until the conflict is fixed.

quill18:
Please check:

https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/CONTRIBUTING.md#resolving-merge-conflicts

Mizipzor:
I think this pull request has been superseded by recent changes to that file. :)

quill18:
@Mizipzor So this should be closed?

brlodi:
I think #362 and #329 contained functionally everything this does. If I make further changes I'll open a new PR.

Issue #20, opened 2016-08-16
List MonoDevelop/Visual Studio settings to produce compatible code


vogonistic:
This came up in PR #19 and should probably go in CONTRIBUTING.md

jokoho48:
Tabs are horrible for github and Tools because every device and tool show tabs in a different way. the most common way to dont get in this issue is using 4 spaces for 1 tab. all teams i work with commit to this "rule"

quill18:
Yup yup -- I'm actually just looking at what happens if MonoDevelop gets reset to default and how closely that parses with the most referenced style guide (https://msdn.microsoft.com/en-us/library/ff926074.aspx), which does suggest have "tabs" map to 4 spaces.

quill18:
Looking to standardize on this (which should be the default in Visual Studios as well):

![screen shot 2016-08-16 at 8 03 22 pm](https://cloud.githubusercontent.com/assets/777633/17719999/920fb534-63ec-11e6-8903-3725f2cd05b0.png)
![screen shot 2016-08-16 at 8 03 36 pm](https://cloud.githubusercontent.com/assets/777633/17719998/920cff6a-63ec-11e6-8f76-0ac7a5fa0c9d.png)

quill18:
Comment now or forever hold your peace. :)

vogonistic:
👍  on converting tabs to 4 spaces.

quill18:
For the record, I still think having braces on the line AFTER the if statement is as ugly as sin.  But I like standards.

Dormanil:
Can we, unlike the referenced style guide, keep strong explicit typing? 

I personally find `var` horribly confusing, especially when it's not about string literals or integer literals.

quill18:
@Dormanil Where's the "million times THIS" upvote on Github?  I agree -- explicit typing adds a LOT to clarity. Doubly so on group projects.

vogonistic:
So the rules so far:
* Microsoft Visual Studio settings
* Convert tabs to spaces, 4 spaces per tab
* Strong explicit typing

I would also add something like this:

If uncertain, follow these guidelines:
* Make your code look like it fits with the code around it
* Make sure you review your diff before creating a pull request

svmnotn:
Any guidelines on whether commits should be squished for a PR?

quill18:
[...]

PR #248, opened 2016-08-19
Could we put open brackets on same line?

Grenkin1988:
Why we need to put open brackets on new line? This just stretches the file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,8 +47,7 @@ interface IInterfaceable { }
 
 // Class names should use PascalCasing.
 // Braces are on a new line. ;)
-class Class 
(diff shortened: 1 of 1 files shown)


NogginBops:
Makes it easier to read and it's Microsofts C# style guidelines, which we have decided on using.

henninglive:
I don't think this up for discussion anymore, see #20. If you disagree you should properly discuss it there instead opening this pull request.

Mizipzor:
No. That part of the style has already been decided upon. Don't get hung up on such a detail, it's much more important that the style guidelines are concise, clear and well defined. _Everyone_ has their preference and there is simply no way to please everyone.

Issue #246, opened 2016-08-19
Problems with searching for inventory to fulfill Job


koosemose:
I'm not entirely sure if this is something recent, or if it only recently became obvious due to the Asteroids PR adding widely spread out plates so it becomes obvious over the long trip, but currently it appears that the Character never recognizes that he already has a path to a tile that contains the inventory that the Job wants, causing a new Path to be generated every frame, and it the slowdown from Pathfinding becomes very obvious.

The problem seems to rely on line 269 of Character.cs:
`if (pathAStar != null && pathAStar.EndTile() != null && pathAStar.EndTile().inventory != null && (pathAStar.EndTile().furniture != null && !(myJob.canTakeFromStockpile == false && pathAStar.EndTile().furniture.IsStockpile() == true)) && pathAStar.EndTile().inventory.objectType == desired.objectType)`
it seems to always return false, I'm currently trying to sort out and figure at what point it fails but that's a very complex conditional, and I'm slightly worried I'll misparse something and bork something else up.

koosemose:
I think I have it sorted out, but it seems to make it all the more convoluted, and I'd appreciate if someone could check my work before I make a PR out of it. 

The problem is that if it's not picking up from a furniture (i.e. a stockpile) then this will return false, and get a new path, which takes a while if it's far. What I've got for now is:
`if (pathAStar != null && pathAStar.EndTile() != null && pathAStar.EndTile().inventory != null && ((pathAStar.EndTile().furniture != null && !(myJob.canTakeFromStockpile == false && pathAStar.EndTile().furniture.IsStockpile() == true)) ||( pathAStar.EndTile().furniture == null && pathAStar.EndTile().inventory.objectType == desired.objectType)))`

vogonistic:
Maybe break it out into a couple of different variables?

```cs
bool hasPath = pathAStar != null && pathAStar.EndTile() != null && pathAStar.EndTile().inventory != null;
bool allowedToTakeFromPathTile = ...
...

if (hasPath && allowedToTakeFromPathTile && ...) {
```

vogonistic:
Make sure you add the new null checks, since they were implied in the long if statement. Now you have to do it in each one.

Another option is to add line breaks and comments
```cs
if (
    // Check that we have an end tile and that it has content.
    pathAStar != null && pathAStar.EndTile() != null && pathAStar.EndTile().inventory != null &&
    // Check if it is a stockpile and we are allowed to grab from it
[...]

PR #250, opened 2016-08-19
Fix for Issue #246 - Character getting new path when EndTile contains no furniture

koosemose:
In CheckForJobMaterials(), when a character already has a path to a tile with the inventory they need for the job, it will get a new path anyways if the EndTile doesn't have a furniture.


kd7uiy:
@quill18 This one should be prioritized, it's really causing some significant slowdowns.

PR #254, opened 2016-08-19
Power system onupdate rework

TheRiven:
Ok lets try this again ---

I have moved the power system check from the furniture update function to its own function so that it is no longer used by ALL the furniture. This function can instead be used in the Lua code for furniture to check if it has power.

This will allow alternative behaviour to be scripted in Lua for if the furniture has power or not.

Once again the only furniture using this at the moment is the oxygen generator.


TheRiven:
hmm, as my github desktop said there was an error (thanks to those pesky font txt files) this should not have been pushed... which is probably why its still showing the old commits.

Mizipzor:
I just tried the github desktop client, is that the one your using? Files can show up as changed, they have a checkbox beside them, you can uncheck the ones you dont want to commit.

It also appear that you are merging in stuff that you dont need. This should either be fast forward merged or rebased.

How about nuking you fork and start over? Copy all source files (that you changed) to desktop or something. Delete the project, delete the fork on github. Refork and download, create a branch again and paste your files. Does that work? I remember doing this _a lot_ when I learned Git. XD

PR #255, opened 2016-08-19
Replaced all Debug.Log with Logger.Log

Mizipzor:
This is the obvious follow up to #185, replacing all log calls with the new Logger.

There's also a ninja fix in here, I realized I was wrong before when I said I matched the method signatures of the built in logger. Turns out, it takes `object` rather than `string` so you can have it implicitly call `ToString()` if you want. That trick was used in [Characters.cs:255](https://github.com/TeamPorcupine/ProjectPorcupine/compare/master...Mizipzor:logger?expand=1#diff-64cc1db7319ecf3cdbc22cea181b644bR256).

No log lines have been added or removed and none have changed level.

This change is obviously rather intrusive, I will do my best to keep up with new commits to keep this conflict free. If there are conflicts, please dont close, I'd rather fix it. :)

### DIFF
--- a/Assets/Scripts/Utilities/Logger.cs
+++ b/Assets/Scripts/Utilities/Logger.cs
@@ -30,7 +30,7 @@ public static class Logger
         }
     }
 
-    public static void LogAssertion(string message)
+    public static void LogAssertion(object message)
     {
         Log(Level.Assertion, message);
     }
@@ -40,7 +40,7 @@ public static class Logger
         Log(Level.Assertion, format, args);
     }
 
-    public static void LogError(string message)
+    public static void LogError(object message)
     {
         Log(Level.Error, message);
     }
@@ -50,7 +50,7 @@ public static class Logger
         Log(Level.Error, format, args);
     }
 
-    public static void LogWarning(string message)
+    public static void LogWarning(object message)
     {
         Log(Level.Warning, message);
     }
@@ -60,7 +60,7 @@ public static class Logger
         Log(Level.Warning, format, args);
     }
 
-    public static void LogInfo(string message)
+    public static void LogInfo(object message)
     {
         Log(Level.Info, message);
     }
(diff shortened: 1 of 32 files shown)


Assets/Scripts/Utilities/Logger.cs
```diff
@@ -30,7 +30,7 @@ public static void LogException(Exception exception)
         }
     }
 
-    public static void LogAssertion(string message)
+    public static void LogAssertion(object message)
```
Grenkin1988: If type is `object`, than parameter should be renamed.
But I think this is bad idea to use objects as Log input.
This is person who use Loger should give appropriate input.

kd7uiy:
@quill18 This one should really be prioritized. It will make a lot of things easier going forward.

Assets/Scripts/Utilities/Logger.cs
Mizipzor: I respectfully disagree. All these methods, their signatures and parameter names, are made so they exactly match the existing Unity `Debug.Log`. Our custom logger will probably do more in the future, but right now I think it's really important that it do exactly this.

kd7uiy: I also agree that object should be used, not string. Requiring a string only forces the user of the log to convert something to a string first. Having it as an object will produce the same thing as calling the .ToString() function, which makes sense.

Mizipzor: It has both actualiy, so you can use whatever you prefer. Again, the point here is not to create the best logger, that we can do later, this is supposed be a carbon copy of Unity's logging. It will be difficult to get this merged as it is. If, and only if, this is merged should we look into going above and beyond.

quill18:
Good work!

Mizipzor:
@quill18 Thank you, means a lot. :)

PR #236, opened 2016-08-19
Power Generator XML and Localization

aleserena:
*Added missing info to Power Gen
*Added EN/US Localization to Power Gen
*Added ES Localization to Airlocj

PR #256, opened 2016-08-19
Updated Localisation for de_DE and en_US

Xeni91:
added power generator Localisation and updated the german localisation


SanteriHetekivi:
@Xeni91 There is already PR for adding power generator localisation (#169) and update for German localisation (#186).

I can not speak for German localisation (only speak Finnish, English and little Russian and Swedish), but both of the PR:s mine (#169) and other German one (#186) already add in missing power generator localisation.

But of course getting these implemented would be great, just pointing out.

SanteriHetekivi:
And there seems to be yet another PR for fixing power generator (#236).
Lets hope that at least one gets implemented.

Xeni91:
Thanks for informing. So it's closed. Because #186 is legit.

PR #257, opened 2016-08-19
Power system - on update rework

TheRiven:
Ok attempt three! --

I have moved the power system check from the furniture update function to its own function so that it is no longer used by ALL the furniture. This function can instead be used in the Lua code for furniture to check if it has power.

This will allow alternative behaviour to be scripted in Lua for if the furniture has power or not.

Once again the only furniture using this at the moment is the oxygen generator.


TheRiven:
Success!  :grinning:

WardBenjamin:
Looks great! Now to make more things that need power! :stuck_out_tongue:

Mizipzor:
Remember the trick if you get stuck some other time, "restarting" in git is easy. :)

TheRiven:
@Mizipzor Will definitely remember that one, it did not help that I was messing around with git before, until it was pointed out I was on the wrong version of unity!

Assets/Scripts/Models/Furniture.cs
```diff
@@ -366,6 +357,28 @@ protected bool DEFAULT__IsValidPosition(Tile t)
         return true;
     }
 
+
+    public bool HasPowerCheck()
+    {
+
+        if (powerValue > 0 && isPowerGenerator == false)
```
Grenkin1988: Maybe easier would be to change `powerValue` and `isPowerGenerator` to something like:
`powerConsumption`
and just play with positive/negative;

for example 
if negative - that this is power supply
if positive - that is consumer

That you do not need: `<PowerGenerator supply="5"/>`

Mizipzor: I would like to see that method renamed to just `HasPower`, the `Check` suffix makes me think that this method does something that has side effects (setting variables).

TheRiven:
@Grenkin1988 That looks like a good change to me, will get on that.
@Mizipzor fair enough, I just used that as it was the first thing that came to mind.

TheRiven:
I have kept the `powerValue` name rather then change it as I feel that fits better then `powerConsumption` for using positive and negative values. Of course this can be changed later :smile:

quill18:
Can you delete GIT_TUTORIAL from your commit please?  Otherwise, looks great!

TheRiven:
Oops missed that

quill18:
<3

PR #258, opened 2016-08-19
Deconstruction yields/spawns steel plates

gunthergun:
A simple addition that spawns  75% of the steel plates it once cost when
you deconstruct the furniture...

This still needs to be fleshed out later so that it works with
other items or that you put in a tag in the xml file that states what
you get back if you deconstruct, but I think for now this is nice :)

please give feedback if you want anything changed :)

### DIFF
--- a/Assets/Scripts/Models/Tile.cs
+++ b/Assets/Scripts/Models/Tile.cs
@@ -123,6 +123,16 @@ public class Tile :IXmlSerializable, ISelectable
             {
 
                 Tile t = World.current.GetTileAt(x_off, y_off);
+
+                Dictionary<string, Job> prototype = World.current.furnitureJobPrototypes;
+                if (prototype.ContainsKey(f.objectType))
+                {
+                    if (prototype[f.objectType].inventoryRequirements != null && prototype[f.objectType].inventoryRequirements.ContainsKey("Steel Plate"))
+                    {
+                        World.current.inventoryManager.PlaceInventory(t, Inventory.New("Steel Plate", 50, (int)(prototype[f.objectType].inventoryRequirements["Steel Plate"].maxStackSize * 0.75f)));
+                    }
+                }
+
                 t.furniture = null;
             }
         }


mmooxy:
What if the furniture gets replaced? You would have to move the items to a different location. Also, couldn't you loop through the keys in the inventoryRequirements to give back the ohter items?

WardBenjamin:
I think customization of drops through XML would be better than just looping through the requirements because it's a lot more customization. Looping through inventoryRequirements isn't a horrible idea for now though.

GamerInspired:
I would also be nice in the future (not right now) to have it spawn different amounts, not just 75% of the resources. For instance between 25% and 75% (example only).

koosemose:
Would it be feasible/desirable to check for a non-empty list of what it deconstructs into, use that if present, and if not default to a portion of the requirements? That way the more common case (I assume) doesn't need to nearly duplicate a list of the same potential items, but something that needs to have different materials can do so still. 

Also, I agree with @GamerInspired , it should (eventually) be a variable amount, possibly (in the far future) based on some skill.

gunthergun:
I like your ideas, I will work tomorrow on multiple resources and looping through the keys list. so that it works with all items. This will be enough for now later we can add xml customisation. The problem with multiple resources is, that you need to search for available tiles. What do you do if there is no available tile ? does the program loop indefinitely if there is no available tile ? It get's a little bit trickier if you consider that as there is no multiple items on one tile system currently... Any Ideas how finding new tiles that are available could be handled ?

RodRitter:
I agree with @koosemose - When we have multiple material types, you'll need to figure out how much of what you want to return. A lot of the configurable data can be turned into properties in the XML.

gunthergun:
I think the way @koosemose stated is the way to go. As of multiple items I have an idea how we could handle this. Instead of randomly spawning them around the deconstructed object, we should let the character fetch it while deconstructing and he should handle taking the resources out. This would mean though this feature would need to wait till we make a deconstruction job. So I will close this pull request and will look into adding a deconstruction job instead :)

Issue #264, opened 2016-08-19
Traveling on open space


kd7uiy:
Just noticed something quite odd when building my first floor, that the worker can travel on spots with no floor. This seems odd. I would suggest permitting it, but have it be slower. It'd be great to have an animation of floundering when this is happening as well, to explain the slowness.

satch3l:
Yes I've noticed this also. I assume it's a bug in the pathfinding code.

BeefEX:
That can be fixed by returning huge cost from empty tiles.

NogginBops:
@satch3l This is not a bug, it's intentional to allow for expansions and building a initial room. It's kind of placeholder for some kind of space system with spacesuits etc. 

Though I like the idea of having the character moveslower on empty tiles.

koosemose:
With the intent being that areas with no tiles are empty space, I'm not sure that slower movement is the right answer. However, the right answer also requires more systems in place, the primary one being a differentiation on if the character is in a space suit or not, and appropriate penalties if they're not in one and in a place with no oxygen (such as empty space) or in a place with no floor traction (this being 2D, it is perhaps best to assume that floor tiles somehow provide gravity). No Oxygen would presumably be them being damage (suffocating), and it would perhaps be nice if there's no floor they just drift (presumably in the direction they were going), if they were mostly in the station they would just drift to the other side, where presumably there's flooring somewhere, but if in empty space they would presumably just drift off until someone rescues them.

But perhaps I'm overthinking things at this stage.

kd7uiy:
The space suit idea is a good thought. For the short term, we should add it as a fairly high movement penalty, say, 5, to discourage such movement.

As for the space suit, if there isn't a discussion about that, we should start one. Or we could use this one I suppose. It needs to be a state of some kind for the character. I could imagine there might be different kinds of suits, so it might be a clothing string or something? Maybe use it for tools that aren't consumed? Not really sure here. Hmmm...

NogginBops:
I think tjat spacesuites should in some way be implemented as items.

koosemose:
I just kind of feel like, at this stage of development, the movement penalty would be more of a hinderance, there would be no way at this point to avoid it, and if I were to, for example, do a test involving picking up all the steel plates that currently spawn on the "asteroids" or to test the mining that someone is doing in a PR, it's just going to take that much longer to actually test it out.

[...]

PR #267, opened 2016-08-19
Restricting character movement to floors.

HexxaGone:
The character will only walk on floors and empty tiles that have at least one floor tile adjacent to them.

This was the easiest way I could think of to still allow placing floors.

This does mean the character will still 'jump' over gaps up to 2 tiles wide.

### DIFF
--- a/Assets/Scripts/Models/Tile.cs
+++ b/Assets/Scripts/Models/Tile.cs
@@ -85,6 +85,20 @@ public class Tile :IXmlSerializable, ISelectable
             //if (Type == TileType.Empty)
             //    return 0;	// 0 is unwalkable
 
+            if(Type == TileType.Empty)
+            {
+                Tile[] ns = GetNeighbours();
+
+                bool canMove = false;
+
+                foreach (Tile n in ns) // Loop through all the horizontal/vertical neighbours of the empty tile.
+                {
+                    canMove = canMove || (n != null && n.Type == TileType.Floor); // If the neighbour is a floor tile, set canMove to true.
+                }
+
+                return canMove ? baseTileMovementCost : 0f; // If canMove is true, return baseTileMovementCost, else, return 0f.
+            }
+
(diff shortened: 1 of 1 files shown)


BeefEX:
Very good solution.

WardBenjamin:
I like this for now, as far as the discussion in #264 goes. This probably shouldn't be a permanent solution though.

BeefEX:
@WardBenjamin true

Mizipzor:
A movement cost of zero feels backwards, the path finding should find the path with the lowest cost. Unless it has a special case for zero cost. I would suggest creating a new property called something like ´IsWalkable´.

koosemose:
I still think this is a bad idea it adds extra difficulty to testing things outside of the station proper (such as having to built a bridge out to an asteroid) with no way to bypass it such as a spacesuit.

Mizipzor:
Maybe add button to spawn an asteroid wherever you want? Like directly adjacent to the base.

Assets/Scripts/Models/Tile.cs
```diff
@@ -85,6 +85,20 @@ public float movementCost
             //if (Type == TileType.Empty)
```
OzieGamma: Why do you keep old code ?

Assets/Scripts/Models/Tile.cs
```diff
@@ -85,6 +85,20 @@ public float movementCost
...
             //    return 0;	// 0 is unwalkable
 
+            if(Type == TileType.Empty)
+            {
+                Tile[] ns = GetNeighbours();
```
OzieGamma: `Tile[] tiles = ...` is more readable. I had some trouble reading

vogonistic:
This reveals a fun bug in the build system where if it can't path find, it builds anyway, but that's not the problem with this bug. Accepting.

PR #253, opened 2016-08-19
Added Cloning pod

MANDAL0R3:
Creates characters at a high cost of steel, and a long build and work
time of 10 seconds.
Once job is complete, the cloning pod is immediately destroyed(maybe
just damaged).
NOTE: after 4 characters building together, lag starts to occur.

screenshot1
screenshot2

Here is the link to my last pull request before i figured reverted and removed the resource files changes.
https://github.com/TeamPorcupine/ProjectPorcupine/pull/243

PR #270, opened 2016-08-19
EventSystem (New crew member, pirates, etc)

relinu:
Adds EventSystem, events can be created in the StreamingAssets/GameEvents folder as xml code. (e.g. attacks from space pirates, meteors, a new crew member,  merchants)

### DIFF
--- a/Assets/Scripts/Controllers/GameEventManager.cs
+++ b/Assets/Scripts/Controllers/GameEventManager.cs
@@ -22,7 +22,7 @@ public class GameEventManager : MonoBehaviour {
     /// </summary>
     void Update(){
         foreach(GameEvent gameEvent in gameEvents.Values){
-            gameEvent.Update();
+            gameEvent.Update(Time.deltaTime);
         }
     }
 
@@ -89,6 +89,7 @@ public class GameEventManager : MonoBehaviour {
     {
         //Debug.Log("ReadSpriteFromXml");
         string name = reader.GetAttribute("Name");
+        int repeat = int.Parse(reader.GetAttribute("Repeat"));
 
         List<string> preconditionNames = new List<string>();
         List<string> onExecuteNames = new List<string>();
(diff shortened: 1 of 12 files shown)


Assets/Scripts/Controllers/GameEventManager.cs
```diff
@@ -0,0 +1,104 @@
...
+            CreateEvent(name, functionNames.ToArray());
+        }
+    }
+
+    void CreateEvent(string eventName, string[] functionNames)
+    {
+        GameEvent gameEvent = new GameEvent(eventName);
+
+        foreach(string funcName in functionNames){
+            gameEvent.RegisterExecutionAction(funcName);
```
NogginBops: Maybe a AddAll method?

bjubes:
How are events present to the player? Is their a corresponding pop up that warns the player kinda like rimworld? Or do they just happen in real time without warning?

relinu:
This can be set in the lua code in the future maybe there you can call a method that display a mesaage on text or changes something in realtime like adding a pirate at the dge of the world that runs to your base and destroys it (like in rimworld)

relinu:
Lua code is added

relinu:
screenshot06

Changed event to spawn a new character every 30 seconds... Quiet funny with that many little guys :D

henninglive:
The spawn a new character event does kind of conflict with #253, which also way of adding new characters.

quill18:
Merged!

ComputerOverlord:
@quill18 I would have used #253 But, this is fine.

relinu:
@ComputerOverlord we can use that too, the current event is only a example of how one can create a event and will definitely be changed in the future

henninglive:
I would prefer we remove the spawn a new character event sooner rather than later. The game begins to lag quite quickly as they multiply.

MANDAL0R3:
This is already merged but might still be a nice place to discuss this. The cloning pod I created might not indefinitely be used to create new crew members. As of right now im leaving it as it is because it works and adds a player controlled way to add new crew members. I have been thinking that instead of creating new crew members, it instead is used as a one time respawn that has to be assigned to a character before being damaged and inoperable.

vogonistic:
@MANDAL0R3 There is an issue open to discuss how characters/crew members are supposed to be created #283. Can you open a PR that fixes the error in `Character.cs`?

PR #271, opened 2016-08-19
Added keyboard reaction

FilippoLeon:
This is a new script that shoul be added, I suggest, to a new controller named "Camera Controller" or something (that is supposed to be extended). This controller should handle all the main camera movements. This is a super small contribution to break the ice with the project.

What can also be added are keys for controlling time (once implemented), for zoom, etc.

Also: tab to switch character?

### DIFF
--- a/Assets/Scripts/Controllers/CameraController.cs
+++ b/Assets/Scripts/Controllers/CameraController.cs
@@ -5,16 +5,17 @@ public class CameraController : MonoBehaviour {
 
     [Range(0, 3)]
     public float scrollSpeed = 0.1f;
-    
-    void Start () {
 	
-	}
-	
-	void Update () {
+    void Update () {
         
         // React to hor./vert. axis (WASD or up/down/...)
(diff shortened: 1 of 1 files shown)


Mizipzor:
You should probably configure MonoDevelop to follow the code style, details in [CONTRIBUTING.md](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/CONTRIBUTING.md). Also, doesnt the camera already do this?

FilippoLeon:
Mhhh, I'm using VS15, so it should already be following the project's conf. 

But I see there is something wonky with tabs/spaces.

AFAIK the camera doensn't respont to WASD.

bjubes:
I think its adding keyboard controls to move the camera, so u can move it if you have a laptop and no mouse making left click drag difficult. However I don't think this requires a seperate script

WardBenjamin:
As of the current version, camera movement with WASD/Arrow Keys is either not implemented or broken.

FilippoLeon:
@bjubes I know that there is no need for a separate script for a one-liner.

But I suggest the following: use this script as base to handle other keyboard presses, like speed change (when implemented), character change (e.g. with tab), ...

PR #279, opened 2016-08-19
Can log any object now, using the object's ToString() function.

kd7uiy:
This will allow for ease of use in complete logging structures, running .ToString() is managed in the logger, as Unity's Debug.Log does.


WardBenjamin:
Already implemented in #255, btw.

Mizipzor:
Good catch though. :)

PR #288, opened 2016-08-19
Polish localization

arrowgroundgames:
Just like in the title - I made a polish translation :)


relinu:
I think pl_PL.lang has eol-issues like the license files

PR #290, opened 2016-08-19
Make floor building update rooms

SquareMan:
Rooms will now be correctly updated when a floor is built, whereas before only furniture building would trigger this


kd7uiy:
This one can't be accepted, due to some extra files you've changed. Still, the change looks good so far! Fix the merge artifacts, and let's get this pushed forward!

WardBenjamin:
@SquareMan Looks great other than the changes that @kd7uiy mentioned. 

To fix the merge artifacts and the extra commits, you can either copy the changes out of the repository, onto your desktop or something, re-clone or reset the repository, and then move them back in and commit them, or you can "cherry pick" the commit onto a new branch. Guides to cherry-picking can be found online.

PR #291, opened 2016-08-19
Lots of names changed according to selected naming standarts

Grenkin1988:
Lots of names changed according to selected naming standards
Changes in Model directory mainly
Other files are affected.

### DIFF
--- a/Assets/Scripts/Controllers/BuildModeController.cs
+++ b/Assets/Scripts/Controllers/BuildModeController.cs
@@ -33,7 +33,7 @@ public class BuildModeController : MonoBehaviour
             return true;
         }
 
-        Furniture proto = WorldController.Instance.world.furniturePrototypes[buildModeObjectType];
+        Furniture proto = WorldController.Instance.world.FurniturePrototypes[buildModeObjectType];
 
         return proto.Width == 1 && proto.Height == 1;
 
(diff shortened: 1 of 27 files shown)


vogonistic:
I don't think we can go about changing it this way. It'll break the 50 (minus translations) other PR's. I suggest we make the change gradually as we improve the different systems.

Mizipzor:
I thought the decision on style wasnt made yet.

behindcurtain3:
Ultimately up to Quill whether he wants to accept all the changes in one PR. The sooner everything is standardized the better though as the code base will just continue to grow.

Grenkin1988:
This need to be done at some point. And I worked only files from Model.
There are lots of new pulles that to not follow standart listen in Contributing

vogonistic:
@behindcurtain3 That's true, and I wasn't implying that it is my choice either. I would take the other PR's with features over a single one that just adjusts things to be a little neater every day.

Mizipzor:
This is also incorrect, assuming we do follow [Microsoft General Naming Convention](https://msdn.microsoft.com/en-us/library/ms229045(v=vs.110).aspx) which says both:

> X DO NOT use **underscores**, hyphens, or any other nonalphanumeric characters.

And

> X DO NOT use Hungarian notation.

And Hungarian notation is, according to wikipedia:

> Hungarian notation is an identifier naming convention in computer programming, in which the name of a variable or function indicates its type or **intended use**. 

I don't recall we settled on the exact wording for the underscore rule, which is another reason to just avoid it, but if it is supposed to show how to field should be used it does not conform with the Microsoft style.

vogonistic:
@Grenkin1988 I agree that it should be done, but I think it's better to do it in a less breaking manner. When the code standards have settled for a bit, we can require new PR's to follow them explicitly and do it over time or in a couple of chunks.

Mizipzor:
I realized I might come across as negative again, need more smilys. :)

The point of all this is to pick  a well defined style and stick to it rigidly. Either change the contribution guidelines to link to another style (please dont make a custom one) or change this pull request to adhere to it.

So to say something positive, there a a few lines where hungarian notation has been _removed_, some callback variables has a `cb` prefix. So on that note, well done.

behindcurtain3:
@vogonistic You bring some good points, there are some interesting PR with new features already waiting. I'm not sure if the short-term value is greater in getting a standardized code base or adding features.

Long-term standardized is the goal :D

Grenkin1988:
Whether would this be merged or not, I made a try.
I am rejected several times a day due to code styles, so this is just my habit.

Mizipzor:
@Grenkin1988 your efforts are appreciated and your intention is good. But you are going about this the wrong way, there's still discussions ongoing regarding the style because the style leaves stuff open to interpretation. Instead of jumping directly to fixing the code, focus on really hammering down the style, there should be no confusion as to what is right before we start changing code. The fact that you are "rejected several times a day due to code styles" is a strong indication that the style is not clear and well defined.

Grenkin1988:
Do we have one code standart desscision?

Mizipzor:
The [contributing](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/CONTRIBUTING.md) file links to two Microsoft standards. As far as I know we follow these with the exception that we never use implicit typing (`var` keyword). Right now theres disagreement regarding the underscore character, which I argue is against those styles as outlined above.

PR #293, opened 2016-08-19
Jobs now have a priority

kd7uiy:
This  is a simple start of improving jobs. All jobs now have a priority. Stockpile is set to the lowest, mining to the second lowest, and all building set to high. This should allow for buildings to be prioritized.


kd7uiy:
Note: I had previously added logging which I removed. This will not be included when it's squashed in to the big build.

WardBenjamin:
Looks great, and commits should be fine once it's squashed.

vogonistic:
Does this work as build wall jobs gets abandoned and put back on the queue because there is no steel to be had? It seems like as they get put back on the queue, they end up being higher priority than mining plates. I can't test right now and just skimmed the code.

ComputerOverlord:
Custom priorities menu?

kd7uiy:
I've set it up now so that new jobs are at the end of the queue, as it should be (They were added to the beginning of the "priority" queue. It could easily be changed.

vogonistic:
I assume you've tested what happens when you run out of raw materials and need to operate the mining drone to get more?

kd7uiy:
@vogonistic I think you are right. I think deferred jobs should be put at medium priority. It's surprisingly hard to test.

vogonistic:
Medium priority isn't going to fix it. If you requeue jobs at lowest priority, it'll work itself out, but be slow and not look any better. It's the kind of problem that leads to people working on better job systems. =)

One way you can approach it is to add more logic before someone can Dequeue a job and check with the InventoryManager if it knows of those resources anywhere. Picking things out of order means you'll have to switch it to a List though.

kd7uiy:
I've set it up to drop the priority a level. I agree your proposed change would be better, I just don't have the time to work on it for now... This should prevent us from getting completely stuck, and overall look better.

WardBenjamin:
One possible solution to this is to pair it with an [HTN](https://en.wikipedia.org/wiki/Hierarchical_task_network), so that you would have a priority queue for each Goal (a meta-job, filled with smaller jobs that depend on one another), and then another queue internal to each Goal to order the jobs. If the Goal becomes [gridlocked](https://en.wikipedia.org/wiki/Gridlock), you could either bump it down in the Goal queue or notify the end-user, or both.

vogonistic:
@WardBenjamin I might be wrong, but I think @kd7uiy did this to fix something small that bothers him now while we wait for a HTN or something better to be built.

WardBenjamin:
@vogonistic I mean, this definitely works for now, yeah. Just pointing out some future routes. I'll repost that info in the issue.

kd7uiy:
This is a medium quality fix, it's better than what's there now, but still far from perfect.

kd7uiy:
For future reference:

High priority- Building activities
Medium Priority- Good producing activities
Low priority- Organizing activities

I think this is a good rule of thumb no matter what future system we do. Making goods should always be higher priority than organizing, and building should always be higher than making. There might be a need for an even higher priority for emergencies, which I did not add but could easily be added, to do stuff like patch a hole in a wall, for instance.

kd7uiy:
gif

vogonistic:
Nice, what did you use to create the gif?

WardBenjamin:
Looks great!

kd7uiy:
I used GifCam. http://blog.bahraniapps.com/gifcam/

Note that building takes priority. Making takes priority over stockpiling. The current job is finished before the next one is taken. I would let this go until all of the resources are gone, but that takes a VERY long time, so I can't prove that everything will work in that condition, yet.

quill18:
Looks good -- just waiting on conflict resolution before merging.

kd7uiy:
@quill18 I think I've got it fixed, tested it and it still works!

quill18:
While you include many changes that should happen, such as logging to any object as well as generating the meta files for language entries, in the future take care to ONLY include changes relevant to one feature in each PR.  We will start to be quite strict and focused about this going forward.

Nonethless, merged.

kd7uiy:
Thanks, I didn't actually look at the list of what I produced. Strange that the meta files weren't included previously... Logging to any object should have already been done, I had an old version that I removed when I realized there was a pending version already... Anyways, thanks!

Issue #164, opened 2016-08-18
Auto resizing UI-elements or fonts for future translations. (Low priority)


SanteriHetekivi:
I was just testing the localization system, with Finnish and noticed that some of the text got cut off.

**English**
![image](https://cloud.githubusercontent.com/assets/9454417/17779702/3f0bac40-6571-11e6-8178-85892824964b.png)

**Finnish**
![image](https://cloud.githubusercontent.com/assets/9454417/17779727/589981aa-6571-11e6-96c0-ea83de85bd2b.png)

**Spanish**
![image](https://cloud.githubusercontent.com/assets/9454417/17779678/2a8384aa-6571-11e6-9c4f-8e111ed347f0.png)

This is low priority because right now whole localization process is low priority (at least in my mind).
But for the future auto resizing elements or fonts would be useful.

SanteriHetekivi:
Just noticed that #78 Power Generator does not have its name set in button. I'm fixing it now.

emilkris33:
@SanteriHetekivi fixing the power gen bottom is a good idea. 
Though localization compatible buttons is something whoever eventually redoes the UI should keep in mind.

SanteriHetekivi:
Yeah as a back-end programmer UI is not my strongest suit, but if someone who has some extra time would at some point make auto resizing buttons, that would be great.

Button size starts to be an issue even without the localization, the "Build Power Generator" text, that I added on pull request (#169), barely fits the button.

tuefekci:
Sorry i hijack this a bit to get more Information about where the GUI should go.

I never worked with the new GUI, is there a way to handle stuff in the new GUI like you would with IMGUI? The Unity Reference does not really tells or i am to stupid to find ;) 
It should not be really hard to resize to fonts even in Canvas mode to size the buttons but if the new GUI is not script able its anyway the wrong direction to go in the long run.

I´m Coming from a more Frontend related Background and would like to have a proper purpose to get back in Unity, so i probably would like doing GUI stuff on a public project.

PR #296, opened 2016-08-19
Add localization text rescaling

Fogsight:
Automatically scales any text and its text box to fit completely within its anchors.


Fogsight:
localization text rescaling

WardBenjamin:
Looks great!

Mizipzor:
Nice. I would've assumed that you had to resize the buttons, and risk screwing up the layout, but reducing the size of the font is a great compromise.

SanteriHetekivi:
Great this fixes problem I stated in issue #164. :)

Fogsight:
@Mizipzor, yes, but actually without resizing the font. Main reason I wanted to do it this way is to avoid "Best Fit" growing atlas out of control as described here: https://twitter.com/pgmuscat/status/763164207495933952

PR #297, opened 2016-08-19
Power Generator Localisation

Dan54:
Adds English localisation for Power Generator.
localisation screen shot2


WardBenjamin:
This is already completed in #236

SanteriHetekivi:
@Dan54 @WardBenjamin And also in #169, #186 and #256 
Seems this fix is popular :)

WardBenjamin:
@SanteriHetekivi Yeah, I think Quill mentioned it in a video or something.

SanteriHetekivi:
@WardBenjamin Jeah my #169 was first to fix it before video went live (Noticed while making Finnish translation) and my comment in the video about the fix being already made is top comment, but still these keep coming.
Good that people are at least active :)

PR #311, opened 2016-08-19
Tell .gitattributes that .txt files are text.

quill18:
Attempts to fix some end-of-line issues that were causing problems for everyone.


henninglive:
Doesn't work i have tried this, see #187 and #143.

PR #287, opened 2016-08-19
License headers 

WardBenjamin:
This pull request is a WIP, do not merge.

Right now, this removes Quill's old license headers. I propose replacing them with something like this (modeled off of the GPL header thing at the bottom of the LICENSE file):

```
// ====================================================
// Project Porcupine Copyright(C) 2016 Team Porcupine
// This program comes with ABSOLUTELY NO WARRANTY; This is free software, 
// and you are welcome to redistribute it under certain conditions; See 
// file LICENSE, which is part of this source code package, for details.
// ====================================================
```

Opinions? Any modifications that should be made?

PR #314, opened 2016-08-19
RE: [Beginner] Prevent furniture from being built along the edges of the world. Fixes #300

### LINKED ISSUE #300: [Beginner] Prevent furniture from being built along the edges of the world
Currently, you can build walls on the last tiles on the edge of the world, but it leads to a lot of checks for tiles outside the map. One common way to fix this problem is to not let the user target the tiles on the very edge of the map. 

Edit `DEFAULT__IsValidPosition()` in `Assets/Scripts/Models/Furniture.cs` so that it doesn't let the user build or place anything within 5 tiles to the edge.

Pay attention to the rules in [CONTRIBUTING.md](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/CONTRIBUTING.md)

This bug was originally reported in #294

lincore81:
In response to this issue: https://github.com/TeamPorcupine/ProjectPorcupine/issues/300
I hope my style is acceptable.


WardBenjamin:
Looks good other than the line changes on the first and last lines. The change on the first line will prevent #287 from being merged. Fix that, and then this would be great! :+1:

vogonistic:
Looks good to me! You can also include the words "Fixes #300" in the message, and it will automatically close the bug when this is merged.

It's pretty much traditional to have to make an additional change, so expect it.

vogonistic:
@WardBenjamin Has a good point on the first line, that should be fixed.

The last line I think can stay. The way most editors are configured, they'll automatically fix the lack of a new line at the end of the file if it's missing, so adding it now prevents it from happening again next time someone edits it.

lincore81:
Ok. I guess the line endings are messed up, I'll have to look into this.

lincore81:
I re-encoded the file with bom, should be fine now. Would it not be a good idea to standardize the charset+eol?

quill18:
@lincore81 CONTRIBUTING.md details our eol standardization and how to follow it.

Good change -- merging!

Issue #295, opened 2016-08-19
Discussion: overlays and "tile data"


FilippoLeon:
I think the project could benefit of having a series of overlays displaying "color maps" (or heat maps) of many variables that depend on tiles:
- display a color map of the oxygen level
- colorize each room separately
- display temperature, if implemented
- display other variables

I've created a branch:
https://github.com/TeamPorcupine/ProjectPorcupine/compare/master...FilippoLeon:overlay
with some dummy code.

If you want to test, you have to add an object and add the script "OverlayMap" to it.

Mandatory picture:
![overlay_example](https://cloud.githubusercontent.com/assets/21125139/17817586/84c578e6-663f-11e6-9603-1d52cdcc2f65.png)

Let me know if there is interest in this.

(I'm also interested in modelling some variable: like temperature, etc... but I'll open another issue)

WardBenjamin:
I was actually just thinking about this, looks awesome. Overlays like oxygen (maybe a generic "gas" one with a dropdown for any other gases?), temperature, and such would be great.

koosemose:
This seems pretty useful, both to in development and gameplay. Though if what's in the picture is rooms, perhaps room 0 shouldn't have an overlay (or rather have a completely transparent color), since the idea is that it's not a room.

FilippoLeon:
Sorry, picture shows oxygen, but is still a bit wonky.

koosemose:
Ah, was just guessing. Do you intend (once you have it dewonked) to make this accessible from in game, or do you intend it to be a developer only thing that one would control through the object or some other method to be used for debugging?

WardBenjamin:
I think this would be a good in-game tool, if refined enough. I think it would be really useful when doing things like planning where oxygen generators or CO2 scrubbers go, or other use cases. This would also be useful if we added gas piping and other stuff.

FilippoLeon:
Also for end-user. 

For instance. Many times I wished that RimWorld had such a feature for the temperature, so that my colonists woldn't burn alive.

koosemose:
Excellent, lots of fun potential uses for this, such as spotting leaks (presumably a part of mapping gas flows). 

Any idea on the feasibility of making this moddable (interact with LUA and/or xml files) so an end-user could mod in their own heatmap for something they're really interested in?

FilippoLeon:
I will proceed with improvements then.

[...]

PR #316, opened 2016-08-19
Overlay map

FilippoLeon:
So, I worked a bit on a overlay for the map:
https://github.com/TeamPorcupine/ProjectPorcupine/issues/295

Especially for developement, but also for end-user.

Still a bit wonky, but should be useful for devs.

Currently I have maps for room id, power and oxygen.
Configurable trough lua.

Unfortunately I had to tweak the unity scene.

Comments appreciated.


WardBenjamin:
Looks fantastic!

TheRiven:
Like the power overlay! But just to let you know the `isPowerGenerator` check is being removed  (assuming my update gets approved) in favour of a positive for generators or negative for consumers `powerValue`

FilippoLeon:
Good, I was just about to ask why we had that flag.

PS: if somebody has ideas for other overlays, I'll be happy to see if I can work on it.

satch3l:
Any screenshots?

bjubes:
@satch3l  there is one in the Issue request:
screenshot-from-295

FilippoLeon:
If you scroll down there is the updated system:
https://github.com/TeamPorcupine/ProjectPorcupine/issues/295

Assets/Scripts/UI/Overlay/OverlayMap.cs
```diff
@@ -0,0 +1,486 @@
...
+
+    /// <summary>
+    /// Script with user-defined lua valueAt functions
+    /// </summary>
+    Script script;
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public string currentOverlay;
```
Grenkin1988: Correct if I am wrong. There should be no public fields in class. Only public properties. (Encapsulation)

FilippoLeon: If it's not a convention that the project uses (tell me if so), then no. By all means: properties are quite nice, but I wouldn't abuse them.

henninglive:
Properties and public members should use PascalCasing.

Assets/Scripts/UI/Overlay/OverlayMap.cs
Grenkin1988: As I understand regarding public fields is just some kind Unity weird thing.

quill18:
Overlays are good! Looking forward to see how much we can do with this.

PR #323, opened 2016-08-19
Added support for executing jobs on an adjacent tile.

Mizipzor:
The character should walk to the tile, but stop beside it rather than on it.
This currently only happens when constructing floors.
Implemented by essentially removing the last tile from the found path.


quill18:
Long overdue. Thanks!

PR #329, opened 2016-08-19
Change several very common logs to Verbose to reduce log spam

kd7uiy:
There are a number of common logging functions that slow things down drastically. These are switched to Verbose, so they normally don't show unless requested.

### DIFF
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -259,13 +259,13 @@ public class Character : IXmlSerializable, ISelectable
             // At this point, the job still requires inventory, but we aren't carrying it!
 
             // Are we standing on a tile with goods that are desired by the job?
-            Logger.Log("Standing on Tile check");
+            Logger.LogVerbose("Standing on Tile check");
             if (CurrTile.inventory != null &&
                 myJob.DesiresInventoryType(CurrTile.inventory) > 0 &&
                 (myJob.canTakeFromStockpile || CurrTile.furniture == null || CurrTile.furniture.IsStockpile() == false))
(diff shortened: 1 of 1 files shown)


Mizipzor:
Hm it appears some job bug fixes has snuck in here.  Quill mentioned that we should be more strict about only accepting pull requests that do what they say they do and nothing more. He might still go easy on it but I would suggest looking into splitting this up. :)

kd7uiy:
If you look at the files changed, there's only actually 1 file that has changed, and it has nothing to do with logs. Basically, I need to clean up my git fork, which I can do, but it's somewhat painful. Bottom line, this should work as is, when it's squashed only the change indicated will be included. And there won't be any difference shown in the commit tree.

kd7uiy:
This is related to https://github.com/TeamPorcupine/ProjectPorcupine/pull/247 and in fact, I believe https://github.com/TeamPorcupine/ProjectPorcupine/pull/247 will contain everything this does after it's conflict is resolved. So if https://github.com/TeamPorcupine/ProjectPorcupine/pull/247 is pulled in, just reject this one.

kd7uiy:
Okay, I've done my git work, and now I'm just commiting the desired change, plus a merge. It still has the same effect.

Mizipzor:
@kd7uiy is awesome that you figured it out. Did you do this with a force push our just recreated fork? Regardless, this way of replacing committs in a pull request is super useful.

kd7uiy:
Force push and reading. It's a real pain...

Mizipzor:
@kd7uiy its not pain, its fun! ^^

Issue #331, opened 2016-08-19
Multi-character optimization


ComputerOverlord:
We know from #270 #253 that many characters slows down the game

kd7uiy:
Some of this is log spam, which will be reduced when my latest PR gets pulled in to the fray. That won't fix all of it, unfortunately. I think pathfinding is the major source of slowdown, in particular the fact that it seems to be re-done every turn. There's an issue report somewhere here about that.

Bottom line, this is closely related to the following:

https://github.com/TeamPorcupine/ProjectPorcupine/pull/329

https://github.com/TeamPorcupine/ProjectPorcupine/issues/246

henninglive:
Maybe disable auto spawning of new characters for now.

koosemose:
Yeah PR #250 fixes one of the major causes of this, though until quill is able to get a team that can accept PRs it may take a bit before it comes to his attention and is accepted.

relinu:
I will limit the amount of repeats a gameevent can have

kd7uiy:
I just pinged quill to get him to prioritize #250 , so let's hope that works. It might help if others join in that? Also, we should prioritize the various logspam ones, which will help.

Finally, the maximum number of characters should probably be re-thought, or at least, how often they spawn.

koosemose:
#250 has been committed, take a look to see if that fixes, or at least reduces the issues. My GitHub Desktop is having fits over some files so I can't at the moment.

kd7uiy:
Sadly I still see lots of log spam... Sigh. Just the A* pathfinding, but...

kd7uiy:
Confirmed the reduced log spam!

koosemose:
Well, it's better than it was, but there's still a massive slowdown anytime a long-distance pathfinding is done, which is going to happen more often the more characters there are. Pathfinding code certainly needs optimized.

vogonistic:
There has been several improvements for this, but it's going to be an ongoing project. I suggest we add issues for specific things that slows the characters down so we can have a criteria to finish issues.

PR #334, opened 2016-08-20
Limits the repeats a GameEvent has

relinu:
Should temporarily fix  the issue with Characters ( #331  ) until better pathfinding is in place


WardBenjamin:
Good change! This is pretty cool since I see a lot of applications, for example you could use it to to have a defined number of meteors in a meteor event, or something like that.

henninglive:
Is it possible to make `maxRepeats` an optional parameter? Maybe combine `maxRepeats` with `repeat`?

vogonistic:
I'm not sure this is the right way to do it, but it gives us time to talk about it.

PR #338, opened 2016-08-20
Make Metal Smelter consistent with new Job prototype

kd7uiy:
Metal Smelter needs to include a priority. Set it at medium.

### DIFF
--- a/Assets/StreamingAssets/LUA/Furniture.lua
+++ b/Assets/StreamingAssets/LUA/Furniture.lua
@@ -232,6 +232,7 @@ function MetalSmelter_UpdateAction(furniture, deltaTime)
 			nil,
 			0.4,
 			itemsDesired,
+			Job.JobPriority.Medium,
 			false
 			)
 			


kd7uiy:
This is a one line of code fix, to not break the Lua code using the Job Creation change made previously.

WardBenjamin:
@kd7uiy Btw, there's an extraneous logging-related commit again. One way to maybe avoid this is by working on different feature branches instead of on master all of the time.

kd7uiy:
Yeah, I should really start doing that... I think I'm going to in the future, thanks for the tip!

kd7uiy:
Now it's just two commits, the original and a merge. I'm content with that:-)

WardBenjamin:
Looks much better!

kd7uiy:
This should really be pulled in as soon as possible, @quill18 . It's only one line of code, and without it there's an annoying error that happens, plus the smelters don't work at all.

henninglive:
What's up with the `pl_PL.lang` file? did you intend to modify it?

kd7uiy:
I'm trying to get rid of that one as we speak... Sigh...

kd7uiy:
@henninglive Got it cleared up. I tried to create a branch to put the other change in, and it didn't work as I expected it to. Bottom line, I've got it resolved now, and will ignore the other change for the time being.

quill18:
It looks like pl_PL.lang was committed wonky eol's. I expect a lot of `git add -A` or Github Desktop users will be including it in their commits.

Issue #31, opened 2016-08-17
Implementing a generalized type for furnitures


koosemose:
Certain features will need to care about the general type of a furniture, such as the recently committed replaceable furniture, as it stands it requires each specific furniture to be named, which works for now, but in the future, particularly for modding, it can become overly complex, if not completely unworkable. As an example, if a mod implements a new wall type, and a second mod adds a new door type, the new door would presumably list the current wall as something it can replace, but would be unable to replace the modded wall. 

Now if there were a generalized type for furnitures, such as "Wall" or "Door" rather than "Steel Wall' or "Basic Door", then each new wall and door could simply be told that it is a wall or a door, and if one mod adds a wall, it will be typed as a "Wall", and a mod adding a door (or even the vanilla doors) would simply be told that they can replace any furniture that is a "Wall", and it would just work.

It could also be of potential use for other features as well, if one were to implement a feature akin to Rimworld with raiders breaking down doors to get in or possibly even walls, they could be told to break any objects of types "Door" or "Wall" rather than either breaking any furniture in their way, or having to base the decision on the movement cost/passability of a given furniture.

vogonistic:
(I've only watched the first 28 videos, so this might not reflect the current code base)

An alternative way of solving some of this would be to have a deconstruction cost and what the output of deconstruction is. That way, it doesn't matter if you replace a chair with a door, or a wall with a door. If it can be deconstructed, it can be replaced.

There there might be a good argument for a wall type in case we want to group elements in the UI by type, but naming everything that it can replace might be hard to maintain.

koosemose:
The intention of this proposal is to minimize having to list every single object you can replace (and any other future features that may care about what  types of objects they are interacting with), so rather than listing every wall there is (which as I mentioned won't works so well with mods), but rather just the type, so that you would just list "Wall". The replacing furniture code is already in, I'm just proposing a way to make it (and any future features that care about object type) easier to maintain.

And if a limited set of object groups is established that would further ease maintenance related to this feature, for example "Door", "Wall", "Workstation", and "Decoration" would cover a fair number of needs, perhaps "Chair" and "Table" as well so that future AI could be told to try to find a table to eat at, and find a chair around that table to sit on without having to tell the AI every table and chair that there is (which would likely not work well with modded in tables and chairs).

[...]

PR #342, opened 2016-08-20
Change BaseType to TypeTag, and allow multiples

koosemose:
Per further discussion in #31, BaseType was changed to TypeTag (to more closely match new functionality), and now allows multiples (previously it would be set to the last one found). Added bool HasTypeTag(String) method to separate usage from implementation (future-proofing for any potential changes in the future).

### DIFF
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -184,7 +184,7 @@ public class Furniture : IXmlSerializable, ISelectable
     {
         this.objectType = other.objectType;
         this.Name = other.Name;
-        this.typeTags = other.typeTags;
+        this.typeTags = new HashSet<string>(other.typeTags);
         this.Description = other.Description;
         this.movementCost = other.movementCost;
         this.roomEnclosure = other.roomEnclosure;
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -472,7 +472,7 @@ public class Furniture : IXmlSerializable, ISelectable
                     roomEnclosure = reader.ReadContentAsBoolean();
                     break;
                 case "CanReplaceFurniture":
-                    replaceableFurniture.Add(reader.GetAttribute("baseType").ToString());
+                    replaceableFurniture.Add(reader.GetAttribute("typeTag").ToString());
                     break;
                 case "DragType":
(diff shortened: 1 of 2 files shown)


WardBenjamin:
Looks great. I think this a good solution that leaves modding and new items open.

koosemose:
Thanks, had an extended discussion with @lameox that helped further refine the original idea. It also has the added side-effect of things not erroring out if trying to replace a furniture that doesn't have a typeTag (or anything else that wants to look for a match)

Assets/Scripts/Models/Furniture.cs
```diff
@@ -183,7 +184,7 @@ protected Furniture(Furniture other)
     {
         this.objectType = other.objectType;
         this.Name = other.Name;
-        this.baseType = other.baseType;
+        this.typeTags = other.typeTags;
```
henninglive: Is this correct? shouldn't you copy the hashset?

koosemose:
You're probably right, I overlooked that, and even though that works in the current implementation since new typeTags are only added to the prototype when it's created, I'll get that sorted out in case someone later adds a way to directly modify a Furniture's typeTags (such as maybe a way to mark a Furniture as broken)

koosemose:
Alright, that should fix any potential problems.

henninglive:
When is the copy constructor called anyway? what its purpose?

koosemose:
When you're creating an actual furniture from a prototype. In case you're not familiar with the setup, it reads the values from XML into a prototype furniture, then when an actual piece of Furniture is being built it uses the copies the prototype.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -183,7 +184,7 @@ protected Furniture(Furniture other)
     {
         this.objectType = other.objectType;
         this.Name = other.Name;
-        this.baseType = other.baseType;
+        this.typeTags = new HashSet<string>(other.typeTags);
```
relinu: @henninglive the copy constructor is called here,  it copies everything inside other.typeTags into this.typeTags

henninglive: I was wondering about the Furniture copy constructor, but @koosemose explained it to me.

RodRitter:
Nice change. Tested and it's working fine.

PR #343, opened 2016-08-20
A few pathfinding tweaks to increase efficiency a little bit.

tickinginstant:
While looking through the pathfinding code, I discovered that the SimplePriorityQueue being used by the pathfinding system has worst-case O(n) Contains and UpdatePriority, both of which are used frequently by the pathfinding system (resulting in a worst-case O(n^2) algorithm, at least in theory). FastPriorityQueue has O(1) Contains and O(log n) UpdatePriority, but the process of wrapping SimplePriorityQueue around it introduces a fair amount of overhead. So I wrote a variant of the SimplePriorityQueue class that uses Dictionaries to speed things up a bit. Pathfinding is still a bottleneck, but hopefully this can make the neck a little wider?

I also removed the lines initializing all of the f_scores and g_scores, since the process that the algorithm goes through adds f_scores and g_scores for everything in the open set and closed set, and they're not accessed for anything outside of those sets. (And initializing them for all 10000 tiles can be kind of waste of time if you're just trying to move a few tiles over.)

### DIFF
--- a/Assets/Scripts/Pathfinding/PathfindingPriorityQueue.cs
+++ b/Assets/Scripts/Pathfinding/PathfindingPriorityQueue.cs
@@ -22,7 +22,7 @@ public class PathfindingPriorityQueue<T>
     protected Dictionary<T, WrappedNode> _mapDataToWrappedNode;
 
     /// <summary>
-    /// Initializes a new instance of the <see cref="PriorityQueueWrapper`1"/> class.
+    /// Initializes a new instance of the <see cref="PathfindingPriorityQueue`1"/> class.
     /// </summary>
     /// <param name="startingSize">The starting size.</param>
     public PathfindingPriorityQueue(int startingSize = 10)
@@ -42,7 +42,7 @@ public class PathfindingPriorityQueue<T>
         public readonly T data;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="PriorityQueueWrapper`1+WrappedNode"/> class.
+        /// Initializes a new instance of the <see cref="PathfindingPriorityQueue`1+WrappedNode"/> class.
         /// </summary>
         /// <param name="data">The data that this WrappedNode represents in the queue.</param>
         public WrappedNode(T data)
--- a/Assets/Scripts/Pathfinding/PathfindingPriorityQueue.cs
+++ b/Assets/Scripts/Pathfinding/PathfindingPriorityQueue.cs
@@ -1,0 +1,134 @@
+﻿using UnityEngine;
+using System.Collections.Generic;
+using Priority_Queue;
+using System;
+
+/// <summary>
+/// Wraps the FastPriorityQueue class so that it's both easy-to-use,
+/// and faster than SimplePriorityQueue (which sports an O(n) Contains
+/// and an O(n) UpdatePriority -- not exactly ideal).
+/// </summary>
+public class PathfindingPriorityQueue<T>
+{
+    /// <summary>
+    /// The underlying FastPriorityQueue instance.
+    /// </summary>
+    protected FastPriorityQueue<WrappedNode> _underlyingQueue;
+
+    /// <summary>
+    /// The map between data and WrappedNodes.
+    /// Used to make operations like Contains and UpdatePriority more efficient.
+    /// </summary>
+    protected Dictionary<T, WrappedNode> _mapDataToWrappedNode;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="PriorityQueueWrapper`1"/> class.
+    /// </summary>
+    /// <param name="startingSize">The starting size.</param>
+    public PathfindingPriorityQueue(int startingSize = 10)
+    {
+        _underlyingQueue = new FastPriorityQueue<WrappedNode>(startingSize);
+        _mapDataToWrappedNode = new Dictionary<T, WrappedNode>();
+    }
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


WardBenjamin:
Looks good. How much of a performance difference is this?

vogonistic:
@WardBenjamin Maybe we need a larger pathfinder speed test world that we can use to benchmark path finding.

FilippoLeon:
I removed the path tree successfully, now we should test it:
https://github.com/FilippoLeon/ProjectPorcupine/tree/pathfinding

tickinginstant:
@WardBenjamin I don't really know quantitively. My testing pretty much involved laying down a bunch of floor, and then building a bunch of walls, and watching as new characters arrive and everything gets laggier and laggier. As far as I can tell, the improvements that I made delay the onset of lag a little bit, but don't prevent it entirely. (The O(n^2) vs O(n log n) is probably more like an O(n sqrt n) vs O(n log n) difference in practice, for the kind of graphs I was testing it on.)

@vogonistic You're right, some kind of benchmark would be good. Maybe I'll try to cook something up.

@FilippoLeon I was wondering about whether removing the path tree would help -- I guess we can test it now! (Sorry for jumping in with my pull request without saying anything over in the issue thread -- I figured that since the discussion there was mostly focused on the graph generation rather than the pathfinding algorithm itself, my pull request wouldn't really affect whatever you guys ended up doing for graph generation. I forgot that bypassing the path tree would require rewriting most of the algorithm.)

tickinginstant:
I wrote up a simple pathfinding speed test script. I only used two maps (one completely empty, one with most of the map enclosed by a box, with a single switch-back corridor inside it), with four tests per map:

1. Trying to find a path from the center to a specific location 10 tiles away.
2. Trying to find a path from the center down to a specific location ~10 tiles from the corner.
4. Trying to find inventory of a type that's in that same position near the center.
3. Trying to find inventory of a type that's in that same location near the corner.

For each test, I pre-generated the pathGraph (and thus didn't count that in the cost), and then ran A* 10 times, recording the number of seconds it took, then took the average of the trials:

test | master | with patch | % cost reduction
-----|--------|------------|------------------
[Empty] Center to Nearby | 0.01104536 | 0.00175519 | 84.11%
[Empty] Center to Corner | 0.01192036 | 0.001194906 | 89.98%
[Empty] Center to Nearby Inventory | 0.02571073 | 0.01130829 | 56.02%
[Empty] Center to Corner Inventory | 9.373183 | 8.793081 | 6.19%
[Switchback] Center to Nearby | 0.106617 | 0.09355621 | 12.25%
[Switchback] Center to Corner | 2.369411 | 2.304765 | 2.73%
[Switchback] Center to Nearby Inventory | 0.1736702 | 0.1598152 | 7.98%
[Switchback] Center to Corner Inventory | 2.395287 | 2.340334 | 2.29%

Some of those results are probably just the luck of the draw (in particular, the two switchback corner tests). I'm not surprised that the patch didn't really improve performance on the switchback tests, because having a single corridor of width ~4 means that the size of the open set probably stays ~10 most of the time. (Maybe I should have come up with a test that was more taxing for the system.) But IMHO there's enough of an improvement in some of the empty graph tests to make the patch worthwhile.

I put the test code and test scene in a separate branch on my repository, since I'm thinking of putting it in as a separate pull request soon. (Unless it belongs in this pull request? IDK.) It's available here, if you want to run some tests yourself:

https://github.com/tickinginstant/ProjectPorcupine/tree/pathfinding-tests

Mizipzor:
I really appreciate that you took the time to both test and _measure_ your changes. Awesome!

WardBenjamin:
Much appreciated! I'm going to give this one more test and then merge assuming all is well.

vogonistic:
Awesome!

Assets/Scripts/Pathfinding/PathfindingPriorityQueue.cs
```diff
@@ -0,0 +1,134 @@
...
+    /// If the data already exists in the queue, it updates the priority instead.
+    /// Should take O(log n) -- O(1) amortized for the resizing, and O(log n) for the insertion.
+    /// </summary>
+    /// <param name="data">The data to be enqueued.</param>
+    /// <param name="priority">The priority of the data.</param>
+    public void Enqueue(T data, float priority)
+    {
+        if (_mapDataToWrappedNode.ContainsKey(data))
+        {
+            throw new InvalidOperationException("Can't re-enqueue a node that's already enqueued.");
```
OzieGamma: Should we also Log this ?

Mizipzor: No, we obviously want the program to stop immediately so it isn't missed. Don't you agree? ;)

OzieGamma: Yes but also write it to a log file no?

So we know what happened

Mizipzor: Indeed. :)

vogonistic:
@tickinginstant Make a separate request to add the pathfinding tests.

tickinginstant:
@vogonistic Okay, I posted it in PR #459. (And I added a computation for standard deviation, too.) I was originally going to try to figure out some more worlds to test it on, but I guess I can always do that later if the PR is accepted.

vogonistic:
Merged!

PR #348, opened 2016-08-20
Fixed bug in save dialog from the delete button

alexisfcote:
Delete button does not show in the save dialog. By design or by mistake of #194. I added a quick fix to check if GO is found. A proper refactor of the delete button might be in order.


alexisfcote:
It messed with my double click code.

henninglive:
You should maybe run the auto formatter on the whole file. The indentation is messed up.

SquareMan:
Just looks as if your tab settings are set to 3 spaces instead of 4. I'm not sure if this would prevent this being merged or not but you should try to fix that.

henninglive:
no, the whole file is messed up, also lines that have not been touched by his commit.

quill18:
Merging this, but yeah it looks like we need to run a format pass on this file after.

Issue #275, opened 2016-08-19
Room stats don't show oxygen levels (Proposal- Indicate items that don't have power somehow)


kd7uiy:
There's no way to status the health of a room, with things like the proper oxygen levels. This should be addressed.

bjubes:
Mousing over a floor tile in a room won't show the O2 in the bottom right?

BeefEX:
I will try to add it? If it doesnt already say it.

bjubes:
Should this be a beginner bug?

kd7uiy:
@bjubes Yes, this is the idea. I think it used to, not sure why it's not working now.

emilkris33:
@kd7uiy build an O2 gen in a room and a Power Station to generating and then it Will show

TheRiven:
yeah, oxygen is currently the only thing needing power as it does not cause any issue if you don't have it (yet!) and can show that the power system is working when you do.

kd7uiy:
Ahhh, didn't realize it required power... Interesting. Will try.

In that case, there should be a visual indicator on the O2 generator that it doesn't have power. Either it should turn red, or have a lightning bolt symbol on it, or something like that.

ComputerOverlord:
@kd7uiy Haha... But, you probably didn't understand how the generator worked.

Tritaris:
I setup a basic system for displaying a power status icon on furniture that consumes power in PR #350 .

![1a](https://cloud.githubusercontent.com/assets/282954/17829536/8e3f551e-667f-11e6-9e3e-1adbc982cf49.png)
![2a](https://cloud.githubusercontent.com/assets/282954/17829537/8e532f3a-667f-11e6-9b44-e046661f21b0.png)

WardBenjamin:
Fixed! Closing!

PR #350, opened 2016-08-20
Adds a Power Status Icon for Furniture

Tritaris:
Saw in #275 that some people wanted to have power icons for machines that consume power.  So I came up with this:

As long as a Furniture.powerValue is less than 0 it will automatically get the status icon applied to it.  The status icons get updated every time the PowerSystem.PowerLevel is changed. The status icon's tint changes to reflect the PowerLevel. Green > 0 Red <= 0


quill18:
I like the red "out of power" symbol, but since "powered" will be the default status I'm wondering if there shouldn't be any icon at all when it's on (or maybe a smaller one?)

Tritaris:
Yeah I like the idea of there not being an icon when it is "powered". I updated it to disappear and reappear based on the PowerLevel

RodRitter:
Tested and it looks good. Merging.

PR #361, opened 2016-08-20
Null Reference Error Fix

RodRitter:
I check to fix the null reference exception


Mizipzor:
Hi. It seems like an extra bugfix regarding job cancelling did sneak into this pull request. It's usually better to create two pull requests if you want to fix two different things. Also, in that commit you used tabs instead of spaces, you can see in the diff here on Github that alignment is a bit out of whack. With all due respect, maybe you should have let this pull request be reviewed before it was merged, especially since you made it yourself. Not crapping on your stuff, I think the fixes are great. :)

RodRitter:
I can revert the merge, but what do you mean an extra bugfix did sneak into the PR?

Mizipzor:
The bug fix (and tabs thing) would be in 9990f09. Seems like theres also a new feature in 68ae786. And it doesnt appear to have been squashed.

I would only revert the merge if you think this could have been done cleaner, as a learning experience.

RodRitter:
Those commits are already accepted and merged into the project. When we merge, Github does it's own squashing.

Mizipzor:
Really? If those changes have already been merged but show up in your branch as well something might be wrong. Or I'm wrong.

Mizipzor:
Oh now I see whats going on! The Github GUI has changed since I last did this. There is one segment that show the commits from four days ago. And I was talking about those changes. The next segment seems to have commits that replaced the ones in the pull request (this workflow I agree with). And in the new pull request you had seven merges and one change. Sorry to bug you. :)

RodRitter:
All good :)

PR #362, opened 2016-08-20
Missing localization keys are now only logged once.

Mizipzor:
As discussed in #247.

Now in hindsight this turned out rather simple and maybe I should have left it for a beginners task. But it does increase performance by not spamming log with the same message.


Mizipzor:
While I was in the files I took the liberty to do some code cleanup. If you look at the files changed tab you will see that most if the file did change. If you want to see what was _actually_ done look at commit 4a6354a.

brlodi:
Exactly the solution I had in mind. Thanks @Mizipzor!

Mizipzor:
@brlodi Thank you. :)

PR #363, opened 2016-08-20
Health system

MANDAL0R3:
I use this script whenever a project i am working on requires some sort of health system. I hope this can be of great use for the project, especially for https://github.com/TeamPorcupine/ProjectPorcupine/pull/135.


satch3l:
Is it implemented into the game or just left as a resource for future use?

Mizipzor:
@satch3l The file is added but not used anywhere.

@MANDAL0R3 Is this file used identically somewhere else? If so, that might be a licensing concern. Though I dont know if we care about that. Im not a lawyer.

MANDAL0R3:
@Mizipzor Should be no licensing concern since all the projects that i have included it with have just been small private ones they just test out new stuff. The main point of why im trying to add this is have some sort of health system that can be easily used by anyone in anyway.

Assets/Scripts/Models/HealthSystem.cs
```diff
@@ -0,0 +1,117 @@
...
+        get
+        {
+            return _currentHealth;
+        }
+
+        private set
+        {
+            if (Invincible == true) return;
+            //We have been Revived!
+            if (_currentHealth <= 0 && value > _currentHealth)
```
OzieGamma: If value is 0, does that revive us ?

Assets/Scripts/Models/HealthSystem.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+    }
+    public float MaxHealth { get; set; }
+    public bool Invincible { get; set; }
+    public bool Healable { get; set; }
+    public bool IsAlive { get { return Value > 0; } }
+
+    private float _currentHealth = 100.0f;
+    private Action _OnHit;
```
OzieGamma: Make them events ? So we can attach several handlers?
https://msdn.microsoft.com/en-us/library/aa645739(v=vs.71).aspx

Assets/Scripts/Models/HealthSystem.cs
```diff
@@ -0,0 +1,117 @@
...
+
+    public HealthSystem(float currentHealth, float maxHealth, bool invincible, bool healAble)
+    {
+        _currentHealth = currentHealth;
+        MaxHealth = maxHealth;
+        Invincible = invincible;
+        Healable = healAble;
+    }
+
+    public void RegisterOnHit(Action act)
```
OzieGamma: If we use events, this is not need I think

Assets/Scripts/Models/HealthSystem.cs
```diff
@@ -0,0 +1,117 @@
...
+    public void UnregisterOnRevive(Action act)
+    {
+        _OnRevive -= act;
+    }
+    public void UnregisterOnDestruction(Action act)
+    {
+        _OnDestruction -= act;
+    }
+
+    public static HealthSystem operator+(HealthSystem h1, float num)
```
OzieGamma: Feels weird, operators should be pure. (Not change the objects they operate on).

Imagine the following code:
```csharp
var health = new HealthSystem(...)
var newHealth = health - 5.0f;

// health == newHealth
// and object.ReferenceEquals(health, newHealth) == true
// which is unexpected
```
@Mizipzor What do you think? 

Sorry if my explanation is not good.

Assets/Scripts/Models/HealthSystem.cs
MANDAL0R3: Yes, for reviving, if the characters health is at  0 or less then he is dead. But if the current value were adding to his health is greater then 0 then he is being revived.

Assets/Scripts/Models/HealthSystem.cs
MANDAL0R3: No your explanation is good. Honestly i added the operator's last night before pulling and this was the first time i ever added them. At the time i did not see the problem but that's what happens when you get something you think you kinda know to work at 4 in the morning haha.

Assets/Scripts/Models/HealthSystem.cs
OzieGamma: Imagine 
```csharp
var health = new HealthSystem
health = -5.0f;
// We are now dead
health = -3.0f; // (_currentHealth <= 0 && value > _currentHealth) will be true, so we revive even though our health is still < 0.
```

I think the condition should be `(_currentHealth <= 0 && value > 0)`

Assets/Scripts/Models/HealthSystem.cs
MANDAL0R3: Correct me if im wrong but doesn't delegates allow for multiple subscribers. So wouldn't that be that same use?

OzieGamma: I might have just learned something ^^

Anyways if you declare them as events:
```csharp
public event Action OnHit;
```

You don't need to manually protect them:
http://stackoverflow.com/questions/29155/what-are-the-differences-between-delegates-and-events

Events are a compiler feature to simplify exactly this use case.

Assets/Scripts/Models/HealthSystem.cs
MANDAL0R3: You make a great point. I never thought of the revive that way. This is my first time doing a group project on github and i have to say this is a great learning experience.

Assets/Scripts/Models/HealthSystem.cs
MANDAL0R3: I see your point. At first they were just a quick addition so i can keep moving through the code but it would definitely shorten the code and make it simpler. Thank you for all of the feedback.

Assets/Scripts/Models/HealthSystem.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (_currentHealth == 0 && value > _currentHealth)
+            {
+                _currentHealth = value;
+
+                if (OnRevive != null)
+                    OnRevive();
+                return;
+            }
+            //we have been hit but we are not dead!
+            if (value < _currentHealth && _currentHealth > 0)
```
OzieGamma: 0 < value && value < _currentHealth

Assets/Scripts/Models/HealthSystem.cs
Mizipzor: @OzieGamma your specific example is absolutely correct, But some operators do change the object they operate on. Like `i++` or `o[u] = 0`, I realize now that Ive forgot the names for these. :S

Assets/Scripts/Models/HealthSystem.cs
MANDAL0R3: To be honest, after testing both mine and your condition for being revived, they both seem to work the same but i think that yours is a bit more clear then mine so if you do not mind i would like to use your change to the condition.

Mizipzor: It's also way more readable, imo. :)

Assets/Scripts/Models/HealthSystem.cs
OzieGamma: `O[u]` is called an indexer. It has a get and set. It’s exactly the same as a property, so it can indeed change the object.

 

`i++` is shorthand for `i += 1`, which in turn is shorthand for `i = i + 1;` So *strictly speaking* it does not change the object it creates a new object and assigns it to the variable name from before. (I am speaking on a language level. In practice it often changes the object for performance reasons)

Assets/Scripts/Models/HealthSystem.cs
MANDAL0R3: I agree. At least this is a good learning experience for me :)

MANDAL0R3: @Mizipzor im still new to using github with other people. Is it ok if i use his change or should i wait for permission?

Assets/Scripts/Models/HealthSystem.cs
Mizipzor: I stand corrected. :) 

Also, you might look into how your email are formatted, seems like Github is expecting you to clear the email before sending?

Assets/Scripts/Models/HealthSystem.cs
OzieGamma: @MANDAL0R3 You are 100% free to use my change. All my contributions are MIT/Apache 2.0 multi-licensed if that's your question.

And copyright doesn't really cover 1 line ;) At least in Switzerland where I am.

Mizipzor: What do you mean by using his change?

OzieGamma: I think that's what he means.

Mizipzor: Oh, I took that as a given. :)

MANDAL0R3: Thanks to both of you :)

OzieGamma:
Otherwise I did a PR on your project:
https://github.com/MANDAL0R3/ProjectPorcupine/pull/1

MANDAL0R3:
The main reason for having the operators was for quicker changing of the health value and also with your PR, i don't see a way to change the health value.

Assets/Scripts/Models/HealthSystem.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+    //all of this data should probably be generated at runtime or loaded from a file
+    public float Value
+    {
+        get
+        {
+            return _currentHealth;
+        }
+
+        private set
```
MANDAL0R3: Since there is no way to change the health outside of this class as is, you should remove the private here

Mizipzor:
I would still like to see this used for something, anything. Nothing wrong with your code, but I think it's unreasonable to fill the code base with bits and bobs that might maybe be useful for someone, some day. Anyone that gets into the project would have to understand the code, as much of it as possible, preferably. We can aid them by removing as much as possible. No offence. :)

OzieGamma:
I agree with @Mizipzor. I hope it helped you learn about GitHub and PRs and giving credit.

My commits from my PR on your fork are here ;)

MANDAL0R3:
@Mizipzor No offense taken. I believe i have mentioned this in one of my past comments but originally i just wanted to give this file to the person making https://github.com/TeamPorcupine/ProjectPorcupine/pull/135 but i don't exactly know a good way to do that so i decided that since someone and im sure many others are working on making combat/damage a thing, i thought just putting this script into the project could help anyone out and make it all alot easier.

OzieGamma:
You can always make a PR on their PR like I did.

MANDAL0R3:
@OzieGamma, you are so right. I can't believe that i did not think of that. Again thank you for everything, this community is amazing :)

BeefEX:
Thanks @MANDAL0R3 i have to rewrite my system #135 so i will inspire from this :smile:

MANDAL0R3:
Awesome im glad that i can help :)

PR #367, opened 2016-08-20
Italian l10n

FilippoLeon:
Fixed some autotranslation and added extra strings.


WardBenjamin:
Thanks! Our auto-translations really aren't that great, so it's nice to have a native speaker do it.

PR #368, opened 2016-08-20
Additional sounds

satch3l:
Additional sounds that can be used.


RodRitter:
Did you create the sounds yourself? If not you need to include licensing info.

satch3l:
Yes

dustinwloring1988:
Nice cant wait to have more sounds, Hopefully it gets commit today. :)

RodRitter:
I'll merge this in, but I'm not sure what our standards are for the file format, so it might need to be converted later depending on what's decided.

Mizipzor:
Should also be noted that the sounds are just added, they are not used anywhere (yet). Is there an issue up to implement that? If not, consider creating one.

PR #374, opened 2016-08-20
More GameEvent fixes

relinu:
So this make Repeat an optional parameter, if maxRepeats is set to 0 or less it repeats unlimited. Also fixes the character moonsharpuserdata issue


WardBenjamin:
Looks great!

WardBenjamin:
Thanks!

PR #380, opened 2016-08-20
Lock inventory

### LINKED ISSUE #339: Metal Smelter- Don't steal from one to give to another one
gif

UnknownSoldier1:
I added the ability for inventories to be "locked". This basically means that inventories that are locked can't be picked up for completing jobs. This is a fix for #339.


RodRitter:
Would this be desirable if we have a huge stack and 2 people need to access half of the stack each? or is the idea to wait for one character to finish using the stack?

UnknownSoldier1:
I don't fully understand what you are asking. It is meant that characters can't pick up for example one Metal Smelter's raw iron for putting it for example to another Metal Smelter.

RodRitter:
Oh right, I apologize. I see that now. Good job.

PR #388, opened 2016-08-20
Added missing meta files.

Mizipzor:
These should probably have been part of #368.


RodRitter:
Nice

WardBenjamin:
Thanks! This was annoying when I submitted my last PR. Good fix.

PR #390, opened 2016-08-20
Moved StyleCop settings to the project root

Mizipzor:
Hey, @quill18 I did it! :)

One drawback remains that users who runs this within MonoDevelop will have a cache file in the Asset folder. Visual Studio places the cache at root. All cache files are gitignored. 

Also took the liberty of disabling one rule that we clearly dont agree with. We place usings outside the namespace, not inside.


Mizipzor:
Oh I cant delete the branch before its merged? Oh well, lets try again.

quill18:
I did that once too. :)

quill18:
As for the `using` statements, I don't have a strong preference either way except that the templates place them outside by default, so we may as well go with it just to use the path of least resistance.

Mizipzor:
The branch is back and those are the same commits, I've learned a new trick in Git. It should be possible to merge now. :)

quill18:
I would probably also do a `git rm /Assets/StyleCop.Cache.meta` as part of this.

Mizipzor:
Hm, I dont think they are under source control actually.

```
λ git ls-files --others -i --exclude-standard Assets
Assets/StyleCop.Cache
Assets/StyleCop.Cache.meta
```

Try `git clean -df` and they should go away for you.

WardBenjamin:
Looks great! Merging!

Mizipzor:
Thank you. :)

PR #396, opened 2016-08-20
First Pass basic dev mode

koosemose:
This is a very simple dev mode to aid in development. It currently has no UI connection nor any hotkeys to activate. At present the only way to turn it on is by setting devMode in the WorldController. However, as it's public, it can be toggled on and off from the WorldController GameObject.

### DIFF
--- a/Assets/Scripts/Controllers/BuildModeController.cs
+++ b/Assets/Scripts/Controllers/BuildModeController.cs
@@ -132,7 +132,14 @@ public class BuildModeController : MonoBehaviour
                 }
 
                 // Add the job to the queue
-                WorldController.Instance.world.jobQueue.Enqueue(j);
+                if (WorldController.Instance.devMode)
+                {
+                    WorldController.Instance.world.PlaceFurniture(j.jobObjectType, j.tile);
+                }
+                else
+                {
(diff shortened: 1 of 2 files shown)


RodRitter:
Perhaps you could add a way for it to be accessed now (settings menu or hotkey) - otherwise we'll end up with people leaving dev mode on/off when they commit. Also people might not know it's available.

Mizipzor:
How about a new menu besides the "build" menu called "spawn"? As in, dont create a build job, create instantly. Only visible if "dev mode" is active. This cuts down on complexity as it doesnt affect the current build menu. Great initiative though. :)

koosemose:
Unfortunately I don't have the time for that part right now. This was kind of a last moment thing as I worked out how to do it. If you would prefer it to not be merged until it has some form of UI support I'll mark it as WIP until I'm able to do so.

RodRitter:
I'll merge it in but please don't let it sit there unused. People must know about it. So add it to a menu or option when you get a chance.

koosemose:
Don't worry, I am a very lazy programmer, there's no way I'll want to constantly have to go in and edit the code or open the WorldController to toggle a checkbox when I could just click a menu or something.

Issue #393, opened 2016-08-20
[Bug] Scrolling in Language Selector also zooms game.


bjubes:
when moused over the language selector drop down, scrolling with both scroll the dialog (very slowly, which is a seperate issue) and zoom the game in and out.

PR #397, opened 2016-08-20
Disable scrolling over ui-elements

ttyz:
A fix for #393. I always thought that it was a good idea to disable scrolling while over ui.


RodRitter:
Nice change

PR #398, opened 2016-08-20
Dutch localization

mrDLSable:
Human made Dutch localization instead of the default not-quite-right robotic one. Also added translations for the new strings.


WardBenjamin:
Looks good, and thanks! Merging!

PR #399, opened 2016-08-20
Language Dropdown Localization + English (UK)

nsfranklin:
language_name = "names you want displaced added"

English (UK) added

English (UK) made default.


Mizipzor:
Doesnt US English make more sense as a default? Though, I must admit I enjoy annoying coworkers by insisting on UK spelling. ^^

nsfranklin:
True. Also they're the same with no translation differences yet.

nsfranklin:
It was mostly as a test to see if the change functioned correctly. As we made it less hard coded so changing the default in the code doesn't break all of the localisation.

WardBenjamin:
@nsfranklin Thanks for the UK localization!

I do think the en_US one should be default since that's what we're using for our codebase, but I'll gladly take the "translation" as well :P

This PR also currently conflicts with something. Make those two changes and it should be good to go.

Assets/Scripts/Localization/LocalizationTable.cs
```diff
@@ -59,7 +61,7 @@ public static void LoadLocalizationFile(string path)
          * </para>
          * </summary>
          */
-        public static void LoadLocalizationFile(string path, string localizationCode)
+        public static void LoadLocalizationFile(string path, string localizationName)
```
Grenkin1988: I do not think that it is goot idea to change "key" of the dictionary to be name. There should be another way to transalate language code to readable format without this.
```currentLanguage``` is changed from outside only when selection of language changes. At this point you should map Name of language to it's key/code.

Grenkin1988:
#425 here I am changing some logic of storing data in library.

Assets/Scripts/Localization/LocalizationTable.cs
Mizipzor: I looked up the [table at msdn](https://msdn.microsoft.com/en-us/library/ee825488(v=cs.20).aspx) and it seems to use "name" for what we refer to as code. While code being just a number. I personally think we should stick to code. Are there any other documentation that names these values we can look at?

Grenkin1988: @Mizipzor I am not arguing on ```localizationName```. I am agains of changing `Key` in dictionary to be "Display Name" of culture.

vogonistic:
I got lost in my review order and merged #425 first, so you need to do some changes before I can merge this. Sorry about that.

I also want you to fix 2 things:
  1. `en_US` should remain the default for now 
  1. I also want you to not use the translated name as the key, like @Grenkin1988 said. If we ever change the name of a language, like `English (UK)` to `British English` (bad example), it would fail to load.

nsfranklin:
I'm going to close it now while I work on it.

PR #400, opened 2016-08-20
made construction buttons toggles

bjubes:
Issue:  the construction menu has buttons for furniture and floor which yield more buttons when clicked. However, when you click furniture again, the furniture buttons don't go away.

Fix: turned furniture and floor buttons into toggles so their menus can be hidden by pressing them again

note: this required scene modification to change the buttons OnClick targets

toggle

### DIFF
--- a/Assets/Scripts/UI/MenuController.cs
+++ b/Assets/Scripts/UI/MenuController.cs
@@ -38,6 +38,11 @@ public class MenuController : MonoBehaviour {
 		floorMenu.SetActive (false);
 	}
 
+    //toggles whether menu is active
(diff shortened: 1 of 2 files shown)


RodRitter:
Nice change

RodRitter:
@bjubes I found an issue where if you leave the Furniture menu open and then try to open the Floor menu, the Furniture menu stays on top. You might want to toggle all other menus off when opening a different menu.

bjubes:
Will work on that. Should I put those changes on this branch or make a new PR now that it's merged?

dustinwloring1988:
nice

PR #401, opened 2016-08-20
new branch for xml settings

zwrawr:
Old branch was a mess and had unneed file modifications in it , Created
a clean branch and updated localisation code to use Settings class

### DIFF
--- a/Assets/Scripts/Controllers/WorldController.cs
+++ b/Assets/Scripts/Controllers/WorldController.cs
@@ -178,8 +178,12 @@ public class WorldController : MonoBehaviour
 
     void CreateEmptyWorld()
     {
+        // get world size from settings
+        int width = int.Parse(Settings.getSetting("worldWidth", "100"));
+        int height = int.Parse(Settings.getSetting("worldHeight", "100"));
+
         // Create a world with Empty tiles
-        world = new World(100, 100);
+        world = new World(width, height);
 
         // Center the Camera
         Camera.main.transform.position = new Vector3(world.Width / 2, world.Height / 2, Camera.main.transform.position.z);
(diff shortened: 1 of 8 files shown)


WardBenjamin:
Looks good, with one small caveat: "localisation" should be "localization" to be consistant with the rest of the code base. We've settled on using American (en_US) spellings and such for code.

zwrawr:
Changed it to the US spelling

benji7425:
Would it not be better to use ORM to generate a settings file from a bunch of defaults set in the class? This would have the advantage of not requiring anyone to modify the settings file directly, they can just add a field to the DataContract class. The DataContract class can then be used throughout development, and serialized out to XML closer to release

zwrawr:
If you specify a default value in `Settings.getSetting(string key , string defaultValue)` then it'll save that into Settings.xml . So you don't need to edit the xml directly. 

ORM and DataContract class are something i've never heard of before so it may well be a better solution in the long run.

benji7425:
ORM = Object Relational Mapping, it just means the mapping of data to and from a runtime object. DataContract is an attribute you can specify on the class which means a serializer can automatically serialize it to a flatfile (ie json or xml realistically). The disadvantage with using a method and passing in a default value is that you need to actually run the method once to generate the entry, to use it in development. Using a DataContract class just means you simply add a field ie

[DataMember]
public string mySetting = "Value here";

Then you can just refer directly to Settings.mySettings. When development is finished, you can serialize this class to XML. Then in the release, you can add a method to deserialize from XML and overwrite the hardcoded fields.

vogonistic:
@benji7425 If you want to, you can supply a PR that uses ORM or start a discussion. This PR fits in well with how the project has been done in general.

Merged!

PR #402, opened 2016-08-20
Slightly increased the scroll-sensitivity for the language dropdown menu

Randshot:
I noticed that the scrolling for the language menu was too slow, so I slightly increased it.
I am also thinking of setting the scroll sensitivity based on the amount of options in the drop-down menu.
Haven't looked into that yet.


bjubes:
maybe in the future scroll sensitivity can be an option that will affect both that drop down and the saveload dialog

RodRitter:
Yep, it's something to consider when the settings.xml is implemented. For now though it's a good fix.

PR #405, opened 2016-08-20
Fix no name being shown

Ceddini:
Okay, finally. Here is the "quick" fix.

Oh, btw... Stupid regex. I don't like it!


RodRitter:
I see you changed `lines.Length-1` to `lines.Length` - Are you sure thats not going to throw an out of bounds exception if it tries to get the last name?

Ceddini:
@RodRitter The tooltip said the max is excluded, so I thought this would be correct

RodRitter:
Nope, I just tested it - You might want to change it back :)

`c.name = lines[lines.Length];`

image

Ceddini:
Fixed it, but this tooltip says differently :(

random_range

RodRitter:
I think that's a bit misleading. I've tested it and it's definitely not exclusive. Good job nonetheless :)

Ceddini:
Thank you very much. From now on, I will just go with what I know from java because there I have to use -1!

PR #406, opened 2016-08-20
Dynamic scroll sensitivity for the language dropdown menu

Randshot:
Like I mentioned in the pull request #402.
The scroll sensitivity of the language drop down menu will be set based on the options-count.
Was easier than I initially thought :)


RodRitter:
Looks good :)

PR #410, opened 2016-08-20
Updated danish localization

racerand:
Updated danish localization with all new words/sentences, and better translation of old ones. Have not translated pathfinding test, as I don't think it needs localization as it will be removed at some time.


WardBenjamin:
Looks great, thanks!

PR #415, opened 2016-08-20
Updated danish localization

racerand:
Sorry, I missed a few words, thought they were not implemented in localizations.


RodRitter:
Thanks

PR #416, opened 2016-08-20
Added Canceling to Jobs when in select mode.

DevOrc:
If in the select mode, and you right click on a job, it will cancel that
job.


RodRitter:
Nice

PR #418, opened 2016-08-20
Missing texture sprite

NogginBops:
Also fixed a bug where objects that link to neightbours wouldn't get any
preview sprite.

Old:
image
New:
capture1


vogonistic:
Merged!

PR #375, opened 2016-08-20
Cosmetic changes to LocalizationTable class

Grenkin1988:
What do you think about this changes?
```line.Split('=');``` i think .NET know how to work with strings efficiently?

PR #425, opened 2016-08-20
Cosmetic changes to LocalizationTable class

Grenkin1988:
#375 Second try after all changes merged


Assets/Scripts/Localization/LocalizationTable.cs
```diff
@@ -6,193 +6,118 @@
...
-                        {
-                            // The char is an '=', set searchingValue to true and ignore the current char.
-                            searchingValue = true;
-                        }
-                    }
-                    else
-                    {
-                        // The loop is searching for a value. Add the current char, regardless of what it is.
-                        currentValue += c;
+                        throw new InvalidOperationException(string.Format("Invalid format of localization string. Actual {0}", line));
```
Mizipzor: Are you sure about throwing exceptions here? For the record, I prefer crashing rather than having bad data. But in the rest of the code is seems more common to log error and then put an empty string there or something. You should maybe make sure that whoever calls initialize is prepared to catch the exception.

Assets/Scripts/Localization/LocalizationTable.cs
```diff
@@ -6,193 +6,118 @@
...
         }
 
         public static string[] GetLanguages()
         {
-            return registeredLanguages.ToArray();
+            return localizationTable.Keys.ToArray();
         }
     }
-}
+}
```
Mizipzor: It is the norm to end the file with an end-of-line character. I usually dont bother with these, but in this you actually removed it.

Assets/Scripts/Localization/LocalizationTable.cs
```diff
@@ -6,193 +6,118 @@
...
-        public static string GetLocalization(string key, FallbackMode fallbackMode, string language, params string[] additionalValues)
+        /// <summary>
+        /// Returns the localization for the given key, or the key itself, if no translation exists.
+        /// </summary>
+        private static string GetLocalization(string key, FallbackMode fallbackMode, string language, params string[] additionalValues)
         {
-            key = language + "_" + key;
             string value;
-            if (localizationTable.TryGetValue(key, out value))
+            if (localizationTable.ContainsKey(language) && localizationTable[language].TryGetValue(key, out value))
```
Mizipzor: This could be slightly misleading. If you select a language that doesnt exist, it will log the _key_ as missing. I think this was discussed in another issue. I'm willing to let this one slip for this pull request, it can be fixed later. :)

Assets/Scripts/Localization/LocalizationTable.cs
OzieGamma: At the very least make sure to log it

Grenkin1988: At work I have practice - crash early. This could happen if someone provided not good localization file.

Assets/Scripts/Localization/LocalizationTable.cs
Grenkin1988: 1. You cannot select language that is not in library. As I anderstand list of languages is taken from ```GetLanguages()```
1. This woulg as ```"Translation for {0} in {1} failed: Key not in dictionary.", key, language```

Assets/Scripts/Localization/LocalizationTable.cs
Grenkin1988: @Mizipzor That's just my habit. Remove all unnecessary empty lines. We are not paid for number of lines :-)

Assets/Scripts/Localization/LocalizationTable.cs
Mizipzor: Yes, yes, yes I couldnt agree more. Crash early, crash often, I say. Maybe this should be discussed and settle on a standard. I have a feeling we are the minority. :P

OzieGamma: Crash early, crash often. Only if needed though. Doesn't make sense to crash on the user machine. You want to crash on dev machines

Assets/Scripts/Localization/LocalizationTable.cs
Mizipzor: Ah, but thats where youre wrong, a line [is defined](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) as:

> 3.206 Line
> A sequence of zero or more non- _newline_ characters plus a terminating _newline_ character.

From [Stackoverflow](http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline). In practice, this means that what you think is the last line isnt. Because it doesnt end with an end-of-line char. This obviously doesnt matter in most cases, or you would know about it. If you look at the Github diff of the file theres a small icon there highlighting the missing eol, and its not just for show.

Assets/Scripts/Localization/LocalizationTable.cs
Grenkin1988: @OzieGamma This is release or development still? How do you prefer to catch that someone screwed up with localization files?

Assets/Scripts/Localization/LocalizationTable.cs
Mizipzor: I see, I might be wrong then. Could you check the last few messages in #153? Where I misunderstand what is supposed to be logged, and it seems to revolve around a missing language, rather than a missing key in that language.

Assets/Scripts/Localization/LocalizationTable.cs
OzieGamma: In a testing environment. Not sure about this particular case, my comment was general

Grenkin1988:
@Mizipzor @OzieGamma Changed exception trowing with logging. Added line in the end

Mizipzor:
Awesome.

I would like to take this opportunity to highlight another cool Github feature. Do you see the discussions we had before, now that the lines have changes, that they are collapsed, since they are out of date? :)

Grenkin1988:
@Mizipzor Yes, interesting feature.

Grenkin1988:
@Mizipzor If I installed StyleCop to my Visual Studio where should I get configurations for in from? And how to apply this configurations?

Mizipzor:
Theres the [StyleCop](https://github.com/TeamPorcupine/ProjectPorcupine/wiki/StyleCop) wiki, but in short:

1. Install StyleCop if it isnt installed (it should be).
2. Get latest master.
3. Run StyleCop under Tools.
4. Stuff it finds is where compile warnings are.
5, And we have a lot of them, you probably want sort on file in that table and only look at relevant files.

Grenkin1988:
@Mizipzor So I do not need to configure it? It is already done on project level and my instance know what to do?

Mizipzor:
You dont need to configure it, thats the point. :) There should be a `Settings.StyleCop` right next to the solution file.

Mizipzor:
Isn't it nice that you don't have to discuss this? You get a list of stuff to do, without room for interpretation, and you do it. And no one needs to be nagging. :)

vogonistic:
Merged!

PR #432, opened 2016-08-20
Test while watching the tutorial video - cleanup up some whitespace and brackets

DeShrike:


Mizipzor:
Good work on the braces, but you're still using tabs instead of space. Try again. :)

PR #434, opened 2016-08-20
Fixed O2Gen producing O2 after powerloss

Tritaris:
Bug Fix  - O2Gen would produce O2 when the PowerLevel dropped to 0 or below

### DIFF
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -384,7 +384,10 @@ public class Furniture : IXmlSerializable, ISelectable
         {
             if (World.current.powerSystem.RequestPower(this) == true)
             {
-                return true;
+                if (World.current.powerSystem.PowerLevel > 0)
+                    return true;
+                else
+                    return false;                
             }
             else
             {
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -384,10 +384,7 @@ public class Furniture : IXmlSerializable, ISelectable
         {
             if (World.current.powerSystem.RequestPower(this) == true)
             {
-                if (World.current.powerSystem.PowerLevel > 0)
-                    return true;
-                else
-                    return false;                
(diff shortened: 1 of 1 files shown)


TheRiven:
What are you doing to create that situation, removing the power gen?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -384,7 +384,10 @@ public bool HasPower()
         {
             if (World.current.powerSystem.RequestPower(this) == true)
             {
-                return true;
+                if (World.current.powerSystem.PowerLevel > 0)
```
OzieGamma: `return World.current.powerSystem.Powerlevel > 0;`

No need for the `if`-`else`

Tritaris:
> What are you doing to create that situation, removing the power gen?

Yeah i was deconstructing the power gen. @OzieGamma thanks for the tip.

TheRiven:
I thought so, I have been thinking about that one myself and I feel the change should actually be in the `PowerSystem` not in the `Furniture`. The main problem is that changing the power available does not rebuild the list of allowed consumers which is why the 02 gen continues to draw power, because it is technically still "powered" if it's in that list.

I don't think changing the `HasPower` function helps as that does not take into account how the power system is currently supposed to work. 

It should be the Power system telling the furniture that it can't run because its no longer powered because of a supply change, not the furniture checking if there is any power in the system.

If that makes sense... :confused:

Tritaris:
The way I saw the consumer list was as a way to know if the furniture was connected to the powergrid. Is that what you mean by its "powered" if in that list?

TheRiven:
Something like that, the system checks if the furniture is in that list and if it is then there is enough power to support it and it can run, if it's not in the list then there was not enough power when it tried to register last.

The problem is that currently it does not recheck if the furniture should be registered when you reduce the power supply, I am currently looking at a change that will fix this and another issue.

Tritaris:
closing #460 is a better fix

Issue #142, opened 2016-08-18
Animated character - is it to early?


Tranberry:
I want to try my hand at making a a place-holder character that is animated. But I would like to know what the community have to say on that topic.

Is it just a waste of time or would it be a good thing to chisel out now?

---

If I was to start I would make something similar to these #102 exploded and animated with the animator feature of unity, but I could also try some hand-drawn static animation (sprite-sheet).

Raketfart:
I think it's fine to work on it now. 
I would prefer sprite-sheet animation, as it's more in line with the rest of the project, using png and xml files. It would not depend on Unity very much, and allow for easy modding.

Mizipzor:
If you want to do it then do it. Animations should be implemented completely separate from any game logic (at least in this game) so it's never too early (nor too late):

Tranberry:
@Raketfart That was my guess (and fear). That will take days to make* compared to the unity animator way.

`*` looking ok'ish.

@Mizipzor I think Raketfart has a valid point in not using prefabs(animator), and frankly I don't think I will spend time on that then.

Raketfart:
If you want, I'll do some spritesheets for you.

Tranberry:
@Raketfart  I would like that

Raketfart:
Here is a modified version of KenneyNL's sprites. It's still very much placeholder art, but should be enough to get us started.

Here is a preview (with background)
![preview](https://cloud.githubusercontent.com/assets/7031323/17773714/faef85fc-654f-11e6-997a-abc35ca5d0fe.png)


This is an xml file, containing coordinates (I couldn't attach .xml directly):
[p1_front_animated.txt](https://github.com/TeamPorcupine/ProjectPorcupine/files/424842/p1_front_animated.txt)

Here is the png file. I hope you can flip the east facing parts for west facing parts. Otherwise I'll add those.
![p1_animated](https://cloud.githubusercontent.com/assets/7031323/17773513/23369100-654f-11e6-9229-1e7b5850ec92.png)

Not sure how you want the xml file to describe the animation, but I'll leave that up to you.

Tranberry:
I'd say that is exectly what I wanted to do, now it's simply adding stuff over this to change the sprite. I wonder if we should dive into clothes and 'wearables' and how that will work.

RodRitter:
Those are pretty cool. I don't see why you can't add it as a Unity animated sprite.

Tranberry:
[...]

PR #266, opened 2016-08-19
[Cosmetic] Character Improvements

SteelFall:
This "feature" is purely cosmetic.
This will give the characters in the world a bit more personality by assigning colors to them at random.

-------

What this "feature" does:

- New World will assign a random color to the starting character.

- Saving the world will also save the characters colors and loading these new saves will load the colors.

- Loading a save made prior to the addition of this feature will assign a random color to the character(s) present

-------

Screenshots:

charactercolor4_1

charactercolor4_2

charactercolor5_1

PR #444, opened 2016-08-20
Added 3 spritesheets for character animation

Raketfart:
As mentioned in issue #142 these spritesheets are for character animation


Raketfart:
These files are for first stab at character animation. 
The xml file has tile information. Theres an idle south east and north facing character, as well as a 2-step walking animation for south, east and north. West facing could hopefully be flipped in the animation script.

Here is a preview.
preview

Also added is the character without helmet, and a spritsheet with only the helmet. Tile sizes are all the same.

Raketfart:
Oh, and these are of course just modified versions of kenney.nl's spaceman, same as the original

dustinwloring1988:
This should be a nice touch.

Tranberry:
test1

For us who can't visualize as good. (wrong fps).

bjubes:
Make sure to open in unity and add the generated meta files.

Raketfart:
Done, thanks

Tranberry:
@vogonistic @WardBenjamin @RodRitter - any input?

WardBenjamin:
Looks good to me! I think the changes in #266 would be reasonably easy to modify so that they work with this animation system, and this looks great. Thanks!

SteelFall:
It should be pretty easy to combine. The only changes I have really made were adding the helmet to a separate sprite and changing the character colour to white.

PR #445, opened 2016-08-20
Inventory desc to lokalization

racerand:
Changes the way description is made in inventories getDescription() method to objectType+_desc to make it easy to implement localization forinventory descriptions. Also added en_US and da_DK implementations for these descriptions.


Grenkin1988:
Localizing description without localizing name is a bit odd as I think

racerand:
I have not added the name yet, as a thread is going on where the naming standard of inventory is going on. I have implemented the Steel Plate and Raw Iron, but not sure which implementations we are going with yet. And it's descriptions for Steel Plate and Raw Iron, which are the only working inventory in game as far as I know.

PR #447, opened 2016-08-20
[Bugfix] Prevent both floor and furn menu being open at the same time

bjubes:
My last PR #400 introduced a bug where both the "build furniture" and "build floor" menus could be open at the same time. This PR fixes that issue.


vogonistic:
I merged some whitespace changes to MenuController.cs, so I had to do some manual merging to get this in. The bug should be resolved now.

bjubes:
so @vogonistic  the bug is fixed in master then?

bjubes:
@vogonistic Someone Reverted both this bugfix AND the original toggle functionality :/

vogonistic:
@bjubes That whole menu system has changed a lot the last 24h. Can you look into it?

Issue #433, opened 2016-08-20
[Bug] Crash when using stockpiles


petroVARS:
I was just trying out the game and when i place a metal miner and smelter and place a stockpile afterwards, Unity crashes.  Before the crash, the fps drops as low as 10. The performance issue also occurs if there is no steel when building a wall.

WardBenjamin:
See #422

Issue #422, opened 2016-08-20
Slowdown: No tile contains objects of type 'Steel Plate' to satisfy job requirements.


sboigelot:
When this condition is met the game almost freeze.

Probably the job is trying to search for the inventory evey tick?

satch3l:
I also reach a point where everything freezes an the fps is plummeting. Nothing I can do but restart. Don't know what causes it though. Maybe this is the answer?

WardBenjamin:
This is probably a pathfinding issue, see #345

Mizipzor:
@WardBenjamin Hold on I believe this one is valid. :) If you place a furniture requiring steel and there is no steel in the world, the log is spammed until there is.

WardBenjamin:
Huh, okay. This is probably still relating to pathfinding in some way though, so the solution might be related/the same. Reopening.

Mizipzor:
Didnt this get rather redundant? 

I thought we common people tagged with [Bug] until you guys could come along and add the label properly and remove the text tag. If not, thats really how it should work in my humble opinion. No need for both. :)

WardBenjamin:
Fair enough!

tomik148:
i have profiler pic here 
![2016-08-20 8](https://cloud.githubusercontent.com/assets/11463094/17833688/0594a57e-6725-11e6-8719-d08d0faeec15.png)

bitope:
I dont think it's a pathfinding issue. It's because jobs are getting aborted and then grabbed again immediately... The character then tries to pathfind to find resources. Just adding a small cooldown to a character when his job gets cancelled fixes the issue.

kd7uiy:
I think this has been resolved.

vogonistic:
Plenty of work that needs to be done for this to be perfect, but the worst offenders are being worked on.

PR #450, opened 2016-08-20
Attempt at pseudo-fixing the performance problems

Randshot:
#422 #433 
This is my attempt at fixing some of the performance issues when the desired object type isn't available.
For more detailed information, look into the changed files.

Please tell me your opinions about this. ( I know that it isn't very elegant ;) )


Assets/Scripts/Models/Character.cs
```diff
@@ -322,7 +336,7 @@ bool CheckForJobMaterials()
                     {
                         //Logger.Log("pathAStar is null and we have no path to object of type: " + desired.objectType);
                         // Cancel the job, since we have no way to get any raw materials!
-                        Logger.Log("No tile contains objects of type '" + desired.objectType + "' to satisfy job requirements.");
+                        //Logger.Log("No tile contains objects of type '" + desired.objectType + "' to satisfy job requirements.");
```
zwrawr: Instead of commenting that out you could do Logger.VerboseLog()

Mizipzor:
Spaces, not tabs. :)

Randshot:
@zwrawr 
I will try that :)

@Mizipzor 
Oops. I used the builtin editor. It seems like that one uses tabs instead of spaces.
I will correct it shortly.

petroVARS:
I am working on the same problem, but without using a timer. Maybe wait with accepting this for a bit.

Randshot:
@zwrawr Unity says, that there is no Logger.VerboseLog() ...

Randshot:
So, here are a couple comparisons of the old and new versions.
(Sry for the large images, don't know how to scale them down.)

**Old Version**
_1 Chars; 1 Jobs_
orig1jobs1chars
_1 Chars; 5 Jobs_
orig5jobs2chars
_4 Chars; 4 Jobs_
orig4jobs4chars
_4 Chars; 5 Jobs_
orig5jobs4chars
-----------------

**New Version**
_1 Chars; 1 Jobs_
new1jobs1chars
_1 Chars; 5 Jobs_
new5jobs2chars
_4 Chars; 4 Jobs_
new4jobs4chars
_4 Chars; 5 Jobs_
new5jobs4chars

Assets/Scripts/Models/Character.cs
```diff
@@ -121,6 +121,9 @@ public Tile JobTile
     /// Our job, if any.
     Job myJob;
 
+	/// Job wait timer (to lessen material checks)
+	int jobWaitTimer = 0;
```
Dormanil: I think you missed some tabs here. ;)

Assets/Scripts/Models/Character.cs
```diff
@@ -191,29 +194,31 @@ void GetNewJob()
...
-        {
-            GetNewJob();
-        }
-
-        if (CheckForJobMaterials()) //make sure all materials are in place
-        {
-            // If we get here, then the job has all the material that it needs.
-            // Lets make sure that our destination tile is the job site tile.
-            DestTile = JobTile;
+        if (jobWaitTimer > 0) {
```
Dormanil: Braces should always be on new lines. Always.

Assets/Scripts/Models/Character.cs
```diff
@@ -191,29 +194,31 @@ void GetNewJob()
...
 
-            // Are we there yet?
-            if (CurrTile == DestTile)
-            {
-                // We are at the correct tile for our job, so 
-                // execute the job's "DoWork", which is mostly
-                // going to countdown jobTime and potentially
-                // call its "Job Complete" callback.
-                myJob.DoWork(deltaTime);
+            if (CheckForJobMaterials ()) { //make sure all materials are in place
```
Dormanil: Comments are on new lines as well, have a space between `//` and the text, the text starts with a capital letter, and ends with a period.

Randshot:
Anything else ;)

petroVARS:
I implemented a pause function for jobs. Every job that does not have the required materials will not be inserted into the queue, but into a waiting list instead. Whenever a new inventory is created, the list is checked whether a job can be restarted (there was already a callback for this). I think this should at least be a more elegant version instead of using a timer. 

Regarding performance, I am not sure if there is any notable difference to your version. For me, the fps counter seems to be bugged, I often get only 30 fps while absolutely nothing is going on while the stats window shows around 300 fps. Anyway, the fps does not drop significantly anymore. I logged every A* path call and it appears that only the minimum number of calls are made. I will make a pull request after i got some sleep and tested it a bit more .

Randshot:
@petroVARS I find this very amusing since I also thought of doing something similar to this xD
I didn't know about the already existing callback function though.

Randshot:
@petroVARS I think I know why you only get 30 fps. You may have the VSync option 'Every Blank' enabled?

vogonistic:
@Randshot I'm accepting this PR now, but I want you to make a follow up PR that replaces `jobWaitTimer` with the creation of a Waiting job instead. Also replace `JobQueue.QuickCheck` with something that check's if the key exists but is of length 0, since if that material has ever existed, `QuickCheck` will always say true.

Mizipzor:
@vogonistic Im still working my way through the notifications, but if there isnt already a pull request for that could we create an issue just so it isnt forgotten?

vogonistic:
I only created #482 for now.

PR #457, opened 2016-08-20
Add ice to asteroids

JosephBywater:
Hi, I've added an ice item to the asteroid spawning. It should be 20% ice and 80% raw iron now. I don't know if I've implemented it right though. I was wondering if you could check it through and see if I've done anything wrong :P Thanks.

Also, it currently only adds 1 ice item to each tile that it spawns on.

screen shot 2016-08-20 at 23 16 20


vogonistic:
Our conventions doesn't have a space between the method name and the start parenthesis. Make sure MonoDevelop / Visual Studio is configured correctly for the project.

Increase the stack size just a little bit. At least 5 or 10. 

Otherwise it looks good! Fix the issues and I'll merge it.

JosephBywater:
oops...

JosephBywater:
@vogonistic they should be fixed now.

JosephBywater:
Fixed and squashed so it's all in one nice commit! @vogonistic

vogonistic:
Looks good!

PR #459, opened 2016-08-20
[WIP] Scene for testing pathfinding speed.

tickinginstant:
I wrote a separate scene, \_TEST\_, that makes it possible to run some pathfinding speed tests. (It's in a separate scene because it creates multiple worlds, and a lot of the controller code is written based on the assumption that there is only one world.) The interface is very rough at the moment, but it looks like this:

pathfinding test screenshot

It logs all of the results of the tests you do in the file log.txt in the Assets directory, so that it's easier to compare across different branches, and it's harder to lose information from previous runs. It also tries to dump your .git/HEAD file to the log so that it's easier to remember which tests were from which branch. (You have to compare the values manually, though.) It's not particularly user-friendly, but it may be useful for development/testing purposes.

It tests pathfinding on two very basic worlds at the moment: the empty world, and a world with a single 4-wide corridor snaking back and forth. It does four different types of tests for each world: two tests for finding paths to another location (near and far), and two tests for finding paths to the nearest inventory (where one is near, and one is far).

A section of the switchback corridors used for the test:

snaking corridors

(Note that it doesn't actually display the world as it does the pathfinding. This is just to show what the world would look like if you loaded it in the main game.)


tickinginstant:
Don't accept this pull request! Another recent pull request added references to Controller code into the Models section, so now this just generates errors. I'll see if I can figure out a fix.

Mizipzor:
Maybe put **[WIP]** in the title of it? Or something similar. :)

tickinginstant:
Good idea! :) It should be back to working again, but I want to polish it some more.

tickinginstant:
I'm going to close this PR until I have the time to buckle down and get the code into a nicer state. Sorry for letting it linger so long!

vogonistic:
Are you on Discord? We should talk pathfinding and testing.

Mizipzor:
I though this was in a pretty good state. Maybe someone can pick it up? I would really like to see this happen.

tickinginstant:
If someone else wants to pick it up, that's fine by me. It's just that I'm gearing up to do Ludum Dare this weekend, so I won't be able to focus until that's out of the way. :)

PR #460, opened 2016-08-20
Power System - Fixed supply reduction issue

TheRiven:
Fixed issue where reducing the supply of power did not cause furnitures to run out of power if they had already been "powered" by the system. Now when supply is reduced the consumers must re-register for power.

As part of this the registration process has been improved and should now only be called when the furniture first needs to use power or when its waiting for power to become available. As soon as power is available it will then be able to run rather then waiting till then next update.

### DIFF
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -213,10 +213,6 @@ public class Furniture : IXmlSerializable, ISelectable
         {
             World.current.powerSystem.RegisterPowerSupply(this);
         }
-        else if(powerValue < 0)
-        {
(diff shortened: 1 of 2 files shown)


Tritaris:
Yeah this is a better fix than what I had. After reading your last comment on #434 and thinking about it, I now see how you are setting the system up. I wasn't considering the check for power in 'RegisterPowerConsumer' before they were added to the 'powerConsumers' list. 

I was thinking about it and 'PowerSystem.PowerLevel' only needs to be set by 'CalculatePower()' maybe we should set it to 'protected set' , what do you think?

TheRiven:
Hrm... I guess this is no longer relevant as the power system just got refactored

PR #479, opened 2016-08-21
fixes #477

### LINKED ISSUE #477: [Bug] Floor building with PR#267
This merge #267 introduces a bug when building the floor. When placing a row of flooring south from the starting area. The jobs start at the end of the column but are unreachable so they throw an error: `ArgumentNullException: Argument cannot be null.`  in `Character` line 177 and spawns in anyway.

1b
2b

Randshot:


Assets/Scripts/Models/Character.cs
```diff
@@ -174,8 +174,8 @@ void GetNewJob()
         if (myJob.adjacent)
         {
             IEnumerable<Tile> reversed = pathAStar.Reverse();
-            reversed = reversed.Skip(1);
-            pathAStar = new Path_AStar(new Queue<Tile>(reversed.Reverse()));
+            reversed = reversed == null ? null : reversed.Skip(1);
+            pathAStar = reversed == null ? pathAStar : new Path_AStar(new Queue<Tile>(reversed.Reverse()));
```
vogonistic: These two lines would have been cleaner with an if-statement.

vogonistic:
What does the change in `Path_TileGraph.cs` do to this bug?

vogonistic:
I get 4 errors each time it tries and fails to find the path to a job with this change.

vogonistic:
Are you on Discord? We've setup an unofficial here: https://discord.gg/68hkpSA

Randshot:
Yeah I do have Discord.

vogonistic:
Fixed in #481, closing.

PR #483, opened 2016-08-21
Made WorldGenerator getting it's settings from an xml file

gunthergun:
Also made it easy to add new ressources to an asteroid, due to the xml files!


RodRitter:
Looks good. Merged.

PR #486, opened 2016-08-21
Fix downgrade prio of job actually increasing it. Temporarily added debug-view of  entire job-queue until there is a proper UI-interface for it.

bitope:
Fix downgrade prio of job actually increasing it. 
Temporarily added debug-view of  entire job-queue until there is a proper UI-interface for it.


gunthergun:
Can you add a screenshot of the debug view ?

bitope:
Might actually remove it entirely instead, and just leave it in my local branch. No need for everyone to see it all the time and i didn't add a toggle for it.

PR #494, opened 2016-08-21
Added MathUtilities that contains helper methods to work with floating point numbers

Grenkin1988:
It is bad practise to compare floating point numbers directly.


bjubes:
What's the difference between your isZero and unity's [mathf.approxametly](https://docs.unity3d.com/ScriptReference/Mathf.Approximately.html)

Grenkin1988:
At Least that you do not need to include ```UnityEngine```

bjubes:
fair enough. I like the other ones thou,` myValue.IsZero` is very clean

fre-ber:
How do we feel about extension methods? Personally I am conflicted on the issue. Sometimes they make things cleaner but on the other hand they can also cause confusion.

bjubes:
I think extensions can be very helpful. Often they make repetitive code cleaner (as they would have to be rewritten in many different files) and can hide complex logic. For example, most people don't care or know about the precison of floats know what `Float.Epsilon` is, so extensions would hide this logic from them, leaving them with  a clean `myValue.IsZero()` Otherwise, someone may have a hard time understanding the powerValue code because they don't know what elipson is and why it keeps appearing.

Grenkin1988:
Merged all conflicts

vogonistic:
Merged!

PR #495, opened 2016-08-21
Some much needed bug fixes in the worldgenerator

### LINKED ISSUE #493: [BUG, Easy] There is more ice spawned then maximum allowed
The maximum is 10 but due to the worldgenerator it spawns more than the maximum. I have no time to fix this now. But it is a really quick fix :)

gunthergun:
Simple fix... stacksize gets set maxstacksize if it is over max


WardBenjamin:
Thanks! Merging!

Issue #435, opened 2016-08-20
Naming standards


racerand:
I am working on setting the right text on the different inventory items selection_panel info. This is made hard by their not being a naming convention. Only few inventory items have been added "Steel Plate" and "Raw Iron" but shouldn't they be called "inv_steel_plate" and "inv_raw_iron" in code? As they all will go through the Localization system to have the right UI text, a very informing naming standard in code seems useful.

`Steel Plate=Steel Plate` doesn't feel good in code.

Sommerbo:
Agreed! This needs addressing soon.

racerand:
I am working on implementing it on what already exists, so if this it what people like, the already coded stuff will be done.

racerand:
Could also be "inv_steelWall" as the standard.

racerand:
Or "inv_SteelPlate" which would be same standard as for furniture

svmnotn:
I think we really need to go over what the standards are for code names vs shown names, because as it stands the art uses the en_US names, other things use the objectType, and then the localization uses something entirely different.

TomMalbran:
It might also be a good time to implement an xml with the inventory types metadata (similar to the one for furniture), which for now it might only include the object type ("inv_Name") and the max stack size, and later additional settings per inventory type can be added to that file.

racerand:
I agree with you @svmnotn. The images right now use the en_US names, which is also currently the same as the objectType. Localization right now would match "inv_steel_plate" which I like. But in furniture there is currently no set naming conventions, as quill started with but never fully implemented "furn_SteelWall", and later commits have use "Power Generator". So just in general I think we need to take care of naming standards soon.

svmnotn:
@racerand as do I, thus why #451 was made ;)

vogonistic:
Closing old issues for inactivity

PR #505, opened 2016-08-21
[WIP] [POLLS] Furniture and Inventory naming standards. Fixes #451 and #435

### LINKED ISSUE #451: Non-Code Naming Standards
Currently nothing really follows any set standard as #435 brought up.
So, here are the ones I'm proposing and lets discuss this:
- Anything inside Furniture.xml is named as follows: `steel_wall` afterwards, in code the `furn_` can be appended at the beginning.
- Lets make something like Furniture.xml for items, so Inventory.xml. The naming here would match furniture, so things inside would be `raw_iron`, and then in code the `inv_` is added.
- For the UI, it is a mess... some of it is doing well and using `menu_` and other parts are not. So let's just use `button_` for all buttons?

Here is also a good place to decide between `furn_SteelWall` and `furn_steel_wall`, as shown I prefer the latter but I can see a lot of the former already in the project (the latter in only in the translation keys).

-------
I will try to update the above as comments come in until this can be formalized into the CONTRIBUTING.md

racerand:
Trying to resolve #451 and #435.

I have updated the Furniture naming standards and is working on implementing inventory naming standards. Also updated CONTRIBUTING.md to descripe this. The new naming standard will be 'steel_wall' in the adding part:

* Imagefiles
* Furniture.xml
* Inventory.xml when that's a thing
* In code

This will only be different in the Localization system to differentiate between the different types, Inventory with prefix 'inv_' and Furniture with prefix 'furn_' as previously. More on this in CONTRIBUTING.md.

This also removes some localization parts of the xml and reader to make this more simple. This uses the ISelectable GetName() function, which has currently been changed to use the objectType instead of name of the furniture.

EDIT: This also completely removes  the to localization parts of the xml.


Assets/StreamingAssets/Images/Furniture/oxygen_generator.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Sprites>
+	<Sprite name="furn_oxygen_generator"     x="0" y="128" w="128" h="128" pixelPerUnit="64" />
```
quill18: Should this maybe be `name="oxygen_generator"`?

Assets/StreamingAssets/Images/Furniture/small_landing_pad.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Sprites>
+	<Sprite name="small landing pad" x="0" y="320" w="192" h="192" pixelPerUnit="64" />
```
quill18: Should this maybe be `name="small_landing_pad"`?

quill18:
Standardizing the naming scheme and updating CONTRIBUTING seems like a great idea.  `naming_things_like_this` indeed seems like a good way to me.

quill18:
I wonder if instead of `steel_wall` we should maybe go with `wall_steel` instead (type, then material).

racerand:
Now the typing mistakes has been changed. Thanks @quill18. The 'type_material' seems like a good idea, will change that. I'm guessing this should only be in code? Should it be the same with 'steel plate' and 'copper wire'? Other inventory?

CONTRIBUTING.md
```diff
@@ -3,6 +3,7 @@
 [Unity Version](#unity-version)  
 [General resources](#general-resources)  
 [Style Guidelines](#style-guidelines)  
+[Adding Furniture and Inventory](#Adding Furniture and Inventory)
```
Tranberry: Should be "[Adding Furniture and Inventory](#adding-furniture-and-inventory)  "

Tranberry:
@racerand sorry for being such an annoying pr**k but: you missed the spacing after the link in CON...md

racerand:
Thanks @Tranberry. Am new to md, so things happen.

Randshot:
I think that we should be consistent with our naming and if we decide to use the 'type_material' format, then we should apply it to all names.

racerand:
I agree on that @Randshot. But what does people think about it? Should we go for the 'type_material' format? Then I will quickly change all the things.

NogginBops:
I vote for type_material.

svmnotn:
@racerand I was going to do this, after more discussion happened in #435 but I don't have to now! Thanks! :D 

Also this shows how quickly this project is moving.

racerand:
Then I will change to 'type_material' then and upload final commit for this one, also updating the description in CONTRIBUTING.md

racerand:
Would this also include generators, such that 'oxygen_generator' would be a 'generator_oxygen' even though oxygen here is not the material for the generator.

svmnotn:
I think it is better if it does include it, since it is just a different way of looking at the material (generator would not be a material, but oxygen would)

racerand:
I would agree, then that's what will happen.

racerand:
We are working on implementing the new objectType naming scheme `type_material` and was looking at `small_landing_pad` and `mining_drone_station` and was thinking they should be `landing-pad_small` and `drone-station_mining`. What do you guys think of this way of doing it, so a multiple word type is seperatet by `-` and `_` only seperates type and material?

Assets/Scripts/Models/World.cs
```diff
@@ -315,7 +315,7 @@ public void SetupPathfindingExample()
                 {
                     if (x != (l + 9) && y != (b + 4))
                     {
-                        PlaceFurniture("furn_SteelWall", tiles[x, y]);
+                        PlaceFurniture("steel_wall", tiles[x, y]);
```
relinu: wall_steel

racerand:
Fixed @relinu. What do you think about what I wrote in my last comment?

relinu:
maybe landingPad_small and droneStation_mining could be better

racerand:
Maybe, but i prefer snake_casing'ish for this, as it's what's used everywhere else.

svmnotn:
Im just going to make this simple and leave all three choices @racerand and I talked about, and then we can pick the one with the most thumbs up?

Here it goes:
`landing_pad_small`

svmnotn:
`landing-pad_small`

svmnotn:
`landingPad_small`

vogonistic:
I think `type_material` works great for resources where there are many similar types, but it's less important for machines.

👍 to `wall_steel`, `wall_inflatable`, `plate_steel`, `plate_copper`, `door_basic`, `door_airlock`

I wouldn't even know what kind of type these are: `oxygen_generator`, `mining_droid`, `landing_pad_small` and I think they work as is.

racerand:
For machines I think it's not that bad. Machines would be:

`generator_oxygen`, `smelter_metal`, `smelter_metal`

And the type for a mining droid station would be droid station, and for a small landing pad, it's a landing pad, When using the `new-type_new-material` or the  'newType_newMaterial` style, it's easy to see the difference, and I would say we need a way to have multible word types and materials connected.

racerand:
I'm going to split this into a  couple PR's starting with the CONTRIBUTING.md, now I just need the final voting [here](http://www.strawpoll.me/11051912).

racerand:
Thank you for all the votes, I will have now made a pr with the finished rules #633.

ComputerOverlord:
@racerand how about mining-station_drone

racerand:
Well that's a good question as to wether it's a mining station or a drone station

Issue #507, opened 2016-08-21
[Bug, Easy] Furniture tab tooltip


kd7uiy:
![capture](https://cloud.githubusercontent.com/assets/5951269/17837214/996d7848-677a-11e6-9694-7e1314e03017.PNG)

Two issues for this bug:

1. Don't include the tag if there isn't furniture (As shown above)
2. Steel Wall has the tool tip "basic wall". That seems off.

PR #509, opened 2016-08-21
Furniture tab tooltip #507

Ohda:
Fix https://github.com/TeamPorcupine/ProjectPorcupine/issues/507

    Don't include the tag if there isn't furniture (As shown above)
    Steel Wall has the tool tip "basic wall". That seems off.


racerand:
You should not be hard-coding this, should be using the localization system.

vogonistic:
Merged!

Issue #482, opened 2016-08-21
Sometimes characters will wait a long time before taking on a job


vogonistic:
This was introduced in #450 and is caused by the characters waiting 20 ticks each time they abandon a job. A good solution to the problem would be for them to try `jobQueue.Count`, or a reasonable number jobs before they decide to take a nap. Each job check when there is missing material or the tile can't be reached has gotten faster now.

@Randshot

petroVARS:
I stated in that pull request that i am working on a version that does not require a timer, but instead pauses jobs. I am still working on it, but I couldn't post a pull request anyway because there are many conflicts now. I am not really sure how to resolve that quickly because I haven't used git before. However, I would suggest that pull requests should not be accepted if someone else is also working on the same issue until both are finished and then it can be decided which solution works better.

vogonistic:
It's always tricky when multiple people are working on the same thing.

Now in the early project, having it done counts for a lot. When things stabilize, we will try to avoid having people work on the same issue.

Randshot:
@vogonistic @petroVARS So, while I was working on this issue, I decided to implement your (vogonistic) mentioned "Waiting Job".
In the end I have come up with a solution (probably) very similar to petroVARS.
Should I post a pull request?

vogonistic:
If you think it's ready, post the PR. Make sure you reference this issue with something like `closing #482`

PR #517, opened 2016-08-21
Better way of fixing the (pathing/job) performance issues

Randshot:
I used a separate JobQueue for jobs that get abandoned because of missing materials.

closing #482


Randshot:
Here is a screenshot 😄 
fix

WardBenjamin:
Looks great! Thanks!

PR #518, opened 2016-08-21
Small performance fix

NogginBops:
Created a accessibility check before trying to pathfind to the jobs tile.


NogginBops:
Thanks git...

PR #522, opened 2016-08-21
Modding managment

ttyz:
Did not plan this but it worked out, so I am very sorry if some one is working on it already.
This is a simple way to load mods.
Currently only loads furniture (xml, lua, and sprites) form `/Assets/StreamingAssets/Data/Mods`.
At some point I would like to add a Mod Manager that will Install and update mods (probably from GitHub)


vogonistic:
Merged!

PR #523, opened 2016-08-21
Renamed CircleCursor to just Cursor to be more generic.

kd7uiy:
I think the best way to make this extensible, at least for now, is to have the file replacable if you want to change the icon. That being said, the best option is probably to just have a generically named cursor, which can be replaced as desired. Changed the name to fit this convention.


jdlaroche:
If we eventually replace the mouse cursor by an image, won't it be confusing?

spexiono:
This could also cause (small) issues if we eventually want the cursor to change when doing a particular task, such as selecting a unit to attack.

kd7uiy:
Fair enough points I suppose, I'll close this.

PR #466, opened 2016-08-21
Coal item, and generator requires coal

wilcomega:
Added a coal item with my original sprite that i created, the coal spawns in the world and is used to power the power generator.

also changed the way things spawn and included Ice in it from someone else's commit.

PR #524, opened 2016-08-21
Uranium added and Power Generator requires Uranium

wilcomega:
Same as the coal one except for it now uses uranium. Sprite made by me. #466


Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -171,8 +171,14 @@
 
       <BuildingJob jobTime="0">
       </BuildingJob>
-
-	
+      
+      <Params>
+        <Param name="burnTime" value="0" />
```
zwrawr: Should be 4 spaces not 2 in this indent, also looks like there's some tab's in this file

Assets/StreamingAssets/Images/Inventory/Uranium.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Sprites>
+	<Sprite name="Uranium"     x="0" y="0" w="16" h="16" pixelPerUnit="16" />
```
zwrawr: tab here

wilcomega:
I fixed it, is there a difference between a tab and four spaces? because i cant see any aside from the binairy difference

vogonistic:
Merged!

ComputerOverlord:
So... It's a fission station

PR #525, opened 2016-08-21
BuildModeController now passes the StyleCop test.

fre-ber:
Removed unused "using", moved private things below public things and changed white space.


vogonistic:
Merged!

PR #530, opened 2016-08-21
Added small accessability check before pathfinding

NogginBops:
This removes the need to make a full A* to find that the tile is not accessible.

### DIFF
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -161,6 +161,14 @@ public class Character : IXmlSerializable, ISelectable
         // Get our destination from the job
         DestTile = myJob.tile;
 
+        // If the dest tile does not have neighbours it's very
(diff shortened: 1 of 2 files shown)


vogonistic:
Nice! Merged.

Assets/Scripts/Models/Character.cs
```diff
@@ -161,6 +161,14 @@ void GetNewJob()
         // Get our destination from the job
         DestTile = myJob.tile;
 
+        // If the dest tile does not have neighbours it's very
+        if (DestTile.HasNeighboursOfType(TileType.Floor) == false)
```
TomMalbran: It should also search if there are Neighbours of tile type Ladder. Right now the Ladders can only be placed adjacent to Floors.

NogginBops: This will be fixed in my next PR that will make tiles not hardcoded.

Issue #537, opened 2016-08-21
[Discussion] XML UI


dustinwloring1988:
I was trying to start on a XML style UI and was wondering what everyone thought about it. This will allow for easy editing with out touching the _SCENE_ file. Also just like the whole concept of this project it will be easy to mod after the project is built.
All input is welcome good, bad or indifferent.

bjubes:
The only concern is that there is A LOT of stuff unity takes care of when making UI. Nine sprite splice, anchors, parent and child anchors and positions, etc....

If this is changed to XML, the work has to be done manually, which would take much longer. Its the same problem with making Spritesheets in XML, its a pain to type out a file of each wall and specify each x,y,width,height,ppu of each variation. This will be nothing compared to the tedious nature of making an entire GUI, which currently is  23326 lines in unity's scene file

fre-ber:
@bjubes I think that you are correct about the UI it would be quite tricky to get a flexible UI this way. As for extracting sprite sheets from the meta-files, I made a little script for that, posted in the comments of episode 64 of the tutorial video series. Perhaps it can be useful. I haven't tested it with the latest version of Unity, though. But if something changed, it shouldn't be too hard to modify the script to make it work again. Perhaps we should add a folder for external tools like this in the repo.

ttyz:
I think the best solution is to make some reusable components, like dialog, buttons, fields, which will be stored as prefabs, or preferably generated on the fly, and only reference those in the xml files:
```xml
<Dialog title="Save Game">
    <FileList name="fileList" />
</Dialog>
``` 

Trying to replicate a UI system with the same level of flexibility Unity has is too big of a task I think.

vogonistic:
One big reason to have something that is entirely covered outside of the scene (but might be assets, idk) is modding. If we want them to add buttons, dialogs, move things around or generally do really cool things we haven't thought about, we need a system that is flexible enough for that.

We should implement our own UI as a mod, IMO.

b4ux1t3:
Would it not be possible to have some kind of parser for the XML files, but that still uses Unity's UI controls behind the scenes? If I'm not mistaken, that's how Kerbal Space Program does it.

vogonistic:
@b4ux1t3 I think that was what @ttyz was suggesting.

dustinwloring1988:
[...]

PR #540, opened 2016-08-21
[WIP] XML UI

dustinwloring1988:
#  Code Driven UI - XML UI
**PLEASE DO NOT COMMIT THIS YET**

This will take some time to do but I think it will be time well spent. This XML UI will allow for easy editing with out touching the _SCENE_ file. Also just like the whole concept of this project it will be easy to mod after the project is built.

It will be able to build UI from just button, drop downs, lists, etc.. it will be able to spawn p remade UI (ie. Dialog boxes, custom prefabs). I opened up a discussion on this here, - https://github.com/TeamPorcupine/ProjectPorcupine/issues/537 -, please comment on the discussion page not this page.

Thanks,
Dustin


Assets/XML_UI/Scripts/UIReader.cs
```diff
@@ -0,0 +1,15 @@
+﻿using UnityEngine;
+using System.Collections;
+
+public class UIReader : MonoBehaviour {
+
+	// Use this for initialization
+	void Start () {
```
zwrawr: Tabs!

dustinwloring1988:
Below is my settings and code view I think there set right please correct me if I'm wrong. But I think it looks different on my computer than when I do a pull request.

settings
codeview

Mizipzor:
You MonoDevelop settings are correct. What I think happened here is that Unity generated that file for you, and then you never touched it. Theres a "format document" item in the menus. Try that.

PR #542, opened 2016-08-21
Changed the world generator and its xml

gunthergun:
so that it allows different stackamounts to spawn.


gunthergun:
Fixed the conflicts with the latest version

gunthergun:
Another sideffect we have now is with this fix removed as it was more likely for ice to spawn with 10 instead of other stackamounts

vogonistic:
Merged!

PR #545, opened 2016-08-21
Converted public fields to properties for Tile.cs

behindcurtain3:
Converted all public fields in Tile.cs to properties and updated them to use PascalCase naming.


Randshot:
Did we agree on also using PascalCasing for object names?

behindcurtain3:
Style Guidelines say to use PascalCasing for properties. I agree the naming can be confusing but I didn't feel I should be renaming properties without some discussion.

Mizipzor:
@Randshot 
> you can easily mistake classes and objects off classes.

Could you elaborate here? I cannot come up with any scenario where thats the case

As for what is the standard I suppose its till up for debate. That looooong thread about it sort of died of with no consensus and hasnt been closed yet. Some code got added to the contribution file but I dont think that makes it set in stone. I added a settings file to automatically check the style but neither that makes it final. As far as Im concerned everyone still do whatever they think is best. Maybe because of that we should avoid these big wide changes?

As for what style is best, that is something you can discuss til the end of time. :) No matter what you suggest there will be a handful of people _strongly_ opposing it, whatever it is. 

So what did I try to achieve with adding StyleCop?

* It's a third party so one here will have their way and no one has to give in, minimizing grudges.
* Yet, its not completely inflexible as we can (and we do) turn rules on or of.
 * Those rules can _only_ be on or of, not modified so it cuts down on discussion and can be boiled down to a vote.
* Correctly understanding style from a description is difficult and leaves room for interpretation, for StyleCop a line is ether good or bad, never something in between.
* Do you _really_ want to go through every line you wrote to make sure it conforms to the style? Of course not, thats boring. StyleCop checks all our code in seconds.
 * How about someones elses lines in a pull request? Thats even more boring.
* Do not underestimate how awesome it is to have this done automatically, whatever other style you pick, no matter how awesome, you need to be the human style cop.

As for this particular case? This is actually the first project I've come across that didnt flat out ban every public (non static) field on day one. As a rule, everything public with uppercase, anything private with lower case unless its a method.

WardBenjamin:
Looks good, but has conflicts. Fix and I'll merge this.

quill18:
@Randshot @Mizipzor CamelCase for properties is good, but I can see the source for confusion:  If a Tile has a property called Furniture, which is a reference to the Furniture class...that's bad.

I suspect the correct move is to rename the property to something like "MyFurniture".

Mizipzor:
@quill18 I agree that it is confusing, but believe it or not it is vaguely recommended by the [Microsoft naming standard}(https://msdn.microsoft.com/en-us/library/ms229012(v=vs.110).aspx):

> ✓ CONSIDER giving a property the same name as its type.
> For example, the following property correctly gets and sets an enum value named Color, so the property is named Color:

quill18:
I think just "Furniture" is fine.  As such, that means this PR can probably be merged.  The sooner we eliminate public fields, the better.

quill18:
Wait!  You need to modify the LUA files to use the newly capitalized public properties.

The automatic refactoring tools won't do that for you.

Lines like: `	if( furniture.tile.inventory == nil ) then`

Mizipzor:
Added title tag so it isnt merged.

behindcurtain3:
@quill18 thanks for pointing out the lua files. I had forgotten to update them! Everything for the Tile class should be updated and working now.

vogonistic:
Merged!

PR #548, opened 2016-08-21
Ensure remove designation removes both the image and job

### LINKED ISSUE #546: [Bug] Removing designation removes the designation image but job still exist somewhere
If you setup the designation for a steel wall for example and then remove them. The image goes away but if you don't add any new jobs the characters go and still build those designated walls but we do not see them. I looked into it and see that when CancelJob is called there is no job in the jobqueue when this happens. So that means the job we are trying to cancel cannot be removed from queue, this means probably the character cached it somewhere. The strange part is though that the character does other jobs first. I do not know the code for the character very well but do they plan ahead and keep jobs somewhere as this was working before ?

Randshot:
Fixes #546


Randshot:
resolved conflicts

PR #549, opened 2016-08-21
Wrapper functions for set and get in Settings.cs

zwrawr:
Wrapper functions for set and get , should make useage of the settings sytem easy and clean


WardBenjamin:
Looks good, thanks!

Issue #532, opened 2016-08-21
Funiture params


sboigelot:
Furniture params can only be float.
We should expand this to allow at least string as a valid param.

fre-ber:
A simple way could be to add another dictionary, stringParameters, but as came up in the atmosphere discussion, we might need a more flexible parameter system that would allow us to specify things like a string AND a float together in a single parameter.

Turning all parameters into strings and have them encoded in json or something would be a possibility, but a possibility that I don't particularily like.

FilippoLeon:
what about:
`<Param name="oxygen" type="float">55</Param>`

store everything as string inside a structure:
`
Dictionary<string, structure_similar_to_tuple>
`
where the key is a string (name) and the value is a "tuple" value + type
and convert using a function:
```
GetParam<T>(string name) {
   read type and value
   do appropriate cast
   check compatibilty with T
   complain or return value
}
```

koosemose:
I have an additional idea that may be useful for some furnitures (such as O2 Generator), that will however require a bit more complexity than @FilippoLeon's current WIP PR allows. Basically ParameterLists (or ParameterGroups) allowing multiple parameters to be grouped and accessed as such, and a very important part is for them to be able to be stored similarly (I'm thinking subclassing would be the most reasonable way to do this). An example of what the XML would look like from what I'm thinking for O2 Generators:
```
        <Params>
            <Param name="gas_limit" value="0.2" />
            <Param name="gas_per_second" value="0.16" />
            <ParamList name="gas_gen"> 
                <ParamList name="O2">
                    <Param name="gas_limit" value="0.2">
                </ParamList>
                <ParamList name="N2">
                    <Param name="gas_limit" value="0.8">
                </ParamList>
            </ParamList>
                        
        </Params>
```

I'm still working out how exactly this would be accessed from the code side (mostly because I don't think LUA can handle some of the C# I'd want to use in my optimal case), but it would be something along the lines of `furnParameters.GetValue("gas_gen").GetValue("O2").GetValue("gas_limit").AsFloat()`.

I'd prefer to implement it with an indexer (access as `furnParameters["gas_gen"]["O2"]["gas_limit"].AsFloat()`) but I doubt MoonSharp handles that well or at all.

FilippoLeon:
What about implementing a type list or dictionary which will be read into a custom class that mimics  the standard list behaviour?

Would be
'''
< Param type=list name=gas>
< entry key=n2>0.2<\>
[...]

Issue #515, opened 2016-08-21
Discussion: Furniture events


FilippoLeon:
Right now, we can have the furniture items respond to Update() events with an OnUpdate() LUA function.

It is conceivable that a furniture object responds to many different events, such as OnInstall, OnUninstall, OnHitByProjectile?, etc.

I think, right now, there is no plan to change this behaviour. If there is no plan to address this issue I propose the following.

I propose a more flexible way to handle this events, using a generalization of the OnUpdate business.

The Furniture.xml items will have elements of the form
```
<Action event="OnUpdate">callThisLuaFUNC</Action>
<Action event="OnInstall">callThisOtherLuaFUNC</Action>
```

etc.

This is then parsed by the XmlReader and handled in code wherever necessary. We should then offer a documentation on when/how this "events" are triggered.

bjubes:
So you propose we add On"event-name" tags you can put in the furniture XML and then implement them in the XML Parsing code so that lua code can react to different furntiure events? Kinda like:
```

 <OnUpdate FunctionName="OnUpdate_Door" />
 <OnUninstall FunctionName="OnUninstall_Door" />
 <OnDamaged FunctionName="OnDamaged_Door" />

```
If this is what you mean, it sounds good to me!

FilippoLeon:
Forgot to tag the xml code I wrote as code ,fixed now.

The idea precisely that one.

FilippoLeon:
Closed in favor of https://github.com/TeamPorcupine/ProjectPorcupine/pull/682.

PR #552, opened 2016-08-21
[WIP] [Breaks Things] Refactoring furniture event system and parameter system

FilippoLeon:
This is aimed at addressing https://github.com/TeamPorcupine/ProjectPorcupine/issues/515  and https://github.com/TeamPorcupine/ProjectPorcupine/issues/532.

FIXES https://github.com/TeamPorcupine/ProjectPorcupine/issues/532 and FIXES https://github.com/TeamPorcupine/ProjectPorcupine/issues/515

The idea is to change a bit the xml/lua exposed interface in order to have a more flexible and moddable furniture system (and possibly everything else (Character, etc. could be all child of a common Class).

Changes:
 - parameters are now stored as dictionary of a small class containing value and type and possibly more
 - they can store virtually any type (string, float)
 - new methods will set/get the parameter as specified type (after mandatory check):
```
GetParameterAsFloat(name, value)
ChangeParameterAsFloat(name, value)
```
 - new event system might specify custom event:
`
<Action event="OnUpdate" function_name="lua_function_name"/>
`
- stored as `Dictionary<string, EventAction>`, where EventAction is a class containing the list of lua function to be called for that event.
- TODO: within LUA or c#, possibility to trigger an Event by calling something like
```
Trigger("OnUpdate")
Trigger("OnMyCUstomEvent")
```
or similar.

TODO:
 - documentation
 - test if changes break things (likely so)
 - get consensus on new interface


Assets/Scripts/Models/Furniture.cs
```diff
@@ -541,62 +561,124 @@ public void ReadXmlParams(XmlReader reader)
...
+    public float GetParameterAsFloat(string key, float default_value = 0f)
     {
-        return GetParameter(key, 0);
-    }
+        if (furnParameters.ContainsKey(key) == false)
+        {
+            return default_value;
+        }
 
+        return float.Parse(furnParameters[key].value);
```
mmooxy: Maybe check if the parameter is of type float first? I also don't really like the fact that there will be more functions per type of parameter (not sure if there is an alternative).

bjubes:
Yeah maybe one SetParameter() that's overloaded with both float and string

FilippoLeon:
will MoonSharp be able to call the correct function?

FilippoLeon:
I came to the conclusion that having methods ending with "AsType", will be clearer and safer, even tough a bit more verbose.

It is likely that beside string, int float and bool, there are not many other parameters, so the clutter inside Furniture is not too bad. On the other hand, we can do more safe checks and the intent of the code within the LUA script is much clearer.

Let me know what you think.

mmooxy:
Alright, implement the type checks and the other types. Although I'd like to know if there are memory problems if we have hundreds of booleans saved as strings (no expert on this).

FilippoLeon:
Well, for the saving, we could differentiate based on the type, and store them in an appropriate dictionary of appropriate type (that could be also more elegant). But I doubt it would influence the memory requirement by much, and the bottleneck (memory-wise) would be the Dictionary keys anyways.

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -50,8 +50,8 @@
             <Inventory objectType="Steel Plate" amount="1" />
         </DeconstructJob>
 
-        <OnUpdate FunctionName="OnUpdate_Door" />
-        <OnUpdate FunctionName="OnUpdate_Leak_Door" />
+        <Action event="OnUpdate" function_name="OnUpdate_Door" />
```
TomMalbran: I think that we should use camelCase for the attribute names since it used for all the other attributes. ("functionName" instead of "function_name")

FilippoLeon:
Closed, cf. https://github.com/TeamPorcupine/ProjectPorcupine/pull/682

Issue #440, opened 2016-08-20
Disallow commented out code?


Mizipzor:
I'm not talking about documentation comments like these:

```
// this is an int
int i;
```

Rather, actual code that has been "removed" by being placed in a comment block. Like this:

```
//if (x > 0)
//    Do(x);
```

I'm seeing this in too many pull requests and would like to see a rule against it added to the contribution guidelines. This is one reason why Git is used, so you can go back and see what the code did before. It does nothing but add to clutter. If you use this trick to keep code around in case you want to go back to it (or something), I'm here and willing to teach you some proper Git. :)

Do we vote? Do a thumbs up or down on this post.

bjubes:
I think we should disallow this if the code is no longer needed and the feature is mostly complete. However I don't think we should deny PR due to small comments like this, only when it adds lots of clutter.

Mizipzor:
My argument is that they _always_ adds to clutter and never adds to readability. :)

Grenkin1988:
@Mizipzor Strongly agree. Commented out code should not be in repository.

vogonistic:
Yeah, commented out code should go unless there is a very compelling and documented reason for it to stay. The biggest exception would if an API has some code showing how it can be used.

Randshot:
I also agree, except for logging calls.
I think that they should be commented out instead of entirely removed.
Just my opinion though ;)

Mizipzor:
@Randshot instead of commenting out, how about logging to the verbose level?

Randshot:
That's ok too :)

vogonistic:
@Randshot @Mizipzor The problem with these kinds of logs is that you only want them when debugging one particular feature, like `job queue` or `pathfinder`. If we make all those logs Verbose, then you'll have to turn on verbose logging everywhere which would probably drown out your logs. It's better to keep commented out logs than turning everything verbose, but I'm not really happy with those either.

I often use something like this:
```cs
const traceLogThisModule = false;
function trace(params object[] things) {
    if (traceLogging) {
        // Debug.Log or Logger.Info or something here
    }
}
```

I haven't done it in C#, but I've done similar things in other languages. The compiler should be smart enough to optimize the calls away when turned off, but it allows you to turn on logging for just one particular subsystem when working on it and then turn it off so that nobody is bothered.

These kind of calls you don't even have to comment out and they'll most likely have some value to the next person testing your system.

[...]

PR #558, opened 2016-08-21
StyleCoped scripts/models/*.cs

zwrawr:
Improved Style of scripts in the models directory. 

Did not rename methods/Varibles or reorder public/private

### DIFF
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -6,14 +6,14 @@
 // file LICENSE, which is part of this source code package, for details.
 // ====================================================
 #endregion
-using UnityEngine;
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Xml;
 using System.Xml.Schema;
 using System.Xml.Serialization;
 using MoonSharp.Interpreter;
-using System.Collections.Generic;
-using System.Linq;
+using UnityEngine;
 
 /// <summary>
 /// A Character is an entity on the map that can move between tiles and,
@@ -24,7 +24,6 @@ using System.Linq;
 [MoonSharpUserData]
 public class Character : IXmlSerializable, ISelectable
 {
-
     /// <summary>
     /// Returns a float representing the Character's X position, which can
     /// be part-way between two tiles during movement.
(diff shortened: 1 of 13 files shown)


Grenkin1988:
You are going to have so much conflicts I think :-) You should do it in smaller parts.

svmnotn:
no conflicts if someone merges it right now ;)

Grenkin1988:
Than everyone else would have them

bjubes:
The code will be clean, but everyone will hate you :)

WardBenjamin:
To be honest, I'm in favor of doing this now while it's not too, too intrusive. 

The only issue I see is that in a bunch of places, there's now 4 forward slashes in comments. (i.e.: `////`).

Grenkin1988:
This is still not finished, It would be better just to force people style file ehen they change it.

zwrawr:
the //// is how StyleCop wants commented out code to be

Grenkin1988:
There should be no commented out code #440

zwrawr:
Before i commit that , Removing commented code will be removing commented out Debug.Log/Logger.Log is that okay?

Grenkin1988:
@zwrawr that people still discussing.

zwrawr:
I made a branch off this and removed commented out code , will merge if people want that

Mizipzor:
I would advise against doing this, as others have pointed out it might be problematic for other pull requests. It is much better to do cleanup in a file youre working in anyway. Preferably as a separate commit if its much, so you can look at just the diff of the other commit to see what was actually done. 

If you really want to do some cleanup, then do one file at the time, try to find files that no one works in.

I appreciate the initiative though. :)

WardBenjamin:
@zwrawr That's definitely an issue then. Is there a way to change that setting? Plus, it //// commented the actual comments people put there, which I don't like. I guess tbh those should be /**/ block comments, but still.

Mizipzor:
There's no way to change, we can turn the rule off or leave it on. This is a good thing, cuts down on discussion. For the comments though, is it really that important? Considering what we would give up by turning if off. Everyone is used to something.

WardBenjamin:
Alright, I'm okay with this now. Merging!

PR #562, opened 2016-08-21
Disabled StyleCop SA1513

Mizipzor:
Heres a case where I think StyleCop is trying too much. The rationale is that space between code increases readability by decreasing clutter.
	
> SA1503 : CSharp.Layout : The body of the if statement must be wrapped in opening and closing curly brackets.

	 if (i < 0 || i > rooms.Count - 1)
	     return 
	 return rooms[i]; // not good

But doing the obvious is not enough because then another rule pops up.

> SA1513 : CSharp.Layout : Statements or elements wrapped in curly brackets must be followed by a blank line.
	
	 if (i < 0 || i > rooms.Count - 1)
	 {
	     return null;
	 }
	 return rooms[i]; // not good either

So it wants this:

	 if (i < 0 || i > rooms.Count - 1)
	 {
	     return null;
	 }
	 
	 return rooms[i];

I think both rules tries to address the same problem in the same place. And it results in a double dip, a tad too much.

I think this is one of the (rare) cases where its best left up to the individual developer. This commit disables the latter rule.

### DIFF
--- a/Settings.StyleCop
+++ b/Settings.StyleCop
@@ -265,5 +265,15 @@
       </Rules>
       <AnalyzerSettings />
     </Analyzer>
+    <Analyzer AnalyzerId="StyleCop.CSharp.LayoutRules">
+      <Rules>
+        <Rule Name="ClosingCurlyBracketMustBeFollowedByBlankLine">
+          <RuleSettings>
+            <BooleanProperty Name="Enabled">False</BooleanProperty>
+          </RuleSettings>
+        </Rule>
+      </Rules>
+      <AnalyzerSettings />
+    </Analyzer>
   </Analyzers>
 </StyleCopSettings>


racerand:
I think the space is actually pretty nice looking, good spaced code is always nice for an overview.

Randshot:
Actually, if there is one thing I don't like about this formatting, then it is that the opening curly braces have to be on a new line.

tickinginstant:
I like the space after the closing curly brace, at least in this example. But then, if I were trying to pick out the most readable form of that code, I would have picked the format (not shown) with no curly braces, and a blank line after the single-line if block. So it might just be that I have weird taste.

ttyz:
I agree with StyleCop on this, empty line after control structure is good, curly braces are great, seems fine for me.

Mizipzor:
Oh look at that I'm in the minority here. :) 

The best thing about leaving this on is that it is the default, the less deviation the better as far as I'm concerned. 

I'm going to bed now, ill give others til the morning to weigh in, otherwise I'll close it.

WardBenjamin:
Looks fine, merging!

Mizipzor:
Doh! Maybe I should have changed the title. Thought it was clear from the comments. I think this was voted down and shouldnt have been merged. I'll create a new pull request reversing this. :)

PR #564, opened 2016-08-21
[Small] Fixed an error message that contained old method name.

b4ux1t3:
OnTileTypeChanged was renamed to OnTileChanged. I just fixed a reference to that old method name in the error logging, because I couldn't find that method when I got the error message.

### DIFF
--- a/Assets/Scripts/Controllers/TileSpriteController.cs
+++ b/Assets/Scripts/Controllers/TileSpriteController.cs
@@ -113,7 +113,7 @@ public class TileSpriteController : MonoBehaviour
         }
         else
         {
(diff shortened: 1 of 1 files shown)


Mizipzor:
As with code comments logging lines are usually forgotten when code is changed. That's why you should avoid method names in log if you can. It's fine if it's really important.

b4ux1t3:
Yeah. I don't know if monodevelop has an option in their refactoring tool to also target strings and comments. Visual Studio does. It's super useful for exactly this reason.

WardBenjamin:
Looks good!

For future reference, for small PRs like this you can add [Small] to the title and gatekeepers should have an easier time merging!

PR #567, opened 2016-08-22
Fix for metal smelter issue

### LINKED ISSUE #559: Metal Smelter only makes plates when full
If you put anything less than 50 iron in the smelter, it won't do anything, but once the metal smelter has a full 50 iron, it makes plates until no iron is left

Randshot:
fixes #559 

If you wonder, why there are so many changed lines, that's because my editors (tried 3) wouldn't save the file without reformatting some of the lines :/

### DIFF
--- a/Assets/StreamingAssets/LUA/Furniture.lua
+++ b/Assets/StreamingAssets/LUA/Furniture.lua
@@ -178,7 +178,7 @@ end
 
 
 function MiningDroneStation_UpdateAction( furniture, deltaTime )
-	
+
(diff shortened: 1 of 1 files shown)


WardBenjamin:
I think the newline issues are eol related, since that also happened on my PC. Merging!

Mizipzor:
@Randshot For the record I think it is correct, trailing whitespace is not useful to anyone anywhere, They should always be removed.

Randshot:
@Mizipzor I think we should also decide on a style guideline for Lua files, because at the moment it is just a mess with tabs instead of spaces, etc...

PR #570, opened 2016-08-22
[Small] Make 'Loaded localization' log message LogVerbose

bjubes:
currently localization spams each lang file to the log on start. Changed so that this only occurs in verbose mode.

Bonus: added space to preceding comment


WardBenjamin:
Thanks! Let's hope that the [Small] tag catches on to make life easier for the gatekeepers!

PR #543, opened 2016-08-21
Add ZeroG Ladder floor type

abackwood:
Adds a ZeroG Ladder type floor, much like the rungs and handholds seen on the outside of real-life space stations for astronauts to move around.

Eventually, this tile is meant to have the capability of letting characters move around as normal on the outside, whereas normal floors don't provide friction or are in another way less useful (details dependent on later implementation of "moving on the outside" of course).

Ladders can only be built on the outside. They don't allow furniture to be built on them. Characters can walk on them and one tile around them just like normal floors.

A button was added to the build UI to build ladders. For now this button is only present in a separate test scene. I'm not sure how to merge this into the main scene without causing major conflicts. The FloorMenu was changed to include the ladder build button.

zerogladderscreenshot

PR #573, opened 2016-08-22
Add build_ladder button to main scene and delete test scene

WardBenjamin:
This moves the button added in #543 to the main scene and deletes the extra test scene added there. Everything is tested and working, at least as far as the floor construction/deconstruction buttons go, so I didn't break anything :P


vogonistic:
Works!

PR #575, opened 2016-08-22
added raw copper

alexisfcote:
Added raw copper with convention from #505.

This is in preparation to make the mining drone have a choice of raw mining #576 .


racerand:
Thanks for noticing @alexisfcote. Will you come vote on the last changes? #505

WardBenjamin:
Looks good! Note that this doesn't add it to worldgen or anything.

PR #514, opened 2016-08-21
[Need feedback/test] Context menu on right mouse click

sboigelot:
Add a context menu to the game.

Right click on any furniture, inventory or character to open a context menu full actions.
This context menu read a list of action defined by the targeted tile content.
There is also support for action only available when the player has a character selected.

Everyone can extend IContextActionProvider.GetContextMenuActions to provide more actions to the differents objects.

image

image

image

PR #576, opened 2016-08-22
Added framework so that the mining drone can mine other things ...

alexisfcote:
than raw Iron.
Waiting for raw copper first  #575 and right click context menu to be able to select #514.
Also waiting for naming convention #505

### DIFF
--- a/Assets/StreamingAssets/Data/Furniture.xml
+++ b/Assets/StreamingAssets/Data/Furniture.xml
@@ -162,7 +162,7 @@
 
     <Furniture objectType="Mining Drone Station">
         <Name>Mining Drone Station</Name>
-        <Description>Creates Steel Plates when worked by a character.</Description>
+        <Description>Creates raw materials when worked by a character.</Description>
         <TypeTag>Workstation</TypeTag>
         <MovementCost>1</MovementCost>
         <Width>3</Width>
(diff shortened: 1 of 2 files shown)


vogonistic:
Updated your title to say `[WIP]` which is short for Work in progress. Remove it when you want us to take a look at it.

alexisfcote:
waiting for #658 to get merged

TomMalbran:
Right now if you set both params to 1, the mining drone would want to mine both items, which would cause issues, since it won't be able to place the second item.

If the idea is to mine one or the other, and use a context menu to select which item to mine, it would be better to have one parameter ```<Param name="Output" value="Raw Iron" />"``` and then you can create the items with ```World.current.inventoryManager.PlaceInventory( j.furniture.GetSpawnSpotTile(), Inventory.__new(j.furniture.GetParameter("Output"), 50, 20) )```

alexisfcote:
I did it like that because param value could only be a float.

TomMalbran:
Right, I totally forgot about that. Once #675 gets merged, we can fix that.

alexisfcote:
It works now, just right the mining drone to change the mined ressource. There should be no conflicts if you change type while it is worked

vogonistic:
Merged!

PR #583, opened 2016-08-22
Update russian localization

Fogsight:
Quick localization update


Assets/StreamingAssets/Localization/ru_RU.lang
```diff
@@ -1,18 +1,30 @@
+﻿menu_construction=Строительство
+menu_work=Работы
+menu_world=Мир
+menu_options=Главное Меню
+menu_settings=Настройки
+menu_furniture=Мебель
```
Grenkin1988: Furniture is not just furniture. Power generator is not furniture :-) So `Мебель` not very good

Fogsight: At this stage everything is a furniture. It's just a placeholder, mostly to tick the box on wiki so that it's not auto-translated.

Mizipzor:
I see no reason to not merge localization changes. If something is wrong, just make a new pull request.

PR #587, opened 2016-08-22
Fixed enumeration modification error in cbJobCallbackComplete

NogginBops:
There was a problem with a lua function registering a callback which modified a currently enumerated list causing a enumeration modification error. This changes the iteration so that they iterate through a copy list.

### DIFF
--- a/Assets/Scripts/Models/Job.cs
+++ b/Assets/Scripts/Models/Job.cs
@@ -182,32 +182,8 @@ public class Job
 
     public void DoWork(float workTime)
     {
-        // Check to make sure we actually have everything we need. 
-        // If not, don't register the work time.
-        if (HasAllMaterial() == false)
-        {
-            ////Logger.LogError("Tried to do work on a job that doesn't have all the material.");
-
-            // Job can't actually be worked, but still call the callbacks
-            // so that animations and whatnot can be updated.
-            if (cbJobWorked != null)
-            {
-                cbJobWorked(this);
-            }
-
-            if (cbJobWorkedLua != null)
-            {
-                foreach (string luaFunction in cbJobCompletedLua.ToList())
-                {
-                    FurnitureActions.CallFunction(luaFunction, this);
-                }
-            }
-
-            return;
-        }
-
-        jobTime -= workTime;
-
+        // We don't know if the Job can actually be worked, but still call the callbacks
+        // so that animations and whatnot can be updated.
         if (cbJobWorked != null)
         {
             cbJobWorked(this);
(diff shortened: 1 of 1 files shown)


Mizipzor:
What was actually fixed here? Arent these lines equivalent? I would like a discussion here before its merged, because if this actually fixes something it might be hiding another bug.

dusho:
List cbJobCompletedLua (that is looped through using foreach) is modified (new item added) by code in LUA script:
`j.RegisterJobCompletedCallback("MetalSmelter_JobComplete")`
So what NogginBops submitted actually fixes the problem.
But ye, the question is whether it shouldn't be safer to enumerate through copy of list (as scripts can add of remove jobs):
`foreach (string luaFunc in cbJobCompletedLua.ToList())
            {
                FurnitureActions.CallFunction(luaFunc, this);
            }`

Mizipzor:
Ah, of course, why didnt I think of that. Yes, continue iterating on a modified collection is not allowed, forcing the developer to make a decision. This is a tricky case, because if:

1. The item is added at the beginning and wont be processed and also offset index by one making the current item being process twice.
2. The item is added to the end and will be processed, which might be what we want.

But, neither of these are apparent. And someone might do something similar but use 1) leading to weird bugs. 

I would be okay with merging this. I just wanna wave the red flag a bit.

quill18:
Would a better fix be to make a copy of the list and iterate over that?  That would ensure that all callbacks present at the beginning of the update would get processed exactly once.

Newly-added callbacks wouldn't get processed until the next frame... which is probably more correct since they probably don't want to be told that already finished is "completed". They're probably looking to be informed about the newly generated job being completed.

Mizipzor:
@quill18 Indeed, the norm as far as Im aware is to make a copy (with linq `.ToList()`) if you need to change to source collection as you iterate. If you **really** need to handle new items added to the collection as it is enumerated, I'd argue that's code smell.

NogginBops:
Ok, then I will change it to that. @Mizipzor @quill18

dusho:
then same should be applied to all the enumerations with LUA calls, right
Job.cs, ln 20; Job.cs, ln 218, ...

NogginBops:
@dusho Will do that!

Mizipzor:
@NogginBops Please add [WIP] or something to the tile so it isnt merged until its complete. We would add a label for this but you cant add or remove labels yourself.

dusho:
actually there is some code duplication there.. ln.194-204 and ln.211-222, so maybe it can be unified there.. then have cbJobWorkedLua.ToList() and cbJobCompletedLua.ToList() only...

quill18:
Looks like there's some mix-ups between cbJobWorkedLua and cbJobCompletedLua in there.

NogginBops:
@quill18 Realized that at the moment after the push

NogginBops:
Now this should be working!

PR #595, opened 2016-08-22
updated DialogBoxSettings.cs to use new Settings.cs

zwrawr:
DialogBoxSettings.cs now uses the new wrapper functions in Settings.cs
to save and load settings with cleaner code


RodRitter:
Tested. Looks good. Merging.

PR #597, opened 2016-08-22
Added the start of the new Inventory.xml

gunthergun:
You can already use it so that you do not have to write what the maxStack is. It is easily added uppon. In the world class there is now an inventoryPrototypes list that functions like the FurniturePrototypes. I will add upon this. Currently nothing uses it

I did not wan't to change everything to the new system yet as that would require to nearly change the whole lua file and merging that would be a pain so, i thought we can start with this and everyone can switch to the new system on their own...


RodRitter:
Great job. Merging.

PR #600, opened 2016-08-22
adds License Headers to Other scripts that were missing them

zwrawr:


Mizipzor:
I just realized we are wrapping the headers in a `#region`. Why? The convention is that the license starts on the very first line. If collapsing is what you want any decent editor can do that on comment blocks anyway. Oh well, not important. It bugs me but I can live with it.

WardBenjamin:
Looks good, thanks!

zwrawr:
@Mizipzor no idea why that is the pattern but that is what all the other headers are like so these should be consistant

Mizipzor:
@zwrawr you are absolutely correct, consistency trumps good.

Issue #538, opened 2016-08-21
Make settings work


zwrawr:
Most of the settings in the settings menu do nothing. 
All of these setting need implementing , some may be easy some might be quite hard to do , some will need to wait till other functionality is implemented *(e.g. a music controller)*
e.g. *as of #535*
```
    public void OnFullScreenToggle()
    {
        /// TODO : impliment full screen toggle
    }

    public void OnQualityChange()
    {
        /// TODO : impliment Quality changes
    }

    public void OnVSyncChange()
    {
        /// TODO : impliment VSync changes
    }

    public void OnResolutionChange()
    {
        /// TODO : impliment Resolution changes
    }

    public void OnAliasingChange()
    {
        /// TODO : impliment AA changes
    }


    public void OnMusicChange()
    {
        /// TODO : impliment VSync changes
    }
```

It would be best to wait until PR #535 is closed.

sboigelot:
This has been implemented, we can close the issue

PR #606, opened 2016-08-22
Implemented OnQualityChange() in DialogBoxSettings.cs

### LINKED ISSUE #565: Quality in Settings is not functioning correctly (reversed)
Basically:
# The items in the Quality Dropdown are reversed:

- When **choosing Low** in the Quality setting, it actually **sets to high** quality.
- When **choosing High** in the Quality setting, it actually **sets to low** quality.

## Why is this happening?

In [DialogBoxSettings.cs](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs#L111):
```
public void OnQualityChange()
    {
        gameSettings.qualityIndex = QualitySettings.masterTextureLimit = qualityDropdown.value;
    }
```
[QualitySettings.masterTextureLimit](https://docs.unity3d.com/ScriptReference/QualitySettings-masterTextureLimit.html) is a texture size limit applied to all textures:

- 0 (the default value) = no limit, therefore it sets the highest quality
- Higher value = higher limit, therefore lower quality.

The problem is **qualityDropdown.value** being set like that:

- **Item 0: Low (when 0 is supposed to be high)**
- Item 1: Med
- **Item 2: High (when 2 is supposed to be low)**

screenshot_1

## How to fix it?

I am sure that by now you got it and know how to fix it...
_Hint: change the order of the items_

The only thing is... **You need to change the** _SCENE_ **file to fix it**, so @quill18 or one of the managers should do it instead of me messing up everything.

Thats it!

_(Please ignore the fact that this issue is too detailed, but thats how I like it...)_

itay347:
As said in the title, **I have edited the quality setting in order for it to work**.

Also, this fixes #565 which I have already closed.
_(It took me some time to understand that I could fix #565 without editing the SCENE file)_

### DIFF
--- a/Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
+++ b/Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
@@ -112,7 +112,12 @@ public class DialogBoxSettings : DialogBox
 
     public void OnQualityChange()
     {
-        /// TODO : impliment Quality changes
+        // masterTextureLimit should get 0 for High quality and higher values for lower qualities.
+        // For example count is 3 (0:Low, 1:Med, 2:High)
+        // for High: count - 1 - value  =  3 - 1 - 2  =  0  (therefore no limit = high quality)
+        // for Med:  count - 1 - value  =  3 - 1 - 1  =  1  (therefore a slight limit = medium quality)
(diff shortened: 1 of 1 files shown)


zwrawr:
Works , changes Quality Settings.
Completes a part of #538

zwrawr:
noticed an issue with the loading of Settings, the load function only updates the UI and not the actual settings.
That's an issue out side of this PR though.

PR #610, opened 2016-08-22
[Small] Added pivot points for sprites and fixed the goddamn character pivot.

NogginBops:
Before the (0.5, 0.5) pivot on sprites was hard coded, now it's changeable though the xml. Note: The pivot tags are optional. I also fixed the character standing slightly below the actual tile.


WardBenjamin:
Looks good! Merging!

PR #612, opened 2016-08-22
Dev tool for spawning inventory

abackwood:
As we add more and more inventory types and requirements for building things, it would be useful to bypass production chains and drop inventory into the world on the fly (for dev purposes, obviously)

This is a UI window that lists all inventory prefabs. When you click on one in the list, you go into a special sort of build mode where 1 of that inventory will be added to a tile you click on. Right-clicking disables it again.

A setting was added to Settings.xml to switch Dev Tools on or off, which relevant controllers look at to determine if the spawner should be created and interacted with

The UI is entirely code-built, no changes to the scene. It's fugly as all get-out, so I'd be happy for someone with higher UI-making stats to revamp that if they like.

invspawnerscreenshot


mikejbrown:
:+1: for dev tools. May as well have a flag in Settings.xml to turn it on or off so players (& modders!) can access it if they really want to. It can just be off by default in the release version.

abackwood:
I'm changing the MouseController. The order of methods seems to be a mess there. Any requests for changing other things in the file while I'm at it?

abackwood:
Removed the WIP status

TomMalbran:
If you are taking suggestions for the MouseController, I think that the SelectionInfo class, should be a standalone class (probably in the UI folder), with a proper constructor and methods. Then the MouseController would call those methods instead of doing all the stuff. But maybe that is better in another PR.

abackwood:
I'm planning on doing a PR cleaning up MouseController a bit, and UpdateDragging() especially, after this is good and ready
... which I do believe it is now :)

Mizipzor:
Tested. I could spawn items on the ground. The GUI is really crude, but its a start. :)

PR #616, opened 2016-08-22
Hotfix for 608... Last one, i promise

QuiZr:
*And i will test all the thins later, for now I blame lack of unit testing*


WardBenjamin:
10/10 testing. 

Merging!

PR #622, opened 2016-08-22
Added background music

Ferr013:
Since I haven't enough programming skills, I tried contributing with some (not really high quality) music.
Thanks for the people that help me figure out how to implement it :smile: :+1:


RodRitter:
This is a good opportunity for someone to implement a sound manager. If you're not that good on the programming end then maybe you can just add your music by itself, without the scripts? Then the next person is able to implement it into the the sound manager feature.

Ferr013:
I think I've fixed the conflict. Thank you :+1:

mikejbrown:
How easy would it be to hook up this MusicController to OnMusicChange() in DialogBoxSettings.cs so that the music volume setting works? This would partially address #538.

RodRitter:
After looking through it again, there are far too many scene additions. Please could you try to move any scene additions into script?

PR #623, opened 2016-08-22
Added Path drag type - Preparing for conduits & piping

RodRitter:
I won't merge this in myself. I'll wait for another mod to look it over first. @Mizipzor @WardBenjamin @vogonistic What is the process for moderator contributions?

I added a new drag type for furniture: `<DragType>path</DragType>`

It's a very rudimentary system for laying down pipe-like furniture. At the moment it does have an x-bias but that can be improved over time with some better logic. 

image

### DIFF
--- a/Assets/Scripts/Controllers/MouseController.cs
+++ b/Assets/Scripts/Controllers/MouseController.cs
@@ -276,6 +276,12 @@ public class MouseController
         int start_y = Mathf.FloorToInt(dragStartPosition.y + 0.5f);
         int end_y = Mathf.FloorToInt(currPlacingPosition.y + 0.5f);
 
+        // Keep an un-flipped copy of start/end points
+        int raw_start_x = start_x;
+        int raw_end_x = end_x;
+        int raw_start_y = start_y;
+        int raw_end_y = end_y;
+
         // We may be dragging in the "wrong" direction, so flip things if needed.
         if (end_x < start_x)
         {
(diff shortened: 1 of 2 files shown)


Sommerbo:
I think an extra set of eyeballs are a good think, so no merging of your own PR is a good policy.

WardBenjamin:
@RodRitter First of all, you should join us in Discord! There's a link in the README

My policy for gatekeeper contributions has been to have another gatekeeper review it, usually Vogonistic since the time he's online usually overlaps with mine.

Anyway, looks good!

RodRitter:
I suppose you can change the parameter of the wall to test it. Nothing so far, it seems.

Mizipzor:
Changing walls to be built this way makes sense and makes for a good test.

RodRitter:
I agree. Changed walls to use path drag type.

vogonistic:
Awesome. Works great! Now, lets get some pipes and electrical conduits

Assets/Scripts/Controllers/MouseController.cs
```diff
@@ -369,6 +396,26 @@ void UpdateDragging()
                             {
                                 isValid = true;
                             } 
+                        } else if (dragType == "path")
```
TomMalbran: The "else if (...)" should go on a new line.

Maybe we should use an enum for the drag types.

RodRitter: I'll keep that in mind when I go back to refactor. Thanks

vogonistic: Maybe move them to a dictionary of functions so that Lua mods can add new amazing ways of laying down furniture?

TomMalbran: Oops, it got merged while I was checking the code.

That could be an option, although I am not sure how many more ways to drag there is in grid system.

I would suggest to remove the restriction of only allowing drag for 1x1 furniture, and make a "none" dragType which would be the default for all the furniture. In the future if some would want to make a 2x2 big pipe furniture, then it would work with dragging.

PR #627, opened 2016-08-22
Make landing pad only work ouside

sboigelot:
Self explaining

Landing pad build indoor will not generate job


Sommerbo:
Why use ~= (bitwise not) instead of a regular not okr even >1 for line 414
` if(furniture.tile.room.ID ~= 0) then`

TomMalbran:
Even better is to use ```Room.IsOutsideRoom()```.

bjubes:
Should they be able to be even built indoors?

TomMalbran:
They probably shouldn't. But at the same time you could build walls around it and make it be inside. So I would suggest to both prevent it from being built outside and working inside.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -410,6 +410,11 @@ function PowerGenerator_JobComplete( j )
 end
 
 function LandingPad_Temp_UpdateAction(furniture, deltaTime)
+    
+    if(furniture.tile.room.ID ~= 0) then
+        return
+    end
+    
 	spawnSpot = furniture.GetSpawnSpotTile()
```
Mizipzor: For some reason it appears that this file uses tabs. Are we going for tabs in lua files?

Mizipzor:
Added [Conflict], remove when conflicts are fixed. :)

Assets/StreamingAssets/LUA/Furniture.lua
sboigelot: There is a lot of tabs and space in the file, I'm not sure what we should use as a default. Is there a decision about it?

sboigelot:
Conflict resolve & check changed to 
if(not furniture.tile.room.IsOutsideRoom()) then 
as @TomMalbran suggested

I don't want to forbid the placement indoors for now as we can build it indoors then remove walls etc. But that could be added. Some rooms could be indoors with doors closed and outdoors with doors open, like a hangar, a ship docking bay etc..

Mizipzor:
So if understand this correctly, I can still place the landing pad inside, it wont just work? Sounds confusing to me. Surely its useful to indicate an invalid position by not allowing to build? If I need to change the room to have a pad there, I change the room first.

RodRitter:
I agree with @Mizipzor - Surely the landing pad is an extension to your starbase?

sboigelot:
@Mizipzor @RodRitter Ok I added code to prevent the landing pad to be constructed indoors.

svmnotn:
alt text

vogonistic:
Merged!

Issue #577, opened 2016-08-22
[Bug] Metal smelter job 'weirdness'


Randshot:
This bug was introduced in PR #567.
I can't really describe it, but the smelter job does not seem to get removed or something similar to this.
Also, somehow, the registration for the jobWorked callback went missing, don't ask me how.

I am filing this issue mostly for myself, so that I remember what I was working on.

gunthergun:
Shouldn't this be fixed now as the fix has been merged --> so please close this issue

Randshot:
Oh, sorry. Thought that a mod would have closed it.

PR #635, opened 2016-08-22
Fixing metal smelter bug issue #577 (with hacks)

mikejbrown:
Hack to fix issue https://github.com/TeamPorcupine/ProjectPorcupine/issues/577 (correct me if this issue is about a slightly different bug! This fixes the only smelter weirdness I can reproduce.)
The problem is that the spawn spot inventory is disappearing so the smelting never gets started. I've just turned the smelter into a little finite state machine using a furniture param "smelting" as flag and manually resetting the spawn spot inventory. I'm sure there's a better way to fix this but it's working for now.


vogonistic:
Merged!

PR #641, opened 2016-08-22
UberLogger

Mizipzor:
This commits actual files rather than having UberLogger as a submodule.


vogonistic:
Merged! This is better than submodules.

PR #644, opened 2016-08-22
door rotation properly based on xml properties rather than hard coded

mikejbrown:
Very small change... just allows future extensibility (easier implementation of new door types)

### DIFF
--- a/Assets/Scripts/Controllers/FurnitureSpriteController.cs
+++ b/Assets/Scripts/Controllers/FurnitureSpriteController.cs
@@ -59,7 +59,7 @@ public class FurnitureSpriteController
         furn_go.transform.SetParent(furnnitureParent.transform, true);
 
         // FIXME: This hardcoding is not ideal!
-        if (furn.objectType == "Door" || furn.objectType == "Airlock")
+        if (furn.HasTypeTag("Door"))
         {
             // By default, the door graphic is meant for walls to the east & west
             // Check to see if we actually have a wall north/south, and if so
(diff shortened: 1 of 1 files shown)


TomMalbran:
We should probably also change lines 159  and 164 so that it checks for ```furniture.HasTypeTag("Wall")``` instead of using the objectType.

mikejbrown:
@TomMalbran good catch.

I've cleaned up a few other references to objectType now. There is still ugly hard coding in GetSpriteForFurniture() though. Any thoughts on the best way to deal with this?

TomMalbran:
@mikejbrown Nice change.


For ```GetSpriteForFurniture()``` I think that the best solution is to make it part of the Furniture prototype, and make it possible to be changed as a LUA function. So:

1. The furniture would have a ```GetSpriteName()``` function, and a string property ```getSpriteNameAction``` (or something like that) with the name of a LUA function. In ```GetSpriteName()``` it will ask if ```getSpriteNameAction``` is set and if it is, it will call it and return the result, and if not it can return the ```objectType```.

2. Then we can set the ```GetSpriteName``` as a property on the xml and have a ```GetSpriteName_Door```LUA function that can return the sprite name depending on the furniture parameters.

Maybe this would be for a separate PR.

mikejbrown:
@TomMalbran Good ideas but yes, logically that would be for another PR I think.

Mizipzor:
@TomMalbran Did you test this? If you vouch for it Ill merge.

TomMalbran:
@Mizipzor No, I haven't tested it. But the code looks good.

Mizipzor:
Tested by building a door and two walls, the door rotated accordingly.

PR #646, opened 2016-08-22
Fixing metal smelter bug issue #577 (with hacks)

mikejbrown:
Reverts TeamPorcupine/ProjectPorcupine#635 -- just noticed I used ChangeParameter when it should be SetParameter -- very sorry


mikejbrown:
There... that should fix it. Wheew.

vogonistic:
Merged

PR #649, opened 2016-08-22
[WIP] Mouse Cursor Information Display

Tritaris:
messed up on #584, I merged instead of rebased I think.

Adds 4 text fields that follows the mouse cursor around to display different information. Currently what is displayed is hardcoded in C# [Test Video](https://youtu.be/TNBDcQpiPUI) . 

Right now this is kinda coded as module that only needs or adds a couple lines to the existing code to get it started. Really this should be added to MouseController as its controlling the mouse pointer and the 4 text fields around it. Working on a proof of concept first then i'll move it over.

So I was thinking about adding Lua support to control what is displayed around the mouse so it could be mod-able. As I think about it that could lead to some weird things or stuff we don't want the player to see. What do you think about using a menu with a text input and writing keywords that will call code to display what the person wants?

Example "@time" would display the current game time, "@mousePOS" shows the mouse position in world coordinates.

If you have any ideas feel free to comment.


sboigelot:
Maybe you can add an option to enable/disable the textbox, so the player only see them when he wish (like pressing shift for instance, or a toggle on the ui)

Otherwise I fear this may add a lot of clutter in the view. But don't get me wrong, it's great to have more info!

Tritaris:
https://github.com/Tritaris/ProjectPorcupine/tree/MouseCursor is my current branch for this.

PR #651, opened 2016-08-22
Moddable tiles

NogginBops:
This adds xml and lua definable tiles with support for mods. Changes the scene to make the tile build menu generated on the fly.

### DIFF
--- a/Assets/Scripts/Controllers/BuildModeController.cs
+++ b/Assets/Scripts/Controllers/BuildModeController.cs
@@ -153,15 +153,9 @@ public class BuildModeController
                 // This tile position is valid tile type
 
                 // Create a job for it to be build
-                Job j = new Job(
-                    t,
-                    tileType, 
-                    Tile.ChangeTileTypeJobComplete, 
-                    0.1f, 
-                    null,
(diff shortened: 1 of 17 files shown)


vogonistic:
`[WIP]` might be a better tag in the title, but this works too.

Noticed things:
  1. Not sure where you are on master, but the characters will happily float in the air to build the tiles in the wrong order.
  1. It doesn't prevent me from rebuilding floor. So putting floor on ladders and vice versa.
  1. What is the change to the scene? Some things are moved and disabled?

This will be really awesome! Good job with this!

abackwood:
Lots of TODO errors popping up when I run it. Is this WIP?

NogginBops:
The log messages are to remind me that there is a possibility to add these features. I don't know if they should actually be in. So yeah this is probably WIP

Mizipzor:
Ok. I started the game and everything still looks the same. Scene changes are scary. Not only because of conflicts, but if you based your work on an old scene we can accidentally remove recently added stuff by merging. But, the diff looks really clean. And I welcome anything that makes stuff generated on the fly rather than having to be in the scene itself.

PR #652, opened 2016-08-22
You can now close a menu by clicking again on the button in the botto…

jdlaroche:
You can now close a menu by clicking again on the button in the bottom toolbar. Prior to this fix, you had to click on another button to close the construction menu.


bjubes:
you can use the toggleMenu fuction which does the same thing. This was orignally in my pr #400 but someone must have changed this code to not use it

jdlaroche:
The toggle menu function is for the submenu (or maybe it work for the main menu?) The submenu toggling is still working... I fixed the main menu.

jdlaroche:
I looked again and toggle menu wouldn't help if we want to call DeactivateAll... It would always deactivate the menu first and reactive it so we would not be able to close it. The IF I added would still be required if we want to call DeactivateAll (which is what we were doing before this fix)

Assets/Scripts/UI/InGameUI/MenuController.cs
```diff
@@ -95,8 +95,13 @@ void Update()
 
     public void OnButtonConstruction()
     {
-        DeactivateAll();
-        constructorMenu.SetActive(true);
+        if (constructorMenu.activeSelf)
+        {
+            DeactivateAll();
+        } else {
```
Mizipzor: But that second brace on a new line as you do in the other places. :)

jdlaroche: Done :)

Assets/Scripts/UI/InGameUI/MenuController.cs
```diff
@@ -113,13 +120,27 @@ public void OnButtonWorld()
...
+        }
+        else
+        {
+            DeactivateAll();
+            settingsMenu.SetActive(true);
+        }
     }
-}
\ No newline at end of file
+}
```
Mizipzor: This I do appreciate. :)

Mizipzor:
Tested. It closes the menu all right and I cant find something that has broken. :)

PR #653, opened 2016-08-22
[UI] Build menu closes after selecting furniture to build

frankitox16:
Pretty straightforward... (that commit name was unintended, got a couple of troubles getting used to git bash ^^)


svmnotn:
alt text

Not sure if this is the exact behavior that is desired however.

frankitox16:
@svmnotn what do you mean? any weird things happening? that line of code should hide te Build [furniture_name] buttons, that's all...

svmnotn:
No, there is nothing wrong with it, I'm just not sure if that is the behavior that is desired. But, it is not up to me to decide that.

vogonistic:
I like it. We can revert it if people dislike it.

PR #655, opened 2016-08-23
2 little fixes for WorldGen 

relinu:
Overlooked 2 little bugs while testing x.x'


Mizipzor:
Tested, world can still be generated. Thats good enough for me.

PR #662, opened 2016-08-23
Cleaning up MouseController

abackwood:
MouseController.cs has some very large functions with lots of nested for- and if-statements. I'm having a look if I can make that a little more organised and clean


abackwood:
No longer WIP

Since this is purely a refactor, the primary testing is:
- Does everything still work the same?

RodRitter:
Tested. Everything looks good to me. Merged.

PR #666, opened 2016-08-23
Move Prefab to Resources for dynamic loading

Fogsight:
To be able to load prefabs dynamically from code.

To use prefab button, for example, scene would have to be edited every time otherwise.


vogonistic:
@Fogsight I don't know Unity enough to evaluate this for want or functionality. How does this work with the other UI PR's we have?

Is this something that has been discussed or just a general improvement? What does it affect?

@WardBenjamin @RodRitter @Mizipzor Got input?

Fogsight:
I think meta will have all needed info about it. It should be painless. All other stuff - textures, etc. should be in Streaming Assets or Resources to be reachable through code.

vogonistic:
It doesn't merge cleanly, even though GitHub seems to think so. `ContextMenuItemPrefab.prefab` and `ContextMenuPrefab.prefab` doesn't seem to exist anymore. Instead there is a `ContextMenuItem.prefab`

Can you fix that and I'll take a look again? I'm adding `[conflict]` to the title of this PR, just remove it when you have taken care of it.

Fogsight:
Is it normal to have such a mess after Rebase?

vogonistic:
No, it seems excessive. Can you rebase on top of the upstream master, squish your commits and force push?

Fogsight:
It's all up-to-date won't let me rebase again. I'll kill this thing with fire and create a new one.

PR #672, opened 2016-08-23
ModUtils: Exposes static Unity C# functions to Lua

bjubes:
This is a replica of the previous modutils #37, without accidentally changing hundreds of files. I don't know if this is needed in the project, but the full discussion never got finished after I messed up that PR.

ModUtils exposes methods in c# by encapsulating them in methods within the ModUtils class. These can then be called in Lua with `ModUtils.MethodName()` and removes the need for a redundant "utilities" section in the Lua files, which was essentially for rewrites of existing c# functions, like `Mathf`.

Bonus: all changes are Style Cop compliant

### DIFF
--- a/Assets/Scripts/Models/FurnitureActions.cs
+++ b/Assets/Scripts/Models/FurnitureActions.cs
@@ -35,7 +35,8 @@ public class FurnitureActions
         myLuaScript.Globals["Inventory"] = typeof(Inventory);
         myLuaScript.Globals["Job"] = typeof(Job);
 
-		myLuaScript.Globals ["ModUtils"] = typeof(ModUtils);
+        myLuaScript.Globals["ModUtils"] = typeof(ModUtils);
+
         // Also to access statics/globals
         myLuaScript.Globals["World"] = typeof(World);
     }
--- a/Assets/Scripts/Models/FurnitureActions.cs
+++ b/Assets/Scripts/Models/FurnitureActions.cs
@@ -35,6 +35,7 @@ public class FurnitureActions
         myLuaScript.Globals["Inventory"] = typeof(Inventory);
         myLuaScript.Globals["Job"] = typeof(Job);
 
+		myLuaScript.Globals ["ModUtils"] = typeof(ModUtils);
         // Also to access statics/globals
         myLuaScript.Globals["World"] = typeof(World);
     }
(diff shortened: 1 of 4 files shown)


Mizipzor:
Whats the reason for naming the method `Clamp01`?

bjubes:
@Mizipzor  sorry not sure what you mean? it was called Clamp01 in the original Lua and i used the Clamp01 function in unity and made a wrapper called Clam01.

Mizipzor:
I was wondering why it wasnt simply called `Clamp`.

mmooxy:
@Mizipzor because `Clamp` has parameters for min and max. [https://docs.unity3d.com/ScriptReference/Mathf.Clamp.html](https://docs.unity3d.com/ScriptReference/Mathf.Clamp.html)

bjubes:
When someone needs to use `Clamp(min,max)` they can add it to modUtils and then call `ModUtils.Clamp(val1,val2)`. Thats basically the point of ModUtils, to be expanded on when needed.

Mizipzor:
Is this some lua convention im not aware off? In the script reference linked, the method is called `Clamp` not `Clamp01`. What is it that im not getting? Why the numbers?

mmooxy:
@Mizipzor Sorry, I guess I should have linkes both Clamps.
`Clamp(float value, float min, float max)` <- not used right now
[https://docs.unity3d.com/ScriptReference/Mathf.Clamp.html](https://docs.unity3d.com/ScriptReference/Mathf.Clamp.html)
`Clamp01(float value)` <- min = 0, max = 1
[https://docs.unity3d.com/ScriptReference/Mathf.Clamp01.html](https://docs.unity3d.com/ScriptReference/Mathf.Clamp01.html)

bjubes:
if u want i can add the normal clamp variation, but it will be unused for now in the Lua

Mizipzor:
Ok now I see. Carry on then. ^^

bjubes:
so should i add the clamp fuciton too?
how about log, logwraning, logerror since these arent accessible via lua either, right?

mmooxy:
I guess you could also add Sqrt, Random.Range (int & float) and Random.value. But I'm no expert on lua, maybe they are already implemented.

bjubes:
i looked it up and all those exist in lua already. I'm going to add debug.log and such since there is currently no way to do that

vogonistic:
Excellent change!

Issue #302, opened 2016-08-19
Discussion: Gas system revamp.


WardBenjamin:
"A Grand System of Gasses, Leaks and Pumping"

How should we proceed from the room-based solution that we have now, with gases leaking through doors and open airlocks? I've seen several mentions of piping and ventilation systems, CO2 scrubbers, and other interesting things. 

(Also see #284)

Discuss!

kd7uiy:
I was just going to open something about this!

First of all, I think we need a way to operate in a vacuum first, a space suit. Otherwise, it's just not going to be very easy to do anything.

Secondly, gas should be able to move between rooms with open doors, and escape if not properly prepared. An airlock should minimize the gas exchange/escape.

Third, the O2 should slowly be used in a room. If there is no O2, a person should likely die. A space suit has a limited amount of air for working outside.

Sommerbo:
Gas amount(tile count) should be taken into account when filling a room.

koosemose:
I think in general room-based is the best approach, as anything smaller scale would likely be too costly for not enough a benefit. @Sommerbo is correct, we need to ensure the roomsize is taken into account when filling it. Ill effects from improper gases in a room shouldn't be implemented until protection from gases (or lack thereof) is available.

A little tightening up is needed on gas leaking, I noticed gas will never quite fully disappear when leaking outside (it just drops to increasingly tiny numbers, I think it got down around 10^-19 before I stopped it, at some point it just needs to vent completely. I haven't gotten around to checking if a similar issue with leaking vanishingly small numbers happens when leaking to another room.

Once that's tightened up, it should be easy to implement a furniture that can force gas to flow at a certain rate and possibly force it to go in a certain direction (possibly 2 separate furnitures) along with being able to be closed off (will possibly need the ability for characters to interact with furnitures beyond the current building focused system).

At some point a decision needs to be made in how complex a breathable air simulation is wanted. Is it simply a question of there being enough oxygen to breathe, should CO2 be generated and need scrubbed/vented, should you also need a proper mix of Nitrogen? I personally think the first or maybe the second would be preferable, the third seems a higher degree of scientific simulation than I imagine the average base builder player would want.

WardBenjamin:
I personally am all for the O2/CO2 system, but it's not like it couldn't be changed in the future if we want to.

koosemose:
[...]

PR #670, opened 2016-08-23
Fixing stockpile logic: implementing taking multiple inventory types and any/all logic

mikejbrown:
Fixes #568 
This is still incomplete and could probably use a lot of refactoring. I'm putting it up to get feedback.

This PR implements any/all logic for stockpile type jobs. Jobs now have a `public bool acceptsAny` & material requirements can be checked in a unified way through the method `MaterialNeedsMet()` which calls `HasAllMaterial()` or `HasAnyMaterial()` as needed.

Character logic has been updated so that when working on an "any" job they will go pick up any reachable inventory listed in the Job's inventoryRequirements.

Furniture now has an `public Inventory[] AcceptsForStorage()` function which is hard coded for now for testing, but which will eventually be hooked up to stockpile UIs. The stockpile LUA has been updated to use this.

I'm thinking about implementing `AcceptsForStorage()` by using the furniture params. The easiest way at the moment would be to store say a JSON list of objectTypes or similar. This way the stockpile settings can be saved easily as well. If PR #552 gets merged this will have to change.

Thoughts?

PR #675, opened 2016-08-23
Multi-type and nestable parameters

koosemose:
As discussed in #532 and #302. While there is still room for refinement, particularly as things change around it (particularly any changes to how Xml is handled), this allows more complex custom parameters, which will aid in modding and limiting the number of variables need to be added to the furniture class (and any other class which is changed to make use of this).

While I don't like including multiple changes in one PR, the changes to furniture.xml will cause errors if applied without the changes to the furniture class and the new Parameter system will cause errors in Furniture.lua, if it isn't updated.

Summary: Allows Params to be nested and directly store strings (which can be read as Floats when needed), such as this example from furniture.xml:

```
        <Params>
            <Param name="gas_limit" value="0.2" />
            <Param name="gas_per_second" value="0.16" />
            <Param name="gas_gen">
                <Param name="O2">
                    <Param name="gas_limit" value="0.2" />
                </Param>
                <Param name="N2">
                    <Param name="gas_limit" value="0.8" />
                </Param>
            </Param>
        </Params>
```

A Get method has been implemented for furnParameters, the contents are accessed as follows (using O2 generator's structure as an example.):
`if(Parameters.HasContents())`
Singular Parameters
`Parameters["gas_per_second"].ToFloat();`
`Parameters["gas_per_second"].ToString();`
`Parameters["gas_gen"]["O2"]["gas_limit"].ToFloat();`
`Parameters["gas_per_second"].ChangeFloatValue(.2);`
`Parameters["gas_per_second"].SetValue("1");`
`Parameters["gas_per_second"].SetValue(1);`
Group Parameters
`Parameters["gas_gen"].Keys();`
`if (Parameters["gas_gen"].ContainsKey("O2"));`
`Parameters.AddParameter(new Parameter("max_pressure","1"));`
`Parameters.AddParameter(new Parameter("max_pressure",1));`
Each "layer" is a Parameter, and the tree can be descended with iterators (Parameter["key"]), and final value accessed or changed with any of the singular parameter functions. Has Contents can be used if you are uncertain of the tree's structure, to determine if it is singular or group, and Keys() can assist in traversing to the next level down.


FilippoLeon:
If you implement parameter as a "property", I think it coul be possible to write:
`furn.parameters["openness"].ToFloat()`
istead of
`furn.GetParameters()["openness"].ToFloat()`

the GetParameters()[], from a functional standpoint, is working, but is screwing up my ocd.

I think there is something wonky with tabs in the switch statement.

koosemose:
My understanding is that there is something weird with accessing Properties from MoonSharp, and that's one of the primary places to access them from, so I had to go with the method version.

FilippoLeon:
I know many do not like it, but "public field"?

mikejbrown:
:+1: I like this, though note that it will likely break any outstanding PRs that touch furniture.

koosemose:
Apparently I was wrong, I could swear I remember in one of the tutorial videos when he was doing the Lua, quill saying there was something wrong with using Properties in Lua. I think for now I'll have both in, just in case it was some other oddity with it I can't remember.

koosemose:
@mikejbrown yeah, I tried to get this done while there were little to no PRs touching anything furniture related, but it's taken so long I think a few managed to sneak in. I'm hoping my little guide to use will ease anyone having to resolve conflicts while using this system, if mine is merged first... otherwise I'm going to have to play catchup again.

mikejbrown:
@koosemose I don't think that mine ( #670 ) conflicts with this _yet_. It would do if I had already started implementing the stockpile accepted inventory stuff as furniture params as I was planning. So I'll hold off on that till this gets merged. (It'll probably be much cleaner with your system anyway.)

sboigelot:
It's ok to conflict with PR about furniture for now I think. We can propose help to other dev too if they need more info about how to fix the lua.

Great work by the way :D

Assets/Scripts/Controllers/FurnitureSpriteController.cs
```diff
@@ -205,17 +205,17 @@ public Sprite GetSpriteForFurniture(Furniture furn)
             // FIXME: All this hardcoding needs to be generalized later.
             if (furn.objectType == "Door")
             {
-                if (furn.GetParameter("openness") < 0.1f)
+                if (furn.GetParameters()["openness"].ToFloat() < 0.1f)
```
TomMalbran: You should use the property here. And maybe remove the ```GetParameters``` function.

Assets/Scripts/Models/Parameter.cs
```diff
@@ -0,0 +1,202 @@
+﻿using UnityEngine;
```
TomMalbran: Missing the license header

vogonistic:
Merged!

PR #676, opened 2016-08-23
Deconstruct Button fix

Dan54:
Fixes Deconstruct button


abackwood:
Tested. Furniture can be deconstructed again.
Backwood seal of approval

RodRitter:
Tested as well and it looks good. Is it necessary to modify the deconstruct button scene object?

dustinwloring1988:
You should not have to change the unity scene just if you got rid of the blank refreance on the button. Note that the PR i have posted has this fix in it.

Dan54:
@dustinwloring1988 Your fix is better - closing.

PR #677, opened 2016-08-23
Changed the resource spawn chance to the weight based system

### LINKED ISSUE #671: [Discussion] World generation resources
Currently, Resources have a chance between 0 and 1 AND a combined chance of 1. Every time someone adds a new resource (either during development or with a mod) they would have to change the chances of the other ressources.
```c#
if (Random.value <= asteroidResourceChance && t.Furniture == null)
{
    if (resources.Length > 0)
    {
        int currentchance = 0;
        int randomchance = Random.Range(0, 100);
        for (int i = 0; i < resources.Length; i++)
        {
            Inventory inv = resources[i];
            int chance = inv.stackSize; // In stacksize the chance was cached
            currentchance += chance;
            if (randomchance <= currentchance)
            {
                int stackSize = Random.Range(resourceMin[i], resourceMax[i]);
                if (stackSize > inv.maxStackSize)
                {
                    stackSize = inv.maxStackSize;
                }
                world.inventoryManager.PlaceInventory(t, new Inventory(inv.objectType, inv.maxStackSize, stackSize));
                break;
            }
        }
    }
}
```
```xml
<Resources>
    <Resource objectType="Raw Iron" maxStack="50" min="5" max="15" chance="0.5" />
    <Resource objectType="Ice" maxStack="10" min="5" max="10" chance="0.15" />
    <Resource objectType="Uranium" maxStack="20" min="5" max="15" chance="0.35" />
</Resources>
```
I sugesst 2 alternatives. 
1. This still has the problem of eventually being over 100%, but I doubt there will be that many resources.
```c#
asteroidResourceChance = 1; // normal rate

if (t.Furniture == null && resources.Length > 0)
{
    int currentchance = 0;
    float randomchance = Random.Range(0, 100f);
    for (int i = 0; i < resources.Length; i++)
    {
        Inventory inv = resources[i];
        int chance = inv.stackSize; // In stacksize the chance was cached
        currentchance += chance;
        if (randomchance <= currentchance * asteroidResourceChance)
        {
            int stackSize = Random.Range(resourceMin[i], resourceMax[i]);
            if (stackSize > inv.maxStackSize)
            {
                stackSize = inv.maxStackSize;
            }
            world.inventoryManager.PlaceInventory(t, new Inventory(inv.objectType, inv.maxStackSize, stackSize));
            break;
        }
    }
}
```
```xml
<Resources> <!-- simplyfied -->
    <Resource chance="0.05" />
    <Resource chance="0.01" />
    <Resource chance="0.03" />
</Resources>
```
2. Have a weight for every resource to determine how ofter they are supposed to spawn (higher weight => more likely) this will however change the likelihood of a resource to spawn depending on how many resources exist.
```c#
if (Random.value <= asteroidResourceChance && t.Furniture == null)
{
    if (resources.Length > 0)
    {
        int currentweight = 0;
        float randomweight = Random.Range(0, sumOfAllWeights);
        for (int i = 0; i < resources.Length; i++)
        {
            Inventory inv = resources[i];
            int weight= inv.stackSize; // cached weight
            currentweight += weight;
            if (randomweight <= currentweight)
            {
                int stackSize = Random.Range(resourceMin[i], resourceMax[i]);
                if (stackSize > inv.maxStackSize)
                {
                    stackSize = inv.maxStackSize;
                }
                world.inventoryManager.PlaceInventory(t, new Inventory(inv.objectType, inv.maxStackSize, stackSize));
                break;
            }
        }
    }
}
```
```xml
<Resources> <!-- simplyfied -->
    <Resource weight="5" />
    <Resource weight="1" />
    <Resource weight="3" />
</Resources>
```

mmooxy:
Everyone seemed fine with the weight based system.

closes #671

### DIFF
--- a/Assets/Scripts/Models/WorldGenerator.cs
+++ b/Assets/Scripts/Models/WorldGenerator.cs
@@ -40,10 +40,10 @@ public class WorldGenerator
         int xOffset = Random.Range(0, 10000);
         int yOffset = Random.Range(0, 10000);
 
-        int sumOfAllSpawnWeights = 0;
+        int sumOfAllWeightedChances = 0;
         foreach(Inventory resource in resources)
         {
-            sumOfAllSpawnWeights += resource.stackSize;
+            sumOfAllWeightedChances += resource.stackSize;
         }
 
         for (int x = 0; x < startAreaWidth; x++)
@@ -89,7 +89,7 @@ public class WorldGenerator
                         if (resources.Length > 0)
                         {
                             int currentweight = 0;
-                            int randomweight = Random.Range(0, sumOfAllSpawnWeights);
+                            int randomweight = Random.Range(0, sumOfAllWeightedChances);
 
                             for (int i = 0; i < resources.Length; i++)
                             {
@@ -178,7 +178,7 @@ public class WorldGenerator
                                         res.Add(new Inventory(
                                                 res_reader.GetAttribute("objectType"),
                                                 int.Parse(res_reader.GetAttribute("maxStack")),
-                                                Mathf.CeilToInt(float.Parse(res_reader.GetAttribute("spawnWeight")))));
+                                                Mathf.CeilToInt(float.Parse(res_reader.GetAttribute("weightedChance")))));
 
                                         resMin.Add(int.Parse(res_reader.GetAttribute("min")));
                                         resMax.Add(int.Parse(res_reader.GetAttribute("max")));
--- a/Assets/Scripts/Models/WorldGenerator.cs
+++ b/Assets/Scripts/Models/WorldGenerator.cs
@@ -40,6 +40,12 @@ public class WorldGenerator
         int xOffset = Random.Range(0, 10000);
         int yOffset = Random.Range(0, 10000);
 
+        int sumOfAllWeights = 0;
+        foreach(Inventory resource in resources)
+        {
+            sumOfAllWeights += resource.stackSize;
+        }
+
         for (int x = 0; x < startAreaWidth; x++)
         {
             for (int y = 0; y < startAreaHeight; y++)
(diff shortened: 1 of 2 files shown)


RodRitter:
I think using "weight" as a naming convention for this system is misleading as we might introduce physical item weight in the future. I can't merge this as it currently stands. Will need some more feedback from the community.

koosemose:
The problem is that weight is the exact correct term for this, anything else is likely to be highly inaccurate and obfuscate what it is.

GamerInspired:
What about using spawnweight? Or something. I would think later on we would use the name itemweight for the weight of an item?

koosemose:
I actually thought spawnWeight would do pretty good. What else would it be used for?

GamerInspired:
After a bit more thinking about it I stand by what I originally said use spawnWeight.

RodRitter:
I think we'd rather have inaccurate parameter names than overall confusion.

mmooxy:
@RodRitter would spawnWeight be alright?

RodRitter:
@mmooxy Yes, that should be fine.

RodRitter:
I'm going to leave this for a little bit of community feedback first.

GamerInspired:
The more I think about the more I don't know which would be best. So will leave it to the community like @RodRitter .

mikejbrown:
Since this is a space themed game shouldn't we be using the term mass for, you know, the mass of things? There is no weight! :-)

ComputerOverlord:
Change it to rarity

koosemose:
The issue with a percentage chance is that it needs to be maintained to not exceed 1, whereas weight doesn't have to add up to a certain number.

RodRitter:
@koosemose I'm not saying the weight system doesn't work. I'm talking purely naming convention.

koosemose:
@RodRitter ahh, I thought you asking why weight was chosen over chance was in reference to the previous chance system as opposed to this weight system.

I'm of the opinion that we need to do whatever possible to include the word weight in there somehow, not using the proper terminology will cause confusion in how it works, and best case scenario require reference to specific docs on how it works.

Perhaps weightedChance? Using weighted over weight should break the mental connection from physical weight while still suggesting that it is using weight rather than percentages or some other method, with chance reinforcing the idea that it is odds of a thing happening.

RodRitter:
Being explicit is definitely the way to go. weightedChance sounds like a good clear name.

RodRitter:
Good job. Merged.

PR #682, opened 2016-08-23
Actions: refactoring interface

FilippoLeon:
I propose the following edit (adresses https://github.com/TeamPorcupine/ProjectPorcupine/issues/515, superseeds https://github.com/TeamPorcupine/ProjectPorcupine/pull/552).

What it does is generalize the concept of action as LUA script call.

In xml:
`<Action event="EventName" functionName="LUAFunc"/>`

In LUA or c#:
`Trigger(eventName, object, deltaTime)`
calls all registered LUA functions.

Actions are registered:
`Register("OnInstall", "LUAFunctioName")`
and deregistered accordingly.

Example is for OnUpdate, OnInstall, OnUninstall. But user and modder can create custom event and call custom function within LUA.

Is implemented as class, in future can be moved to its own file and applied to any other object needing LUA actions.

Missing: commenting and moving to separate file.

Also results in cleaner code.


svmnotn:
please fix the conflicts

vogonistic:
Nice! This seems like a more flexible solution.

When you've fixed the merge conflicts, remove `[conflict]` from the title and we'll test it.

FilippoLeon:
I fixed the conflict: but I have a weird but: the furniture menu is disappears. Is this a problem of master, my branch or my local copy? COuld somebody test please?

vogonistic:
There was a PR merged a few hours ago where the furniture menu is hidden when you make a selection. Is that what you mean?

FilippoLeon:
Won't be able to test for a while, maybe that s it.

FilippoLeon:
Seems to work, so can be merged.

TomMalbran:
I think that we should move EventAction to a new file so that the Character doesn't get too big.

Other than that, this looks good.

vogonistic:
This PR has conflicts, update the title when it's resolved. Thanks!

FilippoLeon:
I moved the class to its own file, and wrote an xmlwriter. Should be much better now.

vogonistic:
Merged!

PR #684, opened 2016-08-23
Adding custom Character spritesheet for Project Porcupine

Tranberry:
Making the sprite sheet available for all who wants to use it.


Tranberry:
same spritesheet layout as #444

PR #687, opened 2016-08-23
Fix caracter names parse text file

Ohda:
The names of my caracters was the whole text file.


JupiterJoe:
Rather than `File.ReadAllText `and `Regex.Split` you could use `File.ReadAllLines` to read all lines into the array directly. It takes care of `\r`, `\r\n` and `\n` line endings.

Ohda:
Thanks !

svmnotn:
@Ohda are characters able to do jobs after this for you? Because they are just siting there when I use your pr.

svmnotn:
Never mind that, it now \/
alt text

vogonistic:
Excellent! Merged! (I'm on a Mac)

PR #688, opened 2016-08-23
Fix for Metal Smelter and Cloning Pod (also code clean up)

Randshot:
I mainly fixed the metal smelter.
While I was at it, I also reformatted the whole file to make it readable.

### DIFF
--- a/Assets/StreamingAssets/LUA/Furniture.lua
+++ b/Assets/StreamingAssets/LUA/Furniture.lua
@@ -1,7 +1,7 @@
 -- TODO: Figure out the nicest way to have unified defines/enums
 -- between C# and Lua so we don't have to duplicate anything.
-ENTERABILITY_YES  = 0
-ENTERABILITY_NO   = 1
+ENTERABILITY_YES = 0
+ENTERABILITY_NO = 1
 ENTERABILITY_SOON = 2
 
 -- HOWTO Log:
@@ -11,9 +11,9 @@ ENTERABILITY_SOON = 2
 
 -------------------------------- Furniture Actions --------------------------------
 function OnUpdate_GasGenerator( furniture, deltaTime )
-	if ( furniture.HasPower() == false) then
-		return
-	end
+    if ( furniture.HasPower() == false) then
+        return
+    end
 
 	if ( furniture.tile.room == nil ) then
 		return "Furniture's room was null."
@@ -60,7 +60,7 @@ function IsEnterable_Door( furniture )
 		return ENTERABILITY_YES --ENTERABILITY.Yes
 	end
 
-	return ENTERABILITY_SOON --ENTERABILITY.Soon
+    return ENTERABILITY_SOON --ENTERABILITY.Soon
 end
 
 function GetSpriteName_Door( furniture )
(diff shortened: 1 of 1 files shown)


svmnotn:
Maybe split all the formatting to a different pr?

abackwood:
What did you change about the smelter? Between the style fixing I can't really tell

vogonistic:
@abackwood Pro-tip, add `?w=1` to the changes url and it won't show whitespace changes

vogonistic:
That said, @Randshot you should describe what was wrong with the smelter and what you did to fix it.

svmnotn:
I've been trying to figure it out... but... I can't reliably test due to characters just not working.

abackwood:
Seems to work okay (apart from bug #699), but until I know what it's meant to do, I can't really say much about it

vogonistic:
When you've fixed the conflicts and described it better, remove `[conflict]` from the title. Thanks!

Randshot:
I was in a bit of a hurry while creating the PR, so I couldn't describe what I changed.

I didn't like the 'ugly hack' that someone created, which involved setting a new parameter 'smelting' to check whether the metal smelter was working or not.
The problem with this approach is however, that the metal smelter will only start working, when an inventory of Raw Iron with a stack size of 50 is placed on the spawn spot.
I don't think however that that was how it was meant to work.
I also added the callback to MetalSmelter_Worked again, which someone had removed for no to me apparent reason.

abackwood:
I thought the smelting was more responsive. Like I said, it all works.

Backwood seal of approval :)

Randshot:
So, it now works like a charm for me.
I added that the metal smelter only starts smelting if there is at least an inventory with a stack size of 5, else it would also start smelting, even if there was only one raw iron.
This was no problem before, because it would only start once the job was completed which means when the 50 raw irons were delivered.

Randshot:
While I was testing this PR, I noticed that the cloning pod didn't work.
The problem was with the job creation which didn't match any Constructor.
I thought I would add this in this PR.

Randshot:
I know that I shouldn't change so many things in one PR but I have modified almost the whole file anyway, so I decided to declare variables as local, if they aren't needed outside of it's context.
Just as side note, variables in Lua are by default global (public) when declared and if you want them to be only accessible in the current context, you have to do so manually by making them local (private).
I personally think that using local variables is much cleaner and less prone to weird bugs, than having all of them global. We do it in C#, so why not in Lua?

Randshot:
I decided to cancel the metal smelter job in the job worked callback and then recreate it with altered desired (stack size) inventory.
I also noticed that the job completed callback doesn't seem to be called reliably. Haven't looked into this though.

Randshot:
Everything should be working.
It's bit annoying to resolve all these conflicts though.
But that's part of contributing, I guess :)

vogonistic:
I'm reviewing it for merge right now so hopefully it'll work fine. I understand the pains of keeping them up to date, but that's because we are changing the engine all the time. When we get to the point where we just change resources and tweak game behavior, it'll be much less merge conflicts.

vogonistic:
Merged!

PR #692, opened 2016-08-23
[Conflicts][sprite error] Heater needs atmosphere to operate

abackwood:
In the 0.1 milestone it is mentioned that some furnitures should require air to be present to operate. I think the Heater would be a prime test subject for this.

The Heater now has a Lua update function (it did nothing before) that normally increases the tile temperature every second using #778 system. If the pressure in the room is lower than 0.5 Atm the effectiveness decreases linearly until it does nothing in a vacuum.

Values are severely unbalanced as I'm not sure what normal ranges are. Temperature spreads extremely slowly, but that is due to the tmeperature system i think and outside the scope of this PR.

heaterscreenshot


FilippoLeon:
+1 for having air changes the strength of the heater. I also have to implement a power requirement based on actual ambient temperature,

For now you can use the Temperature.SetTemperature(int x, int y, float temp), etc... to change the temperature. The heater is not implemented.

As soon as https://github.com/TeamPorcupine/ProjectPorcupine/pull/682 gets merged, I have a fancy interface for the management of heat sinks and sourced.

abackwood:
@FilippoLeon Using Temeprature.SetTemperature(0 doesn't seem to do anything, hence my bypassing it for now. I a definite Temperature system is decided on this can be adapted to that in a later PR

FilippoLeon:
MMmmh, have you tried something like: SetTemperature(x,y,300)? It should be coordinates inside the ship and not a wall. I tough at least that worked....

abackwood:
Okay yeah, now it's doing somethng. My scales were way off. Switching this back to WIP while I work that out

FilippoLeon:
I originally implemented the underlying temperature in Kelvin (since it's space :)) so
0 K is 273 C is -459.67 F
300 K is 80.33 F is 26 C

This way I will make unhappy both Imperial and metric factions :).

abackwood:
Why do i need to add 100K per second to make a noticeable change?

FilippoLeon:
Are you adding, or setting the value to 100?

It might have to do with the fact that the temperature is updated in a different moment than all the rest.

If the actions system gets metged, I plan to have an OnTemperatureUpdate(), where one should modify the temperature.

FilippoLeon:
I like it.

We have to decide what's next. :)

Power usage by temperature difference?

Assets/Scripts/Models/Room.cs
```diff
@@ -173,6 +173,16 @@ public float GetGasAmount(string name)
...
     }
 
+    public float GetGasPressure()
+    {
+        float pressure = 0;
+        foreach(float p in atmosphericGasses.Values)
+        {
+            pressure += p;
+        }
+        return pressure;
```
Randshot: Insert a new line before the return statement.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -579,7 +579,18 @@ end
 -- Dummy heater uninstall function
 -- THis function gets called once, when the funriture is unisntalled
 function Heater_UpdateTemperature( furniture, deltaTime)
-	World.current.temperature.SetTemperature(furniture.tile.X, furniture.tile.Y, 300)
+    --if (furniture.HasPower() == false) then
+    --    return
+    --end
```
Randshot: Avoid pushing commented out code.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -579,7 +579,18 @@ end
...
+    
+    tile = furniture.tile
+    pressure = tile.Room.GetGasPressure() / tile.Room.GetSize()
+    efficiency = ModUtils.Clamp01(pressure / furniture.Parameters["pressure_threshold"].ToFloat())
+    temperatureChangePerSecond = furniture.Parameters["base_heating"].ToFloat() * efficiency
+    temperatureChange = temperatureChangePerSecond * deltaTime
```
Randshot: Try to use 'local' if the var is not needed out of it's context.
The Lua 'local' is the equivalent to C# 'private#.

Mizipzor:

Assets/Scripts/Models/Temperature.cs
```diff
@@ -1,6 +1,6 @@
 ﻿#region License
 // ====================================================
-// Project Porcupine Copyright(C) 2016 Team Porcupine
+// Project Porcupine Copyright(c) 2016 Team Porcupine
```
Dormanil: Was that change to the license deliberate? :)

Mizipzor: I would assume so, its a capital c in the other license header. Fun fact: theres a stylecop rule that forces you to use the actual copyright character.

abackwood: I don't know what that was about. I got the conflict after SCS and just rolled with it. It was nothing to do with my work

Mizipzor:
I just tested this. Is the sprites missing? I heater queued for construction is dark purple and a constructed one is pink.

abackwood:
Broke the link while changing object type to style standard. Will fix when im near my laptop again

Mizipzor:
Im still only get purple and pink sprites, dont know what im doing wrong, but I cant merge. We need someone else to test this. Try discord.

Dormanil:
Nice and pink. I fixed the issue with a little bit of Windows renaming magic. I sent you a PR for that fix, @abackwood.

vogonistic:
I don't see a merge of anything to fix the sprites. @abackwood Are you on that?

abackwood:
I'll have another look today. It's strange that I've never had the issue but others do.

Tranberry:
didn't that PR from @Dormanil fix the sprite error?

Dormanil:
It did on my end at least. 
But renaming a file from "Heater.png" to "heater.png" is always fun on Windows, which doesn't notice that you changed the name.

Dormanil:
But even git knows now that it was renamed. 
snippet

abackwood:
I'm going to close this. It's been a strange mess of things that I can't seem to fix and I'd rather move on to other things. This was meant to be a minor adjustment of existing behaviour.

Someone else who wants to mess with the heater can now do so without fear of conflicts.

Dormanil:
These changes are now being implemented in #1113 (I was only just 2 issues/PRs late).

Issue #639, opened 2016-08-22
Can't deconstruct Furniture


GamerInspired:
Can't seem to decontruct funiture at the moment.

dustinwloring1988:
I will fix in a PR I have pending

dustinwloring1988:
Fixed

Mizipzor:
Pull request was never accepted, I assume this is still an issue.

GamerInspired:
I can't currently test this as I'm away from my main machine. Could someone else check this for me.

Mizipzor:
No rush. :)

abackwood:
Can confirm. The furniture deconstruct button doesn't seem to be doing anything

GamerInspired:
Just found out during my once a day (minimum) test session. That you can deconstuct things but only using the right click and selecting deconstruct method.

sboigelot:
Deconstruction works again (from CM & from the build menu button).
We can close this issue

PR #701, opened 2016-08-23
No longer deconstructs walls when this would cause depressurisation

### LINKED ISSUE #685: No deconstruction of walls under pressure
A requested feature for 0.1 is that pressurized rooms can't simply be breached by deconstructing walls.

This might be fun for someone relatively new to programming to pick up.

Deconstructing happens instantly as of commit #629. BuildModeController's DoBuild() method takes care of that. A check could be added there against accidentally breaching a room.

Any implementation of this would need to figure out what constitutes a "breaching hazard". A wall has access to its neighboring tiles, so pressure info could be accessed from there.

Hope this is helpful :)

mikejbrown:
Fixes #685 
Note: need the fix for #639 to test this.

This fix affects deconstructing through the construction menu but not by the right click menu.

### DIFF
--- a/Assets/Scripts/Controllers/BuildModeController.cs
+++ b/Assets/Scripts/Controllers/BuildModeController.cs
@@ -181,6 +181,33 @@ public class BuildModeController
             // TODO
             if (t.Furniture != null)
             {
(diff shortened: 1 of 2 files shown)


vogonistic:
Merged!

PR #707, opened 2016-08-24
[Conflicts] UI Menus Loaded At Runtime

dustinwloring1988:


dustinwloring1988:
NO MORE MESSING WITH THE SCENE FILE

Some screenshots below: 

hierarchy

grid

-Change the construction menus to use the grid layout (you can change this now with out touching the SCENE)

Right now I'm using a work around and loading all the menus for one game object that has the prefabs of each menu as a child of it. (TempMenus).

I tried to load a a game object that then loads each menu for you but it does not position them right. (UIMenus). 

If you do change any menu just change the menu its self (ie  MenuRight, MenuTop ) not (TempMenus).

I think we should still merge this now and work on the menus loading individually by code later.

vogonistic:
What's the reason you changed the UI to be in a grid?
Is this a WIP or final? I'm asking because I don't usually merge files that has `Temp` in the name.

dustinwloring1988:
This PR is final the next one will be resizing try to keep each one focused on one issue.

The Temp is used to hold the menus so there position stays the same on the UI. If you load a menu by its self it will just goto the center. 
Temp contains all the menu prefabs so you can still change each menu separately.

The reason for the grid is for when the Canvas rescale is set. This will allow you to not have a column going straight up it will end up filling side ways like RimWorld.

Just submitting this before I tackle resizing so I can make my PR's small like many have asked.

vogonistic:
This PR has a couple of problems:
  1. Multiple FPS counters in the same place
  1. Overlays are broken

After testing it, I'm still not a fan of the Grid view. Can we get additional input on that? @WardBenjamin @RodRitter @Mizipzor

RodRitter:
Haven;t been able to test yet, but I also don't like the grid view. It doesn't seem scalable.

Mizipzor:
@vogonistic I care less about the look and more about that this is moving more stuff out of the scene.But I didnt actually pull this to have a proper look.

TomMalbran:
The grid view will look nice once we change the ugly buttons with nicer ones with images, and making them squares like in RimWorld.

dustinwloring1988:
@Mizipzor yes that is the main point of this upload.

@TomMalbran I agree.

@vogonistic  I can fix the problems in about 9 hours after work.
As for the grid view after I make the scaleable script it would like alot better. But anybody can change it all I ask is you give it a try even if we only want it 2 colums wide. Atleast until we get sub menus.

Mizipzor:
@dustinwloring1988 Could we split this work up in two? One pull request to lift the stuff out of the scene, and another for a redesign of the menus? I have a feeling that the grid will be controversial and we should let more people weigh in, which takes time, which makes difficult to merge scene changes.

dustinwloring1988:
@Mizipzor  ok no problem will do when I get home, will you be able to add it tonight then.It will be in about 7.5 hrs from now

Mizipzor:
Did you just, like, add every file to Git? :P

dustinwloring1988:
lol. I did not notice that, just tried to update my files. This is my first week ever using GitHub. Ill just redo the whole thing later.

dustinwloring1988:
If you guys could merge all PR's that will effect the UI in the next couple hours. Ill wait until then and have it done by morning. Please try to add the new PR I do in the morning waiting a day makes it hard for me to manage the changes.
Thanks

PR #718, opened 2016-08-24
Add Autoscroll to UberLogger

Fogsight:
It's not perfect, but it works.

### DIFF
--- a/Assets/UberLogger/Editor/UberLoggerEditorWindow.cs
+++ b/Assets/UberLogger/Editor/UberLoggerEditorWindow.cs
@@ -13,6 +13,11 @@ using UberLogger;
 
 public class UberLoggerEditorWindow : EditorWindow
 {
+    //Settings
+    float logDetailsWindowHeight = 100;       // Customise details window size
+    int pushStep = 50;
+    int lastItemsCount = 0;
+
     [MenuItem("Window/Show Uber Console")]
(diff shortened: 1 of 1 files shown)


Mizipzor:
Nice, that was really bugging me. UberLogger is here on GitHub, I actually created [an issue](https://github.com/bbbscarter/UberLogger/issues/1) for it and was planning to fix it. Maybe you can beat me to it? You seem to hint at there is room for improvement here. Let me just test it and I'll merge.

Randshot:
I feel dumb now, but how do I open the UberLogger?
Until now I always used the console output from Unity3D.

Mizipzor:
@Randshot In Unity menu bar "Windows -> Show Uber Console". I really need to find the time to create a wiki page for it.

Mizipzor:
@Fogsight Now I see how it works, every time a line is added it scrolls one line downwards. I would expect it to always scroll to bottom. Maybe with a button to toggle. But this is a step in the right direction. :)

Fogsight:
I spent way to much time on that simple change, I wanted to scroll exactly one line each time, but it has a lot of kinks and won't work like that in every case, so I could not be precise. Without extensive rewrite you can't control its behavior. Also I didn't touch its "In-app" version, so yea, there's a lot of room to improvement.

PR #719, opened 2016-08-24
Fixed log in localization

Sommerbo:
Switched to UberLogger. I had to add "ToString()" in a few places, nothing broke but I didn't test the specific exceptions.

Removed a line break that caused log line to be truncated and meaningless


Sommerbo:
Redoing this

PR #720, opened 2016-08-24
Switched to Uberlogger and fixed a typo

Sommerbo:
Had to add ToString() in a few places. Checked exceptions and output is identical
Removed a line break that truncated the log message rendering it useless.


Mizipzor:
Trivial change, no need to tests. :)

PR #722, opened 2016-08-24
New Sprites for Wall and Power Generator

Spool5520:


GamerInspired:
Looking good :smile:

abackwood:
Nice. It was about time those got replaced :smiley:

If you don't mind me being a little honest:
- The walls have kind of a concrete vibe to them IMO. They look a little rugged and heavy for steel walls
- I would personally prefer a more united power gen. This looks like a few different furnitures stitched together.

I like the radiation symbol. Drives home that it's a nuclear generator that is supplied with uranium.

Spool5520:
Yeah I know what you mean. I just kind of threw these together so the perspective would be in place. I plan to redo them with more variants of wall types. Yeah about the generator again just threw it together for perspective. I don't really like it to be honest and do plan to make it more "together" or one generator. Thanks for the feedback. :)

Spool5520:
Can this be pushed to master or is there something I need to do?

bjubes:
i think the pixels per unit of the power generator is off, shouldn't it be at 128?

Spool5520:
It's supposed to take up a 4x4 tile area. Atleast that's what the original was.

Spool5520:
2x2 my bad

Spool5520:
It works in-game

vogonistic:
Merged!

PR #728, opened 2016-08-24
Mouse Cursor

Tritaris:
Adds new mouse cursor to override the default one. Hold Ctrl and press M to toggle it on and off.
[Test Videos](https://www.youtube.com/playlist?list=PLQOtekaZrrsET7qxzPBzSvrvc0JyiQZuP)

This lends itself to be usefully as a dev tool or for the player to display information. Eventually will be setup to be customizable using keywords like `@mouse_position` or `@valid_position_count` in a settings window to display different things. 

Art by longtomjr


abackwood:
Tested. Works perfectly
I assume the green and yellow are for furnitures and characters respectively?

Tritaris:
Thanks @abackwood . Yes you are correct. Once we get the UI setup more i'll make a settings screen to be able to change the colors to whatever we what.

GamerInspired:
Would be possible to have an option to change the size of the Mouse Cursor UI? At a later date of course.

Tritaris:
@GamerInspired  Yeah I believe it will be possible. At the moment I am using the smallest font size of 1. I believe I have to set the canvas to a different rendermode, or I'd have to scale the map up. The latter would require changing how the map get made and positioned.

RodRitter:
I can't seem to see any different mouse cursor. Is there a setting I need to turn on?

Otherwise from what I can see in the preview video, could you maybe scale them down a bit so it's not too intrusive. Unless its only for debugging purposes?

Tritaris:
@RodRitter Yeah there is, you have to hold "Control" and press "M" to turn the new cursor on. 

I am going to try today and figure out the scaling issue. I believe its due to the way I have its canvas setup. It is currently set as `cursor_canvas.renderMode = RenderMode.WorldSpace;` so its drawing it in the world. Which is why it looks so big at font size of 1, its drawing it 1 tile high and changes when you zoom out. Swtiching to `RenderMode.Overlay` and rearranging so thing it should look better.

vogonistic:
Merged!

PR #674, opened 2016-08-23
[Ready] Animated character color change

Raketfart:
This PR depends on PR #669 and changes 3 colors in character sprite to use the random character color. The replaced colors are:
Color32(71, 86, 163, 255);
Color32(142, 154, 212, 255);
Color32(42, 56, 126, 255);

PR #731, opened 2016-08-24
Character - Using new spritesheet

Raketfart:
Changed the character spritesheet to use the one from PR #684.
Removed the floating helmet.
Colors are still applied to the entire character, but that will change when PR #674 is done.


vogonistic:
Should we wait for #674?

Raketfart:
No, this can be merged so everyone can enjoy the new graphics. #674 will build on this, but it still needs some work.

PR #733, opened 2016-08-24
Added spritesheets for new Door graphics.

longtomjr:
I created door sprites for horizontal and vertical doors. Both animations have 6 Frames.
The original door sprite has not been removed because there is still some code I that needs to be tweaked for the sprites to appear ingame.

Here is the gifs for preview:
doorhorizontal
dooranimated


RodRitter:
Good job. Merged.

PR #737, opened 2016-08-24
Lua utilities

NogginBops:
Adds lua utilities for loading and calling lua functions.


RodRitter:
Tested and seems to be good. Great refactor! Merged

PR #738, opened 2016-08-24
[Small]Follow up on PR #720. Missed one instance and removed unneeded comment.

Sommerbo:
Minor minor fix.


bjubes:
you can add [Small] to the title so the gatekeepers know this is a 2 liner

PR #678, opened 2016-08-23
Added Zoom Lerping - Smooth zooming

RodRitter:
A quick addition while I had a chance. Added some nice smooth zooming action. :sunglasses: 

Also added it's settings into the settings.xml - I'm sure someone can add it to the settings modal at some point.

PR #749, opened 2016-08-24
Fixed mouse zoom focus position not working from #678

NoobPete:
Made zoom focus on mouse position like in #526


RodRitter:
Good job! Merged

Issue #729, opened 2016-08-24
[Discussion/Enhancement] Gas Compressors


abackwood:
There has been some talk of making the O2-Generator require ice to operate. I think that makes a lot of sense and is cool. However, that could leave the player with no way to produce oxygen at the start if there is no ie nearby. It would be lame for the characters to suffocate so quickly.

Therefore, I was thinking of adding Gas Compressors. They are essentially big tanks that release their gas into the air when there is too little and slowly refills when there is enough. In vacuum they lock down and don't release anything.

The starting ship/station would then have a compressor filled with O2, given the player a grace period to set up production while they live off the reserves.

Thoughts?

GamerInspired:
Seems like a good idea to me, would also other uses later on, as it sounds like you could pump gas into it later on (once pipes are in) making it more of a gas tank.

abackwood:
That's my thought too. It could be tied into the pipe system. But while that's not there (or as default behaviour) it works locally

GamerInspired:
I think once pipes are in then there should be a separate furniture that does the releasing of gas. But maybe also having a early game all in one system would be a good start but also nice to have for progression sake.

vogonistic:
Sounds like what we are going to need in airlocks too. When you want to go out, we pump the gas into a tank and when the outer door is closed, we can put it back.

abackwood:
I considered making a general compressor that just takes "air" instead of O2, but I wanted to see if this worked first before diving into it.

For the airlock, that general one would be needed, and it needs to be changeable by the characters. They press a button and the room is sucked into the tank.

sboigelot:
As oxygen compressor are in the game now, should we close this issue? There is also many issue about how should the airlock work.

PR #751, opened 2016-08-24
Adding Oxygen Compressor

abackwood:
As suggested in #729 this adds an Oxygen Compressor that acts as a storage buffer for a room's oxygen supply.

The compressor starts taking in O2 if the partial pressure rises above a threshold (in this case 0.01 Atm) and stores it in its internal tank. The relative level of the tank can be seen in the meter. When room O2 pressure drops below a certain (lower) threshold (0.005 Atm here) the compressor starts trying to backfill that with its own reserves.

The initial intended use of this machine, and similar machines for other gases, is to provide oxygen in the very early game. The starter station can come with a filled O2-Compressor that can serve as a substitute for the O2-Generator until that one can be built and stocked with ice (or whatever resource). 

The system can further tie into a pipe/utility system. If hooked up to a pipe network, it shares its contents. This thing can then serve the role of inefficient early hardware later to be replaced by dedicated life support systems.

- It is 2x2 now. Should it be smaller? Larger?
- Sprites are placeholder. I forfeit all emotional attachment to them and the consideration that comes with it

Here's how it works in action:
The oxygen generator is producing oxygen, but the compressor is inactive as there is not enough to fill itself up
o2comp1

The compressor detects that enough oxygen is present and starts storing it for later use. At full capacity, the compressor contains enough O2 to bring a 25 tile room to 0.2 Atm Oxygen
o2comp2

The power generator has tragically died and now the oxygen generator is useless. The compressor detects a low O2 level and starts trying to back-fill the room from its reserves
o2comp3


GamerInspired:
I think it's a good size because as discussed (opinion only) in #729 we would probably want to have a more advanced system later on in the game (once researched) which may include different size tanks, and have the parts of the system separated.

Also I like the level bar that shows how much gas it is storing. Nice edition :smiley:

abackwood:
Thanks :smiley: 

I wanted to try my hand at sprite sheets and changing between them. That being said, i recognise that it's not at all in the art style that we're going for.

GamerInspired:
But it shows what can be done; especially as it will be easier with the completion of the animation controller that people are working.

I hope something the bar showing have much is inside, is use with other things later on in the development cycle.

RodRitter:
Tested and works well. Very cool addition. Merged.

PR #753, opened 2016-08-24
[small] Changed localization missing log to info level

### LINKED ISSUE #745: Localization downloader throws non-error errors
Right now, the Localization table will throw errors if new Localization is downloaded or when a key is not present. This seems overly dramatic. Errors are problems that mean the program is not running well. "Hey a new localization has been downloaded" is working as designed and doesn't need the game to pause to make me check what's wrong?

Same with missing loc keys. It's not game-breaking if there isn't a text available for some new furniture I'm implementing.

Just throw a warning for both or something.
Or am I not understanding what Errors and Warnings are supposed to be. I am open to correction.

Ohda:
Fixes #745


vogonistic:
Merged!

PR #755, opened 2016-08-24
Cleanup

cskiwi:
- removed unused Furniture instance
- removed unused GameObject refrence from dialogBox
- removed unused DropDownObject

- Added override for ToString()  on Parameter.cs
- Logged error on overmayDescriptor

- initialized sinksAndSources (was null, but you could register extra on it ... )


RodRitter:
Good job. Merged

PR #759, opened 2016-08-24
[WIP] Utility Objects

SteelFall:
This does not yet contain pipes but is a base for utility objects. Most of the scripts are just adaptations of the furniture scripts and will be modified as needed in the future.


NogginBops:
You shouldn't create a new UtilitiActions class, use the LuaUtilities class instead instead

SteelFall:
@NogginBops Ah yes, I saw that when I encountered some conflicts. Just forgot to remove the class.

SteelFall:
I appear to still need it for parts of the BuildModeController but I will change the UtilityActions class to look like the updated FurnitureActions class.

SteelFall:
Trying to get things fixed has started to break more things.

SteelFall:
Encountering technical difficulties which removed my classes and will not let me revert to a stable commit. Will redo everything tomorrow but for now this PR is closed.

PR #760, opened 2016-08-24
[WIP] Added a Room graph

mattpott:
This should create a graph that shows how each room is connect to each other. Currently it is not used in the path finding system but that is work for the future.


Jorchay:
Could you add a photo?

ComputerOverlord:
don't think it does have one

FilippoLeon:
Good job, this was long due. I was about thinking implementing myself: I know how a PITA implementing this has been.

@ComputerOverlord uh?

TomMalbran:
@mattpott If you want to remove the ```GenerateEdgesOutside()``` function, you can implement it after you find a new room in ```GenerateEdgesByRoom(Room r)```, you can check if that room is the outside room and if it is you create a new edge from the new room (the outside one) to ```r``` (the current room).

ComputerOverlord:
@FilippoLeon Someone was asking for an image

mattpott:
@TomMalbran I had thought to do that first but then I ran into some issues with combining two arrays which I wasn't quite sure how to do. 

Also in general I am not sure what to do about the movement cost between nodes and would love any suggestions. At the moment I made it default to 1 and just treat every edge as the same cost but in reality this is not the case as moving through a small room is quicker than moving through a big room.

TomMalbran:
@mattpott I haven't really tested it, so I am not sure if it works, but my idea was to do something like:

At the ```Path_RoomGraph``` constructor you create a Node for the outside room, and a list of edges for that node. You then pass de outside edges list to ```GenerateEdgesByRoom```, and after line 90 of that method you add something like:

```
if (t2.Room.IsOutsideRoom())
{
    Path_Edge<Room> edge = new Path_Edge<Room>();
    edge.cost = 1;
    edge.node = nodes[r];
    outsideEdges.Add(edge);
}
```
Then on line 53 you can do something like ```outsideNode.edges = outsideEdges.ToArray();```


For the movement cost, it will depend on how it is integrated with the path system. I dont know how other games do it. A couple of posibilites is to just calculate the door to door distance in tiles, or the actual movement cost of walking from one door to the other, that can be updated when the furniture, room, floor changes.

koosemose:
It would be greatly appreciated if an API to look up the rooms connected to a given door was able to be made available.

mikejbrown:
This is very good! I just seem to be getting NullReferenceExceptions. Can you confirm https://github.com/mattpott/ProjectPorcupine/pull/1 is a fix?

mattpott:
@mikejbrown I have merged the fix, I also fixed a few extra bits but i don't have the ability to test this at the moment but I will check it when i get home tonight.

mikejbrown:
It's testing fine for me, but I've only had a quick look at it so far.

mattpott:
@mikejbrown Is it generating the correct graph and edges? you should be able to check in the uber logger

mattpott:
Also the graph shouldn't be updating properly when you delete a room

mikejbrown:
Yes, it appears to be generating the correct room. It should be possible to wrap this up into an automatic test (assuming we start with the default map).

mattpott:
yep as long as there is no rooms outside of the ship then the graph should start the same everytime.

mattpott:
One last thing to add is regenerating the graph when a new room is added / destroyed.

mattpott:
This should be hopefully ready now, I will refactor later to make it look nicer.

mikejbrown:
Works for me. :+1: I really like this PR :smile:

mattpott:
yay it no longer conflict.

mattpott:
There are some strange bugs that I am still trying to iron out.

Mizipzor:
This is the second oldest pull request and I feel bad for letting it hang for so long. Poke us when youve ironed out the bugs. :)

mattpott:
It is just frustrating as I am away for 2 days at the moment so I can't work on it. The bugs seem to be when you are creating rooms as they are not properly being cleaned up in the code.

Dormanil:
Your test code seems to have run without problems, even though I couldn't see the UberLogger log messages you sent. Debugging in Visual Studio let me run through the code though, so I know that no errors happened.

abackwood:
Code works on my machine. hen I add a set of new rooms the graph has regenerated correctly. Good job :+1:

Mizipzor:
@abackwood thanks for testing. :)

PR #761, opened 2016-08-24
[WIP] Power system initial prototyping

Grenkin1988:
This is initial prototyping of new "live" power system.
PowerSystem would contain different PowerGrids. This is made for future enhancement with power cables.
PowerGrid containes list of PowerRelated objects connected to this grid.
Each Update() run grid check if it is still operating.


Grenkin1988:
I think `PowerRelated` covers most possible situations:
1.	Power generator:
```new PowerRelated
        {
            OutputRate = 10,
            InputRate = 0,
            Capacity = 0
        };```
2.	Power consumer:
```new PowerRelated
        {
            OutputRate = 0,
            InputRate = 5,
            Capacity = 0
        };```
3.	Battery:
```new PowerRelated
        {
            OutputRate = 10,
            InputRate = 10,
            Capacity = 100
        };```
If you can come up with more please tell me.

vogonistic:
That looks good! Does `Capacity` mean that every object can buffer a bit of power on their own?

I noticed that you aren't exposing it to Lua. How does a furniture change the values or turn itself on and off?

Grenkin1988:
For now I think that `Furniture` class is going to have property
```public PowerRelated PowerRelated { get; private set; }```
It would be `null` if furniture is not related to power.

NogginBops:
I think the name `PowerRelated` needs to be thought about some more... There just needs to exist a more elegant wording...

vogonistic:
I agree on the name, but I can't think of something better right now. `PowerDemands` implies it uses power only. `PowerInterface`? `PowerParameters`? I'll let you know if I think of something I like.

bjubes:
`powerRequirement`?

Grenkin1988:
@NogginBops @vogonistic @bjubes How about `PowerAgent` ?

NogginBops:
I could get behind that.

Grenkin1988:
Oh shit... :-( So screw this up ...

PR #763, opened 2016-08-24
Making Oxygen Compressor meet naming standard

abackwood:


racerand:
Wouldn't it be smarter to get the basename as the objectType of the furniture?

abackwood:
@racerand How do you mean?

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -563,7 +563,7 @@ function OxygenCompressor_OnUpdate(furniture, deltaTime)
 end
 
 function OxygenCompressor_GetSpriteName(furniture)
-    baseName = "Oxygen_Compressor"
+    baseName = "compressor_oxygen"
```
NogginBops: @abackwood Here you should get the name of the furniture programmatically instead of having it hardcoded.

abackwood:
Good point. Will fix in a bit

vogonistic:
Please fix the merge conflicts and remove `[conflict]` from the title when it's done. Thanks!

PR #742, opened 2016-08-24
Created XML files for DoorHorizontal and DoorVertical spritesheets.

longtomjr:
Just created the XML files. Now the only step left is to do the animation in the lua controller and fix the furniturecontroler to use the different sprites

PR #766, opened 2016-08-24
Door sprites updated

TheRiven:
This replaces the old door sprites with the new ones created by @longtomjr

It adds a flag to the furniture (probably not the final solution!) that is then set by the `FurnitureSpriteController` instead of rotating the game object.

This is checked by the Lua code in `GetSpriteName_Door` to determine which sprite sheet to use.

`openness` still determines which sprite to use but now has a wider selection.

Includes the XML from #742 also created by @longtomjr


RodRitter:
It seems the horizontal doors aren't working quite right.

image

longtomjr:
I am going to bed now, that is an issue with the sprites, I will fix it in
the morning

RodRitter:
@longtomjr That's ok. It's only a sprite adjustment. It still looks good.

PR #746, opened 2016-08-24
Fix for [BUG] KeyNotFoundException in Parameter.cs

mikejbrown:
Fixes #725 
Related #740 #743 

Would be better to get a Lua line number. Is that possible at all?

PR #773, opened 2016-08-25
Added unit tests for Parameters

mikejbrown:
It's good practice to start adding as much testing as possible. Here's a start on a particularly easy to test class: `Parameters`.

Uses Unity Editor tests based on NUnit. For those unfamiliar (assuming there are some newbies around) see the [Unity docs](https://docs.unity3d.com/Manual/testing-editortestsrunner.html) and [NUnit docs](http://www.nunit.org/index.php?p=quickStart&r=2.6.4). Run the tests in the Unity editor by going Window > Editor Tests Runner then clicking Run All.

Had to make some minor fixes to `Parameters` to get all the tests running. Some test(s) will have to be changed if #746 is merged.

From now on changes to the `Parameters` API can be spec'ed out in the tests in terms of the _expected_/_desired_ behavior, and _then_ the implementation coded to make the tests pass (test driven development).

### DIFF
--- a/Assets/Scripts/Models/Parameter.cs
+++ b/Assets/Scripts/Models/Parameter.cs
@@ -138,13 +138,13 @@ public class Parameter {
 
         writer.WriteStartElement("Param");
         writer.WriteAttributeString("name", name);
-        if (!value.Equals(""))
+        if (value != null && !value.Equals(""))
         {
             writer.WriteAttributeString("value", value);
         }
         foreach (string k in contents.Keys)
         {
-            this["k"].WriteXml(writer);
+            this[k].WriteXml(writer);
         }
         writer.WriteEndElement();
     }
--- a/Assets/Scripts/Models/Parameter.cs
+++ b/Assets/Scripts/Models/Parameter.cs
@@ -138,7 +138,7 @@ public class Parameter {
 
         writer.WriteStartElement("Param");
         writer.WriteAttributeString("name", name);
-        if (value != null && !value.Equals(""))
+        if (string.IsNullOrEmpty(value) == false)
         {
             writer.WriteAttributeString("value", value);
         }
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Models/Parameter.cs
```diff
@@ -138,13 +138,13 @@ public void WriteXmlParamGroup(XmlWriter writer)
 
         writer.WriteStartElement("Param");
         writer.WriteAttributeString("name", name);
-        if (!value.Equals(""))
+        if (value != null && !value.Equals(""))
```
svmnotn: here you could just use `string.IsNullOrEmpty(value) == false`

mikejbrown: I was not aware of that function. Made the change. Thank you! :+1:

koosemose:
From a read through, the tests look good... though seeing that I apparently wrapped what should be a variable in quotes... wow I was tired when I got to WriteXml...

mikejbrown:
@koosemose We've all been there. That's why it's so important to write tests!!

Assets/Scripts/Models/Parameter.cs
svmnotn: No problem, it is a very hidden very useful function.

Assets/Scripts/Models/Parameter.cs
```diff
@@ -138,13 +138,13 @@ public void WriteXmlParamGroup(XmlWriter writer)
...
         writer.WriteAttributeString("name", name);
-        if (!value.Equals(""))
+        if (string.IsNullOrEmpty(value) == false)
         {
             writer.WriteAttributeString("value", value);
         }
         foreach (string k in contents.Keys)
         {
-            this["k"].WriteXml(writer);
+            this[k].WriteXml(writer);
```
Grenkin1988: k -> key

vogonistic:
Merged! Thanks for starting the work with adding tests. I've figured out how to run them from the cli as well, but one of the tests fails now that we've fixed the exception.

From inside the ProjectPorcupine directory (with Unity closed): `/Applications/Unity/Unity.app/Contents/MacOS/Unity -batchmode -runEditorTests -quit -projectPath $PWD`

PR #775, opened 2016-08-25
Added visual path for selected characters

MANDAL0R3:
With this PR we are now able to see where the characters are going similar to RimWorld. Once someone implements being able to select characters, then this should be changed to only show who is currently selected. But i think this is a good addition right now to help show where the characters are going and easily see if there is any hiccups in the path. All of the line drawing was done with GL and to my knowledge, should be pretty efficient but this is my first time actually working with GL.

image

Edit: removed second picture because now only one character path will be shown.

### DIFF
--- a/Assets/Scripts/Controllers/MouseController.cs
+++ b/Assets/Scripts/Controllers/MouseController.cs
@@ -244,8 +244,13 @@ public class MouseController
 
             if (mySelection == null || mySelection.tile != tileUnderMouse)
             {
+                if (mySelection != null)
+                {
+                    mySelection.GetSelectedStuff().IsSelected = false;
+                }
                 // We have just selected a brand new tile, reset the info.
                 mySelection = new SelectionInfo(tileUnderMouse);
+                mySelection.GetSelectedStuff().IsSelected = true;
             }
             else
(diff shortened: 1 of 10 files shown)


TomMalbran:
You can already select a character by clicking on it. The character info will appear on the right grey panel. You can ask the MouseController for the current selection and if it is a character, cast it to one. The mouse controller might need a function that returns the selected character.

MANDAL0R3:
I figured it was more like selecting the tile.

MANDAL0R3:
I have changed it to where only the character that is selected will have their path shown.

GamerInspired:
Maybe we could have an overlay that shows all character paths in the future?

MANDAL0R3:
An overlay in the future would be nice but im not entirely sure how to go about that yet.

Mizipzor:
I dont get the red line the first time the character walks out to get materials. But subsequent walks renders it. I agree that this should *probably* be an overlay. And the aesthetics is... questionable. ;)

However, Im gonna take this anyway. Its a good start and someone else might build upon it. I also think its a great help to anyone debugging path finding code. :)

PR #778, opened 2016-08-25
Temperature finally doing something useful

FilippoLeon:
The temperature will now increase when an heater is placed.

Some parameter values are also fixed.

Temperature is now also non-blocking.


Assets/Scripts/Models/Furniture.cs
```diff
@@ -34,7 +34,7 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider,
     /// These actions are called when Trigger is called. They get passed the furniture
     /// they belong to, plus a deltaTime (which defaults to 0).
     /// </summary>
-    protected EventAction eventActions;
+    public EventAction eventActions;
```
Grenkin1988: No public fields in class

Assets/Scripts/Models/Temperature.cs
```diff
@@ -27,13 +28,7 @@ public class Temperature
...
     /// All heaters and refrigerators shoul register themselves here using the public interface
     /// </summary>
-    Dictionary<Furniture, Action> sinksAndSources;
-
-    /// <summary>
-    /// All heaters and refrigerators shoul register an Action here
-    /// TODO: find elegant way to modify this through LUA
-    /// </summary>
-    Action sinksAndSourcesActions;
+    Dictionary<Furniture, Action<float>> sinksAndSources;
```
Grenkin1988: Accses modifiers should be implicitly specified

Grenkin1988:
Please run StyleCop for Temperature class

abackwood:
This is going to conflict with #692 and does essentially the same thing. Why make a duplicate PR?

FilippoLeon:
Is going to conflict, but we just have to move your implementation OnUpdate for heater to the lua function OnTemperatureUpdate. This is just because the temp. will be update on a different frame. Otherwise the conflict should be minor.

Assets/Scripts/Models/Furniture.cs
Mizipzor: This can easily be replaced with an automatic property. Do that.

`public EventAction EventAction { get; set; }`

FilippoLeon: May I ask why you are strictly forbidding public fields? To me is just extra boilerplate.

Assets/Scripts/Models/Temperature.cs
FilippoLeon: Also this one, I do not really get it.

abackwood:
Sounds good. I'll wait for yours to be comitted then ans change mine accordingly.

FilippoLeon:
Fixed conflict.

Mizipzor:
Hm. This is a tricky one. I built a heater. But once built it is invisible, I can still select it though. I would like to see that fixed. 

However, youre touching many lines so this will probably be tricky to merge and right now its conflict free. And, @abackwood is waiting on this.

Since youre not breaking anything current, as far as i can tell, Im gonna take a risk on this one. @abackwood @FilippoLeon Please congregate your works and test once youre both done, mkay? :)

FilippoLeon:
I can't reproduce on the current master. Mhhhhh... Need to think about this.

PR #779, opened 2016-08-25
Merge dev mode and dev tools and add UI toggle

koosemose:
The Dev tools previously implemented (spawn inventory system) didn't have a UI element to turn it on and off, and my previously implemented dev mode didn't either, and only had a hackish solution (toggling the option on the world controller game object or manually setting a boolean). Since they both had the same general intent (tools to aid development, and possible later use in map creation) I merged them into one setting, and added a toggle to the settings menu.

Additionally the spawn inventory controller was slightly tweaked to allow it to be turned on and off during gameplay.

### DIFF
--- a/Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
+++ b/Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
@@ -67,11 +67,6 @@ public class DialogBoxSettings : DialogBox
                 OnFullScreenToggle();
             });
 
-        developerModeToggle.onValueChanged.AddListener(delegate
-            {
-                OnDeveloperModeToggle();
-            });
-
         resolutionDropdown.onValueChanged.AddListener(delegate
             {
                 OnResolutionChange();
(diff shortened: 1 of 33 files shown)


Mizipzor:
The settings menu is now visible by default when starting the game, cant have that. Otherwise looks good, fix and I'll merge. :)

koosemose:
Whoops, forgot that such things don't save until the scene saves.

Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
```diff
@@ -100,6 +107,10 @@ public void OnLangageToggle()
         langDropDown.SetActive(langToggle.isOn);
     }
 
+    public void OnDeveloperModeToggle()
+    {
+    }
```
vogonistic: Is `OnDeveloperModeToggle()` supposed to do something? Otherwise, maybe it should be removed. If you remove it, take out the listener as well.

vogonistic:
Remove `<DevTools_enabled>True</DevTools_enabled>` from the Settings.xml. Minor comments, ready to merge when fixed.

Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
koosemose: Honestly I only put it there for consistency with everything else. Removing it now.

koosemose:
Git went a little nuts, but everything should be fixed now.

Mizipzor:
The settings line has been deleted, merging.

PR #780, opened 2016-08-25
[conflict] Need Lua fix

Dan54:
StreamingAssets/LUA/Need.lua is now loaded


quill18:
Heya!  This seems to do a little more than advertised, including a lot of whitespace changes in seemingly unrelated places.  Could you trim this down to a more focused PR please?

Issue #631, opened 2016-08-22
Music style/feel for Project Porcupine


longtomjr:
This is a discussion thread to try and figure out what music feel and style we want to go for. I just want the PPMusicians to know what style and feel to have in mind when composing and playing.

Here is my first improv track. I just wanted to put something on tape. This is just to see if this style resonates with the team. https://soundcloud.com/longtomjr/space-in-time
It was inspired by the Bastion OST. 

Don't be afraid of anyone's feelings, at this time we are just trying to get a feel for what we want to go for.

Tranberry:
If I had a valid vote I'd vote for 2001: A Space Odyssey 1968 - Movie Soundtrack like. But that might be a bit too bombastic and odd for a game.

longtomjr:
because we are not story driven I am looking for something a bit more ambient. I will give the OST a listen and see what I think. :) Thanks for sharing!

longtomjr:
Oh and another problem is that we do not have a symphonic orchestra at our disposal unfortunately. So I am also trying to keep it simple.

Tranberry:
yeah, but as with art you can boil it down to a few notes I'm sure ;)

ajahraus:
That's a good tune, though perhaps a bit somber for ambient music. The acoustic guitar brings with it a bit of a 'space western' vibe, which could be the aesthetic we end up with, but I'm not sure we're there yet. 

Now that I'm thinking about it, perhaps there may be a way to tie the ambient music to the recent in-game events or moods of the characters? If there was just an attack and a few people died, maybe a nice somber tune is just what's called for.

ajahraus:
Would it be appropriate for us to link to songs that have the style/feeling we'd like to see emulated, even if we aren't the artist? Or is musical style too personal to crowd source like that? Maybe we need some executive orders to get us started in an appropriate direction.

SquareMan:
I dig it! I'm not sure if it would fit as ambient, anytime music, but as @ajahraus said, i'm sure it could be tied into the eventsystem and trigger when appropriate.

koosemose:
[...]

PR #781, opened 2016-08-25
[help] Added MusicController and SpaceTheme

Ferr013:
I 've commited only the .wav file and the script.
[help] Still need someone able to implement it in the game without messing up the scene as I did in my previous attempts.


MANDAL0R3:
I could be wrong about this but i think its better to wait until we can import the sounds/music from StreamingAssets.

Sommerbo:
I agree with @MANDAL0R3 use of "Resource" folder should be minimal. Music and sound should reside in /Assets/StreamingAssets/Music

Ferr013:
I moved the file from Resources to StreamingAssets. I'll leave the code fixing to someone more experienced.

quill18:
Since .WAV files are so large (due to being uncompressed), I'm *thinking* we should change it out for a compressed version instead.

Please refer to issue #795, which I just opened up, for a discussion.

quill18:
Also, you may want to join the conversation over at Music style/feel for Project Porcupine #631

Ferr013:
@quill18 Thank you for the advise! I've heard that mp3 has some problem with looping, because it add some kind of information on the head or the end of the file that interrupts for some (perceivable) istant the music when it starts over. Do you have any suggestion about the codec to adopt in this case? 
I'll look up the conversation right now!

longtomjr:
@Ferr013 Check out here #795 We will use .ogg files it seems.

Ferr013:
I'm trying to load from StreamingAssets, but it throws a null reference exception.
Can anyone help me figure it out please?

`#region License
// Project Porcupine Copyright(C) 2016 Team Porcupine
// This program comes with ABSOLUTELY NO WARRANTY; This is free software, 
// and you are welcome to redistribute it under certain conditions; See 
// file LICENSE, which is part of this source code package, for details.
#endregion

using UnityEngine;
using System.Collections;

[RequireComponent (typeof (AudioSource))]
public class MusicController : MonoBehaviour
{
    AudioSource backgroundMusic;
    public string defaultTrack = "SpaceTheme";

    void OnEnable()
    {
        PlayTrack(defaultTrack);
	}

    public void PlayTrack(string title)
    {
        backgroundMusic = GetComponent<AudioSource>();
        AudioClip ac = LoadMusic(title);
        backgroundMusic.clip = ac;
        backgroundMusic.loop = true;
        backgroundMusic.Play();
    }

    private AudioClip LoadMusic(string title)
    {
        string filePath = Application.streamingAssetsPath;
        filePath = System.IO.Path.Combine(filePath, "Music");
        filePath = System.IO.Path.Combine(filePath, title + ".ogg");

        byte[] dataAsByteArray = System.IO.File.ReadAllBytes(filePath);
        Debug.Log(dataAsByteArray.ToString());
        float[] dataAsFloatArray = ConvertByteToFloat(dataAsByteArray);

        AudioClip audioClip = new AudioClip();
        audioClip.SetData(dataAsFloatArray, 0);

        return audioClip;
    }

    private float[] ConvertByteToFloat(byte[] byteArray)
    {
        float[] floatArray = new float[byteArray.Length / 4];
        for (int i = 0; i < floatArray.Length; i++)
        {
            if (System.BitConverter.IsLittleEndian)
                System.Array.Reverse(byteArray, i * 4, 4);
            floatArray[i] = System.BitConverter.ToSingle(byteArray, i * 4) / 0x80000000;
        }
        return floatArray;
    }
}`

mattpott:
I have got it to play a noise but it is not music. Instead of using

`new AudioClip();`
use
`AudioClip.create()`
This is what i used to make sounds 

`AudioClip audioClip = AudioClip.Create(title, dataAsFloatArray.Length, 1, 44100, false);`

mattpott:
Looks like we need to use a solution something like this. http://answers.unity3d.com/questions/499028/onaudioread-buffer-changing-sizes.html

Ferr013:
@mattpott Thank you very much! Now it works fine!
I've added a prefab but not in the scene to avoid messing up things.

mattpott:
At the moment I am not getting any sound when I test this unless I add the script from with in the editor.

mattpott:
Also one potential problem with use nvorbis is that the licence is not compatible with the GPL licence we are using. https://nvorbis.codeplex.com/license

svmnotn:
@Ferr013 what is the state of this PR? Also please fix the conflicts

mattpott:
We may have to revert this back to using the unity system for loading music from the assets folder and later down the line we can move this back to streaming assets.

sboigelot:

sboigelot:
@Ferr013 What is the status of this PR on your side?

Maybe you should make a PR with just the music files in the streaming asset?

Ferr013:
Sorry for the lack of care about this project. Here is what I can do, if it can't be of any help, I'll open a PR with only the music.

sboigelot:
@Ferr013 no problem about the lack of care, we all have life ;) It would be great if you could just make a pr for the music indeed. I would do it, but it would be better with your name on it for posterity.

sboigelot:
We now have a PR allowing us to load sound files from the streaming asset. Closing this PR.

PR #783, opened 2016-08-25
Rewriting Parameter unit test to match implementation & increasing coverage

mikejbrown:
Do as I say, kids, not as I do. #746 broke one of the tests by changing the spec (such as it is) as expected, so I rewrote the test. Also wrote a new test (already passing) checking behavior of Parameters which have a value as well as a sub-parameter.

Question: should Parameters override equality operators? It appears they compare by reference, not value as I would expect... There should be test coverage either way but I've held off because there isn't really a defined behavior yet. What do people want?

I could either put a [WIP] on this if you want to discuss it here or open an Issue and let this be merged for now.

### DIFF
--- a/Assets/Editor/ParametersEditorTest.cs
+++ b/Assets/Editor/ParametersEditorTest.cs
@@ -112,6 +112,22 @@ public class ParametersEditorTest
     }
 
     [Test]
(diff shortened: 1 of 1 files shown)


quill18:
Going ahead with the merge.

PR #787, opened 2016-08-25
Lua script file loading centralized with try-catch

quill18:
This moves all the Lua file loading to a centralized function, which additionally wraps the loading in a try-catch block that outputs an informative message indicating which line of the file caused the problem.


Randshot:
This will help when debugging Lua files :)

PR #686, opened 2016-08-23
Furniture.xml Style (See #789)

aleserena:
standarized indentation and style for Furniture.xml

PR #789, opened 2016-08-25
Style update adressing problems in #686

aleserena:
*Updated to use 4/8/12/16 spaces not tab.
*Rearranged tags so every furn uses the same order.


quill18:
Merging this and closing #686 -- but for the future note that if you make changes, commit, and push to your previous branch, that will update your pre-existing PR.  You don't need to start a new one.

PR #790, opened 2016-08-25
Compressor sprite update

TheRiven:
This replaces the current Compressor sprite sheet with a new one.

compressor-animation2


quill18:
Very sexy.

PR #797, opened 2016-08-25
LocalizationDownloader refactor

QuiZr:
Someone maintaining this later will have less trouble understanding this... thing >.>


QuiZr:
Wanted to create some unit tests but this class is horrible for it, nvmind.

PR #801, opened 2016-08-25
Lua need fix

Dan54:
The game now loads ```StreamingAssets/LUA/Need.lua```


Randshot:
When I add a simple line of code like `ModUtils.ULog("Need.lua loaded")`, Unity throws a lot of exception errors.

Dan54:
Fixed

WardBenjamin:
In NeedAction.cs, you used tabs instead of spaces.

Dan54:
Fixed

Assets/Scripts/Models/World.cs
```diff
@@ -166,7 +166,7 @@ void SetupWorld(int width, int height)
                 tiles[x, y].Room = GetOutsideRoom(); // Rooms 0 is always going to be outside, and that is our default room
             }
         }
-
+        new NeedActions ();
```
Mizipzor: This doesnt look right. Why are you instantiating an object and then never use it=

Dan54: @Mizipzor The NeedActions instantiator creates a static reference to itself, keeping a reference in World is unnecessary.

Mizipzor: I think there's cleaner implementations of the singleton pattern, but that can be saved for another time. :)

Mizipzor:
Tested. Game starts without errors.

PR #802, opened 2016-08-25
Update CONTRIBUTING.md

Mizipzor:
Adding a new step to merge with master before testing and pushing to expose problems earlier.

I assume everyone test their stuff before creating a pull request. But not everyone merges/rebases with upstreams master before doing so. Githubs GUI already flags merge conflicts. But theres another kind of "soft" conflict, as in a code or game logic conflict, that goes unnoticed if you do.

You maybe removed a null check because a null should never happen, but at the same time someone else made it possible to happen. A lousy example maybe, but you get the point.

I also think this supersedes step six, rebase. I wanted to keep the commits down before when I was nagging about this, because I didnt know Githubs web GUI could rebase and squash. But it is also now about to keep the number of steps down.

This is commonly my main concern when deciding if a pull request should be merged or not.

Edit: Apparently my editor silently trimmed all trailing whitespace in the file. Oh well, consider it a bit of drive-by cleanup.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,27 +22,27 @@ If you would like to contribute to this project by modifying/adding to the progr
  * From the command line: `git remote add upstream https://github.com/TeamPorcupine/ProjectPorcupine.git`
 5. Create a branch for your new feature.
  * From the command line: `git checkout -b my-feature-branch-name`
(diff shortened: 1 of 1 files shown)


Randshot:
Looks clean xD

RodRitter:
Brilliant. Definitely needed.

PR #803, opened 2016-08-25
small lua optimization

DeShrike:
Nothing major, but every CPU cycle saved in Update() is a good thing IMHO


Sommerbo:
That looks good to me and a good saving in cpu time.

Randshot:
Just tested it and it causes no problems.
I also looked at the code and it looks very clean.

RodRitter:
I like it! Good job! Merged.

PR #807, opened 2016-08-25
Room deletion

mattpott:
There was a bug in the furniture deconstruction code that when a wall was deconstructed the surrounding walls where never deleted.

This just deletes the rooms before doing the flood fill however this won't preserve the gas calculations.


Dan54:
Why don't you make it so that you can pass `Room.DoRoomFloodFill()` a room and a size. Then, before store the room's size and unassign all tiles, then do the flood fill and finally delete the room.

Mizipzor:
@Dan54 If you have a better implementation implement it and add a pull request. :)

I'm going to consider this a follow change to the room graph.

Issue #796, opened 2016-08-25
Game slows down/freezes after playing a while


Sommerbo:
Playing the game whether in the Unity editor or as a standalone program after placing a few furniture the garbage collection (we think) goes nuts and freezes the game, pretty good sign in Unity that GC is taking place.

The length of freezes seems to be tied to the number and complexity of the furniture placed, for example 40 stockpiles creates small but very frequent freezes while a smelter and a miner creates longer less frequent freezes that leads to the game freezes for more than 2 minutes.

Some users see a increased memory usage,I didn't. I think that is just when running in Unity editor. Many saw an increased cpu usage during the freezes mine jumped from 10% to 22% during the freeze events.

We discussed this in the discord chat and one theory is that has to do with the number of foreach loops since a foreach triggers GC.

Developers involved where @Mizipzor , @NogginBops and @longtomjr

quill18:
Such large freezes are extremely unlikely to merely be GC collection issues.

NogginBops:
We also noted that we all where running Nvidia GTX 7xx series graphics cards. If anyone with another graphics card has this problem please tell.

longtomjr:
Ok. after some testing. I don't know if this is the same issue, but I get hanging if I build somewhere unreachable. I see in the debug console that a job was found, then a pause while, then the same message until the char finds another job.

Sommerbo:
I just tested with Intel 4000 gfx same issue maybe after a little longer.

mikejbrown:
Tried locating the issue with `git bisect` and got this https://gist.github.com/mikejbrown/d91607de1dbbebed74a1353bff056968. Certainly happy if someone wants to try reproduce that.

sboigelot:
Same issue for me with a GTX980m, I don't think it GC related or gfx card related. After a while the game pick up speed then slow down again. 

If it's lua related, I've added code in the quest system to not run the LUA every frame (onUpdate). Furniture really don't need to process 70 times a second, maybe 2 or 3 time a seconds is enough so we could do the same. (<400ms is not visible for most of us, un-evolved human :p)

sboigelot:
Here is the sample code for the quest (I use a check every 5 seconds here)


 ```
   private float totalDeltaTime;
    private readonly float checkDelayInSeconds;

    public ConstructorOrInitialize()
    {
        checkDelayInSeconds = 5f;
    }

    public void Update(float deltaTime)
    {
[...]

PR #809, opened 2016-08-25
Tile movement cost calculations moved to C#

quill18:
We should attempt to eliminate any case where a Lua function gets called multiple times per frame.

This is not the fix to #796 -- but it should help performance a fair bit.

### DIFF
--- a/Assets/Scripts/Models/JobQueue.cs
+++ b/Assets/Scripts/Models/JobQueue.cs
@@ -45,7 +45,7 @@ public class JobQueue
         }
 
         jobQueue.Add(j.jobPriority,j);
-        Debug.Log ("Enquued job for " + j.jobObjectType ?? "");
+        Debug.Log ("Enqueued job for " + j.jobObjectType ?? "");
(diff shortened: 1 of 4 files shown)


NogginBops:
@quill18 I think #808 is a better solution to still be able to mod the MovmentCost.

Assets/Scripts/Models/JobQueue.cs
```diff
@@ -45,7 +45,7 @@ public void Enqueue(Job j)
         }
 
         jobQueue.Add(j.jobPriority,j);
-        Debug.Log ("Enquued job for " + j.jobObjectType ?? "");
+        Debug.Log ("Enqueued job for " + j.jobObjectType ?? "");
```
mikejbrown: Thank you! I know it's a minor thing but that typo was _so_ annoying!

quill18:
Merged #808.  Still merging this for minor logging cleanup and to purge the Lua MovementCost_Standard completely lest someone link to it in their piece of furniture.

PR #810, opened 2016-08-25
Hotfix/game tick

sboigelot:
I know it may not change much to #796 but it may help in the overall

Reduce the amount of LUA call to furniture to X (5 currently) times a second by creating a GameTick

(not sure if this is a hotfix but I'll call it like that anyway for now)


vogonistic:
Is this compatible with the speed control? Bound to 1, 2 and 3 on the keyboard.

sboigelot:
Yes this still use the TimeScale variable, so speed still works

vogonistic:
Looking at your commits, it looks like you merge `upstream/master` into your `master`. Is that correct?

If so, next time you are on master *with no local changes*. Run these commands: `git fetch upstream; git reset --hard upstream/master`. It'll take your `master` and forcibly move it to the same commit as `upstream/master` and it'll look a little better on github after that.

sboigelot:
@vogonistic I just executed the git bash commands. It doesn't seems to have affected my other PR for now, but may affect my future ones. Thanks a lot for the tip, maybe we should have a explanation about it in the contribution file. I tried to solve this issue alone a few time without success. (First time I'm working on a forked repo)

vogonistic:
Yeah, it affects only the new PR's.

vogonistic:
Although, if you see this again, ping me and we'll figure it out.

Randshot:
@vogonistic Isn't fast-forwarding enough (after fetching upstream)?

vogonistic:
@Randshot It is, unless you are already in a state like this, then you can't fast-forward because you already have merged earlier.

sboigelot:
@Randshot   fast-forwarding? How do you do that?

Randshot:
@sboigelot I think something like this: `git pull -ff master`
Basically something with '-ff'.

Skalicon:
@sboigelot for future reference a hot fix is a bug fix that goes directly from the production code (usually master) and merges directly back into the production code. Since we haven't released the game yet nothing we do should be considered a hotfix, this is just a regular old bugfix :)

sboigelot:
@Skalicon oki doki! I'll keep it in mind, thanks

PR #811, opened 2016-08-25
Made a better version of MovmentCost

NogginBops:
This version does not rely on lua but is still modifiable from lua via the MovmentModifier property in tile.


Assets/StreamingAssets/LUA/Tile.lua
```diff
@@ -6,7 +6,6 @@
 -------------------------------------------------------
 
 -------------------------------- Tile Actions --------------------------------
-
 function CanBuildHere_Standard ( tile )
 	return true
```
zwrawr: Are tabs in .lua okay?

NogginBops: I don't think we have any set guide lines yet. Could be wrong though.

vogonistic: We've been accepting PR's that convert tabs to 4 spaces, so we are moving in that direction.

NogginBops: For me locally those are spaces... weird.

Assets/Scripts/Models/Tile.cs
```diff
@@ -66,6 +66,8 @@ public Furniture Furniture
 
     public int Y { get; protected set; }
 
+    public float MovmentModifier { get; set; }
```
vogonistic: `Movement` is misspelled

Assets/Scripts/Models/Tile.cs
```diff
@@ -99,6 +96,7 @@ public Tile(int x, int y)
         this.X = x;
         this.Y = y;
         Characters = new List<Character>();
+        MovmentModifier = 1;
```
vogonistic: It helps showing that it's a float if you set it to either `1f` or `1.0f`. I prefer the second, but both are good to indicate it's not an int.

Mizipzor:

NogginBops:
@Mizipzor Done! :)

Assets/Scripts/Models/Tile.cs
```diff
@@ -111,18 +112,16 @@ public float PathfindingCost
 
     public int Y { get; private set; }
 
+    public float MovmentModifier { get; set; }
```
vogonistic: This should be `MovementModifier`

kd7uiy: For that matter, why bother having {get; set;}? Either make it a protected set, or else just remove the accessors.

Assets/Scripts/Models/Tile.cs
```diff
@@ -112,7 +112,7 @@ public float PathfindingCost
 
     public int Y { get; private set; }
 
-    public float MovmentModifier { get; set; }
+    public float MovementModifier { get; set; }
```
kd7uiy: This should either be a proper accessor, with a private or protected set, or just a variable, without the {get;set}

TomMalbran: I vote for the first option.

NogginBops: @TomMalbran The reason it's public is so that it is accesable from lua. But making it a public variable is easy. Will do when I have time.

TomMalbran: A public get and private set should work to be able to read from Lua. I don't think that we need to set it from lua?

NogginBops: That was the idea, to be able to change the movmentcost of a tile on a tile by tile basis

vogonistic:
I'm wondering if instead of doing it this way we figure out how to handle alternative statues on tiles and how it should be drawn? Some statuses that might modify movementCost could be `sticky`, `slippery`, 'dirty`, `broken`

Mizipzor:
@vogonistic do you suggest closing this and looking into better alternatives? Or should we merge this in the meantime?

NogginBops:
I think this is better performance than what we have right now (Even though nothing uses custom movement costs). It's a better solution than we currently have, even if it has to go later.

Assets/Scripts/Models/Tile.cs
```diff
@@ -112,7 +112,7 @@ public float PathfindingCost
 
     public int Y { get; private set; }
 
-    public float MovementModifier { get; set; }
+    public float MovementModifier;
```
Grenkin1988: Why have you decided to change to `public` field?

NogginBops: I though that was what was decided in the discussion above. Am I wrong?

kd7uiy: It was changed in a previous PR to be public so that it can be changed by LUA scripts. The accessor verbiage was left in, which was removed to make it look cleaner, both of which I believe are the correct decisions.

vogonistic:
@Mizipzor We now know that the performance dies if it's ever used so this is in the right direction. The discussion on if a mod needs to change it and if it should be public or private is what I was thinking of.

kd7uiy:
@vogonistic My preference would be to pull this in, and then work on better methods in another PR.

vogonistic:
Merged!

PR #819, opened 2016-08-26
[Conflicted][Waiting on Resolution of #1000]Fix Loading Params from save file and save Params inconsistency

koosemose:
Parmeters are now saved in the same format they are loaded in, wrapped
in a Params tag, and empty Parameters aren't written.


Assets/Scripts/Models/Parameter.cs
```diff
@@ -28,14 +28,14 @@ public class Parameter
     private Dictionary<string, Parameter> contents;
 
     // Tracks if value has been explicitly set
-    private bool uninitializedValue = true;
+//    private bool uninitializedValue = true;
```
TomMalbran: If you are not using this value you can just remove it.

Assets/Scripts/Models/Parameter.cs
```diff
@@ -222,9 +224,16 @@ public bool HasContents()
...
+        if (string.IsNullOrEmpty(name))
+        {
+            writer.WriteStartElement("Params");
+        }
+        else
+        {
+            writer.WriteStartElement("Param");
+            writer.WriteAttributeString("name", name);
+        }
+        if (!(value == null) )
```
TomMalbran: it is more clear to use ```if (value != null)```

koosemose: dangit, one escaped me... And I'm sure I had a reason to write the null check like that... it just may have been insanity however.

Assets/Scripts/Models/Parameter.cs
```diff
@@ -233,7 +230,7 @@ public void WriteXmlParamGroup(XmlWriter writer)
             writer.WriteStartElement("Param");
             writer.WriteAttributeString("name", name);
         }
-        if (!(value == null) )
+        if ((value != null) )
```
Grenkin1988: Should be - if (value != null)

koosemose: It is now... I think you got in just as I was pushing the commit.

TomMalbran: There are double parenthesis and an extra space.

Assets/Scripts/Models/Parameter.cs
```diff
@@ -222,9 +221,16 @@ public bool HasContents()
...
+        if (string.IsNullOrEmpty(name))
+        {
+            writer.WriteStartElement("Params");
+        }
+        else
+        {
+            writer.WriteStartElement("Param");
+            writer.WriteAttributeString("name", name);
+        }
+        if ((value != null) )
```
Grenkin1988: @koosemose No this is still exist.

Assets/Scripts/Models/Parameter.cs
koosemose: I'm dumb... I was just reading it as you correcting me on what it was before... *sigh*

Assets/Scripts/Models/Parameter.cs
koosemose: What are you talking about, I can't tell from the large marked part.

TomMalbran:
Have you tried loading a save? It is not reading all the parameters and I am not sure why.

Assets/Scripts/Models/Parameter.cs
```diff
@@ -27,15 +27,12 @@ public class Parameter
...
-    // Tracks if value has been explicitly set
-    private bool uninitializedValue = true;
-
     public Parameter(string name, string value) 
     {
         this.name = name;
         this.value = value;
         contents = new Dictionary<string, Parameter>();
-        uninitializedValue = false;
+//        uninitializedValue = false;
```
TomMalbran: You should remove this too. And the one on line 45.

koosemose:
I am utterly baffled as to what's going on with reading the Params... as far as I can see it SHOULD work.

TomMalbran:
It works when reading the prototypes, but it fails when reading from the save file. My thought is that when it gets to ReadXml when reading from a save file the cursor is at the Furniture tag, instead of the Params one, as it when reading the prototypes. I tried moving the reader to the Params, by moving to the descendant, but then it started skipping the first param.

koosemose:
Yeah, that's basically the same thing I found... I may have gotten it sorted out, but Pathfinding is grinding things to a halt, atm.

koosemose:
Ah Hah! I've got it. Flipped the while() around to a do...while(), and it works just fine... of course, that being the case, I have no idea why it didn't have that issue reading from prototypes.

koosemose:
Ok, everything works now, Params are properly loaded (tested with an edited save file with a souped-up O2 Generator), and Prototypes still load properly (tested with running the "New Game" map), along with fixing the little inconsistencies in how Params saved to xml that was the original intent of this.

Assets/Scripts/Models/World.cs
```diff
@@ -1029,7 +1029,11 @@ void ReadXml_Furnitures(XmlReader reader)
                 int y = int.Parse(reader.GetAttribute("Y"));
 
                 Furniture furn = PlaceFurniture(reader.GetAttribute("objectType"), tiles[x, y], false);
-                furn.ReadXml(reader);
+                if(!(reader.IsEmptyElement) )
```
Grenkin1988: this could be if(!reader.IsEmptyElement ) (too many parentheses)

Mizipzor:
Tried saving a game, then restating and loading. Got a blank map and NullReferenceException spam in the logs. Am I wrong or you? :)

koosemose:
Lemme check, things may have gotten a little mixed up since Stylecop Saturday.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -132,9 +132,9 @@ private Furniture(Furniture other)
         UnlocalizedDescription = other.UnlocalizedDescription;
     }
 
-    public event Action<Furniture> Changed;
+    public event Action<Furniture> OnChanged;
 
-    public event Action<Furniture> Removed;
+    public event Action<Furniture> OnRemoved;
```
Grenkin1988: No not know why people think that On... is good name for event. OnChadged is good name for method attached to event.

Assets/Scripts/Models/Parameter.cs
```diff
@@ -98,14 +98,20 @@ public string Value
...
     {
+        Parameter paramGroup = new Parameter(reader.GetAttribute("name"));
         XmlReader subReader = reader.ReadSubtree();
-        Parameter paramGroup = new Parameter(subReader.GetAttribute("name"));
 
         // Advance to the first inner element. Two reads are needed to ensure we don't get stuck on containing Element, or an EndElement
         subReader.Read();
+
+        // In case the reader gets passed early, we descend to Params if it's not a Params or Param
+        if (subReader.Name != "Params" && subReader.Name != "Param")
```
Grenkin1988: This would be - .Name.Contais("Param")

koosemose:
I cannot recreate. To check my assumptions, by restarting you mean stopping playmode in the editor and then starting it again? Not fully restarting Unity, right?

Assets/Scripts/Models/Furniture.cs
koosemose: The bulk of events (13 out of 18) use On... name until and unless a different standard is discussed and agreed upon, I will be continuing to use this convention.

TomMalbran: It might be a jQuery thing where the register methods start with on, and now it feels like a good name mainly because there isn't a Register method and it is used like ```Furniture.OnRemoved += function```.

Assets/Scripts/Models/Parameter.cs
koosemose: No, it wouldn't, I am checking for precisely what I want. It is bad practice to check for something that only generally fits what you want and hope that nothing changes later that will cause false positives.

Further this makes it explicitly clear what this line is looking for.

Assets/Scripts/Models/Furniture.cs
koosemose: Ultimately, I don't care what it is in the end, I just prefer that it is consistent. At the moment I consider this an informal standard solely through majority.

I personally think that with how our events are registered this reads better. But if a different standard is decided upon then of course we change all to whatever the standard is.

vogonistic:
Waiting on #1000 to reach some kind of conclusion before changing the events to have an `On` prefix.

koosemose:
Alright, I'll wait to fix any conflicts until a consensus is reached, so I don't add another hundred conflict resolutions. :)

TomMalbran:
Could we merge this PR without the event name changes, which don't really affect this PR? Those can be fixed after we get to a convention on the names (if required).

Mizipzor:
@Yes it was just starting and restarting the game, not Unity. I can test again.

koosemose:
@Mizipzor to the best of my knowledge this is on hold until a conclusion to 1000 is reached, if that is not the case I will look into fixing any conflicts and see if I can track down any bugs that may need tracked down.

koosemose:
I am closing this as due to its age and multiple conflict resolutions, it has become a mangled mess of old and new code, and it is easier to fix the issues this solves from scratch than to sort out what code is old code staying in due to a conflict resolution and should be removed, and what code should stay. New PR is #1056

PR #820, opened 2016-08-26
[Conflicts] Added conditional debug print messages to pathfinder

vogonistic:
What do you guys think of this methodology? The methods are only compiled if `DEBUG_PRINT_ASTAR` is defined, but is not spending any time otherwise.

Not super pretty, but works.

@Mizipzor @RodRitter @WardBenjamin

### DIFF
--- a/Assets/Scripts/Pathfinding/Path_AStar.cs
+++ b/Assets/Scripts/Pathfinding/Path_AStar.cs
@@ -6,15 +6,18 @@
 // file LICENSE, which is part of this source code package, for details.
 // ====================================================
 #endregion
+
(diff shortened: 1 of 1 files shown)


Mizipzor:
Sure. Is it easy to define compile flags in Unity? Or do you do it like always directly in Visual Studio?

Sommerbo:
I think is a good way to do it.

Mizipzor:

vogonistic:
I have no idea how compilation flags works in C#, MonoDevelop or Unity in general. This was mostly an example. Since @WardBenjamin is replacing it with a better pathfinder, I'm closing this.

PR #831, opened 2016-08-26
NeedActions singleton.

Mizipzor:
A cleaner implementation of the singleton pattern, not requiring someone to explicitly call the constructor.

### DIFF
--- a/Assets/Scripts/Models/World.cs
+++ b/Assets/Scripts/Models/World.cs
@@ -167,7 +167,7 @@ public class World : IXmlSerializable
                 tiles[x, y].Room = GetOutsideRoom(); // Rooms 0 is always going to be outside, and that is our default room
             }
         }
-        new NeedActions ();
+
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Models/World.cs
```diff
@@ -167,7 +167,7 @@ void SetupWorld(int width, int height)
                 tiles[x, y].Room = GetOutsideRoom(); // Rooms 0 is always going to be outside, and that is our default room
             }
         }
-        new NeedActions ();
```
Mizipzor: ...because this just feels wrong.

NogginBops:
This should probably make use of the LuaUtilities class

Mizipzor:
@NogginBops I think theres a separate pull request for that.

NogginBops:
It might be my PR (If it is #837 you're thinking of)

PR #833, opened 2016-08-26
Added Tile.GetGasPressure function

Dan54:
Added `Tile.GetGasPressure(string gas)`. If `Tile` is in a room it returns `Tile.Room.GetGasPressure(gas)` otherwise it returns the lowest pressure from `Tile`'s neighbors that are in a room.


Randshot:
I have no problems running the code and the performance seems to be about the same.
One thing to note is that this PR has some (many) whitespace changes.
Nothing world changing but wanted to mention it anyway.

Dan54:
I realized after making my changes that I had left tabs in the code, so I replaced all tabs with spaces.

Randshot:
@Dan54 I don't mind but wanted to mention it for the Gatekeepers.
They will have to decide what to do :)

Raketfart:
I've read this through, and it seems good. Haven't tested it though.

Assets/Scripts/Models/Tile.cs
```diff
@@ -316,6 +316,40 @@ public Tile West()
...
+    public float GetGasPressure(string gas)
+    {
+        if (Room == null)
+        {
+            float pressure = Mathf.Infinity;
+            if (North().Room != null && North().GetGasPressure(gas) < pressure)
+            {
+                pressure = North().GetGasPressure (gas);
+            }
+            if (East().Room != null && East().GetGasPressure(gas) < pressure)
```
quill18: I might be missing something, but shouldn't this be an `else if`?

Raketfart: No, it takes the lowest gas pressure from all adjacent tiles

Mizipzor:

Mizipzor:
Looks good. :) The gas and pressure systems are a bit in a state of flux at the moment, I'm going to be lenient with pull requests for it.

PR #837, opened 2016-08-26
Changed Action classes to use LuaUtilities

NogginBops:
I made this change to remove a lot of code duplication that existed in e.g. FurnitureActions and NeedActions.

The only drawback I think should be considered is that if every lua file that gets loaded is in the same script, the search space for global functions could get pretty big.


TomMalbran:
Maybe we could make LuaUtilities into a BaseAction class and have the other Action clases inherit from it. That way we could have a different environment for each file.

NogginBops:
@TomMalbran good idea! I likey. I'm really not thinking inheritance in this project as it was built from the ground up without much inheritance, so I'm missing all these smart solutions... :P Though we don't really know if global space bloating is an issue.

TomMalbran:
@NogginBops No idea about global space bloating, but I like having several smaller files with concise functions than huge ones that do way too much. We can use namepsaces to avoid class names clashes or big names and we can also organize the files in folders. All the Actions classes can be moved to an Actions folder inside Models.

Mizipzor:

NogginBops:
@Mizipzor Done! :)

TomMalbran:
@NogginBops I thought that you were going to try to move into a more inheritance approach so that not all the functions share the same global space?

Are the Needs functions loaded? It looks like you removed the loading part in the latest merge.

NogginBops:
@TomMalbran Yeah I think the inheritance idea is good but I think that this is a good stepping stone. Added the LoadsScript functions back in but it was never actually called inside the code so I added a static constructor to do that automatically when it's actually going to be used.

TomMalbran:
I would like to eventually have something similar to what I did with the PrototypeManager, some sort of ActionsManager.

LoadScripts was called when creating the instance, and the instance was created when a method called the Instance property. Check how the FurnitureActions is done. The constructor is created inside WorldController and the ModsManager calls the mods loading function.

When we make the current xml/lua/images files be part of a default mod, my idea was that the ModManager would be in charge of reading the files and call all the required loaders passing them the file content.

koosemose:
Error on Launch that prevents proper startup.

```
NullReferenceException: Object reference not set to an instance of an object
NeedActions..cctor () (at Assets/Scripts/Models/NeedActions.cs:20)
Rethrow as TypeInitializationException: An exception was thrown by the type initializer for NeedActions
ModsManager.LoadPrototypes () (at Assets/Scripts/Controllers/ModsManager.cs:33)
ModsManager..ctor (System.String dataPath) (at Assets/Scripts/Controllers/ModsManager.cs:22)
WorldController.OnEnable () (at Assets/Scripts/Controllers/WorldController.cs:91)
```

Assets/Scripts/Models/NeedActions.cs
```diff
@@ -13,43 +13,18 @@
...
-        myLuaScript.Globals["Inventory"] = typeof(Inventory);
-        myLuaScript.Globals["Job"] = typeof(Job);
-        myLuaScript.Globals["ModUtils"] = typeof(ModUtils);
-
-        // Also to access statics/globals
-        myLuaScript.Globals["World"] = typeof(World);
-
-        LoadScripts();
+        LoadScript();
+        LoadModsScripts(WorldController.Instance.modsManager.GetMods());
```
koosemose: LoadModScripts causes the NRE, mod handling is already handled through ModManager.

koosemose:
This looks like a good first step to cleaning up the Action classes, and once the NRE is fixed, it runs just fine. 

Though we really need to follow this up with possibly getting each object type to have it's own LuaScript, or some other way to separate the namespace of different script types. Additionally EventAction really needs to be generalized so all the Action classes can use it, rather than always calling FurnitureActions.CallFunctionsWithFurniture, and having its Trigger method require a furniture to be passed in.

vogonistic:
@NogginBops Can you fix the NRE and test this? I'm getting a NRE on each Update() for a missing mouse controller.

NRE 1:
```
NullReferenceException: Object reference not set to an instance of an object
NeedActions..cctor () (at Assets/Scripts/Models/NeedActions.cs:19)
Rethrow as TypeInitializationException: An exception was thrown by the type initializer for NeedActions
ModsManager.LoadPrototypes () (at Assets/Scripts/Controllers/ModsManager.cs:34)
ModsManager..ctor (System.String dataPath) (at Assets/Scripts/Controllers/ModsManager.cs:22)
WorldController.OnEnable () (at Assets/Scripts/Controllers/WorldController.cs:91)
```

NRE 2:
```
NullReferenceException: Object reference not set to an instance of an object
WorldController.Update () (at Assets/Scripts/Controllers/WorldController.cs:142)
```

NogginBops:
@vogonistic Fixed!

vogonistic:
Merged!

PR #834, opened 2016-08-26
BaseSpriteController

relinu:
All spriteControllers now derive from BaseSpriteController.
Adds RemoveAll, this lets you delete the visual representation of SpriteControllers

PR #838, opened 2016-08-26
[StyleCop warnings][Conflicts] Added WorldFactory, creates Worlds in code.

### LINKED ISSUE #828: Create a world through code
Right now some of our code makes difficult assumptions. Like that there ever would be only one world which will be created once and live until the game shut down. It also relies on the scene being set up with some game objects and controllers already in place.

I think this needs to change for several reasons. I'm gonna pick up #459 and to do that I need to solve this first. Any help appreciated.

Mizipzor:
Very useful for tests.
Arguably how it should be handled in the main scene as well. 
There is however not yet a clean way to destroy a world.

Added a `Tests` folder. The scene is supposed to be as empty as possible. I allowed `Main Camera`, `EventSystem` and `Canvas` as they are not made by us. Other than that there a single `GameObject` with a `WorldFactoryBehaviour` which wraps `WorldFactory`.

There are some things i havent solved yet, like how to give a behavior a certain prefab.

Next step is to cleanly take down a world, which should result in the same empty scene.

Fixes #828.

### DIFF
--- a/Assets/Scripts/Controllers/InventorySpriteController.cs
+++ b/Assets/Scripts/Controllers/InventorySpriteController.cs
@@ -70,7 +70,7 @@ public class InventorySpriteController
         }
         sr.sortingLayerName = "Inventory";
 
-        if (inv.maxStackSize > 1)
+        if (inv.maxStackSize > 1 && inventoryUIPrefab)
         {
             // This is a stackable object, so let's add a InventoryUI component
             // (Which is text that shows the current stackSize.)
--- a/Assets/Scripts/WorldFactory.cs
+++ b/Assets/Scripts/WorldFactory.cs
@@ -1,0 +1,33 @@
+﻿using ProjectPorcupine.Localization;
+using System;
+using UnityEngine;
+
+public class WorldFactory
+{
+    private GameObject rootController;
+
+    public void Create()
+    {
+        if (rootController != null)
+        {
+            throw new InvalidOperationException("Creating more than one world is currently not supported.");
+        }
+
+        Debug.Log("Creating world.");
+
+        rootController = new GameObject("Controllers");
+
+        GameObject worldControllerGameObject = new GameObject("WorldController", typeof(WorldController));
+        worldControllerGameObject.transform.parent = rootController.transform.parent;
+
+        rootController.AddComponent<SpriteManager>();
+        rootController.AddComponent<GameEventManager>();
+        rootController.AddComponent<LocalizationLoader>();
+    }
+
+    public void Destroy()
+    {
+        // TODO Theres currently no way to cleanly take down a world.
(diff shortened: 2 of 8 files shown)


relinu:
#834 adds removeAll to SpriteControllers which lets you delete the sprites

Mizipzor:
@relinu Awesome, thanks.

Sommerbo:
@Mizipzor Consider using UberLogger instead of unity's log. And class name "WorldFactory" is epic!

Mizipzor:
@Sommerbo Doesnt UberLogger hook into and replace the regular log calls? Or have I already forgotten how it works?

Sommerbo:
@Mizipzor No you must use `Debug.ULog` or `Debug.ULogChannel`, itis  better that way since UberLogger truncates multiline log messages.
UberLogger also passes all log messages to Unity console so you can choose several ways to view messages. Personally I like UberLoggers channel system, makes it easy to concentrate on one issue

Mizipzor:
I stand corrected. :)

vogonistic:
You have StyleCop errors, can be merged when those are fixed.

bjubes:
@vogonistic I checked this PR out, for me it doesn't have any violations even after merging in master....?

vogonistic:
@bjubes Really? Unless we've changed the rules the last 3 days I think you need to check your StyleCop setup. I get these StyleCop warnings:

WorldFactory.cs:1 Missing header
WorldFactory.cs:2 System using directives before other using directives
WorldFactoryBehavior.cs:1 Missing header
WorldFactoryBehavior.cs:9 Method must have access modifier

svmnotn:
can confirm the StyleCop issues that @vogonistic pointed out, @Mizipzor please fix :)

Mizipzor:
I struggle to find time for this project. What do you think of this? Should I continue on it or merge as it is (after fixed conflicts)? Or let someone else pick it up?

sboigelot:
I'm sorry but I don't see much in this PR for now. It's not link to anything else in the code and doesn't do much. If you don't have time, I think it will be better to close it. When someone else will be able to tackle the issue, that person will be more free in his/her implementation.

So it's up to you: if you have time to continue, let us know, if not, close it ;)

kd7uiy:
I'm not quite sure what this is actually supposed to do either. I like the idea of having a test scene, but I don't think this quite does the trick...

TomMalbran:
I don't even think that this is the best solution for #828. Instead of creating world controllers, we should create Worlds, and have a WorldManager that manages them. We should then change all the Controllers to be able to work when we switch from one world to another. Everything in a controller that keeps state of something specific to the current world should be moved to another class that is part of the world manager.

sboigelot:
this is a very old PR with no update and the author doesn't have time for now, I'm closing it for now. Feel free to re-create it if you update it!

PR #793, opened 2016-08-25
Power system initial prototyping with tests (reopened)

Grenkin1988:
This is reopen of #761

On next PR I am going to implement:
1. Parse `Furniture` and enchance it with `PowerRelated`
2. Change `PowerSystem` with `LivePowerSystem`

Firsts part of work for #602

PR #841, opened 2016-08-26
Moved Power system classes to namespace Power and rename of classes

Grenkin1988:
Fixing suggesions from @vogonistic and @NogginBops  made on #793


Randshot:
Works without errors for me :)
I only scrolled through the code and from what I have seen, it looks pretty clean.

Randshot:
Since he only added summaries, it obviously still works.

Grenkin1988:
Fixed conflicts

PR #846, opened 2016-08-26
Switching to UberLogger. 

Sommerbo:
Switched to UberLogger. First part of my pass over "Models" folder
Added missing License header.
Removed old commented out debug calls.
Removed a few debug calls that where nearly useless.
Used the "Lua" channels for a few calls that where linked to Lua functions.

NB. I was unable to test/trigger the change in `Assets/Scripts/Models/Inventory.cs` but it should work and in any case UberLogger passes all debug messages to Unity's console.


WardBenjamin:
FYI you still have the original commit that was already merged, here. Shouldn't affect anything hopefully.

Sommerbo:
It should be fine. Not touching the same files.

PR #848, opened 2016-08-26
Prototypes manager

TomMalbran:
I moved all the prototypes from the world and moved them into its own classes. This will be useful when having multiple worlds, as they can all share the same prototypes.

Right now each prototype class is loading the files, but when we change how mods work, the file loading will be done by the mod manager, and it will just pass the file contents to the prototypes classes for parsing.

There are still more prototypes outside of the world (like the tile types) that should be moved. But for now I think that this is already a big change.


Assets/Scripts/Models/Prototypes/BasePrototypes.cs
```diff
@@ -0,0 +1,101 @@
...
+    public Dictionary<string, T>.KeyCollection Keys
+    {
+        get
+        {
+            return prototypes.Keys;
+        }
+    }
+
+    /// <summary>
+    /// Returns the prototype Keys.
```
Grenkin1988: Values :-)

TomMalbran: Woops. Not sure if that was me, or Mono that failed. hehe

NogginBops:
This! 🙏

Mizipzor:

TomMalbran:
Conflicts fixed. This time it was harder.

Fatalexceptionz:
Seems to do what it says on the tin -- it runs, and my quick skim through the code didn't turn up anything glaringly obvious, though I'm getting sleep deprived, so...
Looks a lot cleaner and self-descriptive than it was.
BuildModeController could do with a cleanup next time someone touches it.
Could maybe be generalised even more to cut down on the boilerplate code, but, eh..

TomMalbran:
@Fatalexceptionz I agree. I don't like how the BuildMode / Mose controllers combo works. They both depend on each other. I would like it to work in a way where the BuildModeController would ask the MouseController to call it back if something happens, so that the mouse controller wouldn't need to call it manually. And similar with a new SelectionModeController. But does seems like a big refactor that should be done on its own.

vogonistic:
There are merge conflicts. Also remember to check if there are StyleCop warnings before removing `[conflict]` from the title.

TomMalbran:
Conflicts fixed

vogonistic:
Merged!

PR #852, opened 2016-08-26
Fixed StyleCop issues of Tile class

Grenkin1988:


Mizipzor:
It has begun.

These are difficult to merge so I'm taking it right away. Please test that stuff works one extra time. :)

PR #857, opened 2016-08-26
StyleCop-ed Character.cs

Randshot:


Randshot:
It runs on my machine now.

PR #865, opened 2016-08-26
StyleCop-ed Settings.cs

itay347:


vogonistic:
This PR conflicted and we ended up merging #928 in it's place.

PR #946, opened 2016-08-27
Fix Dev Mode leaving phantom jobs in BMC - Fixes #945

### LINKED ISSUE #945: [Bug] Dev mode blocks furniture building
When in dev mode, furniture and floors are built instantly. But a job is still created, just not added to the job queue. This means that no furniture can be built on insta-created floors because the BMC thinks there is still pending jobs there.

Am I the only one experiencing this?

koosemose:


MANDAL0R3:
Awesome :D

PR #949, opened 2016-08-27
Live power system: fixed bugs and it is should be ready to use

Grenkin1988:
Next step would be use this system.
Add to `World` and `Furniture`

### DIFF
--- a/Assets/Scripts/Models/Power/Grid.cs
+++ b/Assets/Scripts/Models/Power/Grid.cs
@@ -80,7 +80,7 @@ namespace Power
             connections.Remove(connection);
         }
 
-        public void Update()
+        public void Tick()
         {
             float currentPowerLevel = 0.0f;
             foreach (Connection connection in connections)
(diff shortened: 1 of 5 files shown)


Assets/Scripts/Models/Power/Grid.cs
```diff
@@ -118,28 +118,45 @@ private void ChargeAccumulators(ref float currentPowerLevel)
...
 
         private void DischargeAccumulators(ref float currentPowerLevel)
         {
             if (currentPowerLevel +
-                connections.Where(connection => connection.IsPowerAccumulator && !connection.IsEmpty).Sum(powerRelated => powerRelated.OutputRate) < 0)
+                connections.Where(connection => connection.IsPowerAccumulator && !connection.IsEmpty)
+                .Sum(connection =>
+                connection.AccumulatedPower - connection.OutputRate < 0.0f ?
+                connection.AccumulatedPower :
+                connection.OutputRate) < 0)
```
Mizipzor: Ive been staring at those four lines for a while now (144-147), something needs to happen but Im not sure what. The expression is very complicated. Could we indent 144 with one tab and 145, 146, 147 with two tabs? Or, consider breaking out the logic into methods so that you just do `.Sum(MyMethod);`.

Mizipzor:
Nice! :)

PR #951, opened 2016-08-27
Ladder sprite transparency

abackwood:
Makes a minor change to the Ladder sprite to make the open part transparent. The original sprite stems from a time where empty tiles had their own sprite and it looked best for laders to be superimposed over that as if there was an empty tile below them. Now with spacey backgrounds, this can go back to being transparent


Mizipzor:
I like your reasoning. :)

PR #953, opened 2016-08-27
Put JobPriority enum into Job class

Randshot:
I put the JobPriority enum back into the Job class to make it accessible from Lua without having to define global replacements for them in the Lua files.
For this, I had to rename the JobPriority property to Priority.

**_I do think that this is an acceptable name for the property._**
**_If you think otherwise, then please explain why and maybe how to solve it differently._**


mikejbrown:
That's how I named it in my PR which was superseded, so yes, I for one :+1: the name.

Mizipzor:
Tested. Cant find anything wrong.

TomMalbran:
Great. It will be nice to do something similar to the Enterability.

PR #955, opened 2016-08-27
Fixed pause behaviour - Fixes #853

### LINKED ISSUE #853: Pausing bug
When you pause the game all of the characters within view of the camera pause but some outside of the camera view do not. Also the animations seem to keep playing, please let me know if someone else has the same problem.

Fatalexceptionz:
Added IsPaused check on call to World.UpdateCharacters()
Added comments denoting places for different update frequencies.


mikejbrown:
tests fine for me :+1:

PR #957, opened 2016-08-27
StyleCop-ed the newly changed files

Dormanil:
Because apparently we just happily continue to break our coding standard.


Mizipzor:
Comments superfluous.

PR #960, opened 2016-08-27
Fixing sub menu toggle

vringar:
The old version set set all submenus to false just to then toggle the submenu it was clicked on.
The results were invisible or just a short blinking of the submenu.
My version doesnt scale very well but it does the job for now.
Would like to hear better ideas.


Mizipzor:
Love those commits messages. Why do you say this doesnt scale?

vringar:
because you have to add every submenu off the main menu seperatly

Mizipzor:
Good enough for me, we solve scaling later when we have to.

bjubes:
Did u actually see flickering? I though deactivating and activating in the same frame does nothing because those calls set a flag and unity actual hides it after the function is finished running.

vringar:
im not sure if this was just my pc being slow af or the programm, i have about 10 fps with the game

bjubes:
Okay as long as it actually happened and your not just assuming. I thought it wouldn't change on and off in one frame.

PR #961, opened 2016-08-27
Created a work priority tab.

dustinwloring1988:
Created a work panel.like in RimWorld where you can get job priorities.
It will display the Character's name.
The priorities you set don't affect anything yet and don't save.

Control + Left Click - Set To Min
Control + Right Click  - Set To Max
Left Click - Goes Down By One
Right Click  - Goes Up By One

### DIFF
--- a/Assets/Resources/Prefab/WorkerPrefab.prefab
+++ b/Assets/Resources/Prefab/WorkerPrefab.prefab
@@ -1,0 +1,1133 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!1001 &100100000
+Prefab:
+  m_ObjectHideFlags: 1
+  serializedVersion: 2
+  m_Modification:
+    m_TransformParent: {fileID: 0}
+    m_Modifications: []
+    m_RemovedComponents: []
+  m_ParentPrefab: {fileID: 0}
+  m_RootGameObject: {fileID: 1000012504224464}
+  m_IsPrefabParent: 1
+--- !u!1 &1000010003625826
+GameObject:
+  m_ObjectHideFlags: 0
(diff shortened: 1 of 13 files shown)


Dormanil:
Looking good, except it seems like the localization system somehow expects that the `4` and the `X` should have translations. 
snippet

Anyway... how about `right-click` to reduce priority? Also `Ctrl+left-click` to set priority to maximum and similarly `Ctrl+right-click` for priority to minimum?

Dormanil:
Also, characters don't get automatically updated. If a new character spawned or an old one left, you will still see the same menu items.

dustinwloring1988:
@Dormanil will do once I get home, Hum I get the list from World.current.charaters.count should I get it from some were else

Dormanil:
Where you get that list from is probably fine, you just have to update your work panel every time that list changes.

dustinwloring1988:
I do everytime the tab is opened it updates is there a chance that the charesters are not add right

Dormanil:
Oh, I saw it updating every time it is opened, that is not the issue. The issue is: You had the tab open for a longer period of time and one of the characters died. You didn't notice that, and the tab doesn't update either, so when you try to change the priorities of that character, `NullReferenceException`s could be seen in the console.

dustinwloring1988:
o ok will look at that also. Thanks for the help.

dustinwloring1988:
@Dormanil - Fixed Thanks for testing it.

You can now do the following;
Control + Left Click - Set To Min
Control + Right Click  - Set To Max
Left Click - Goes Down By One
Right Click  - Goes Up By One

NogginBops:
@dustinwloring1988 What happened to this? Did it get merged or was it dropped?

PR #962, opened 2016-08-27
[Conflicts][WIP] Fire! [requested contribution]

FilippoLeon:
Ever desired to burn down your spaceship? From now you can use fire!

fire

What is there:
 - fire spreads and destroys
 - integrated with the temperature system
 - need a "Post Update hook" to let LUA delete/place furniture
 - fire consumes oxygen and produces CO2
 - fire dies if pressure too low
 - randomly starts when mostly inconvenient
 - conveniently buildable by your spacers

What is not there (and you can help and fix):
 - the "fire spread" sound is the same as placing furniture
 - there is no "fire crackling" sound
 - some error is thrown if you try to spread to invalid position
 - sprite is horrible
 - who knows?


sboigelot:
You should add in the todo list: fire will die in absence of oxygen.

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -373,6 +373,33 @@
         <UnlocalizedDescription>furn_oxygen_compressor_desc</UnlocalizedDescription>
 
     </Furniture>
+  
+  
+    <Furniture objectType="fire" skipBuildMenu="true">
+        <Name>Fire</Name>
+        <Description>Stores and releases oxygen as needed</Description>
```
sboigelot: You may want to update the description ;)

FilippoLeon: :P

FilippoLeon:
Yes, good idea.

Tritaris:
There is the event system that can be used to randomly have a fire start. It was added in PR #270. This would be a good way to have the fires start. I would have it when a machine malfunctions it has a chance to burst into flames. You can get a list of the different machines installing into the world from `World.current.furnitures`.

Assets/StreamingAssets/Data/Furniture.xml
Mizipzor: Yes, this needs fixing before merge.

FilippoLeon:
Updated!

bjubes:
not necessary for this PR, but flammability as a property on furniture and tiles would be a very cool addition

sboigelot:
@FilippoLeon What your status on this? Do you still work on it or should someone take it over?

sboigelot:
This was a great start, but the OP didn't give sign of life since 21 days. Closing this PR for now. @FilippoLeon feel free to open it again if you come back!

PR #963, opened 2016-08-27
Make job selectable

herrflockig:
This makes the pending job in a tile selectable


Mizipzor:
Tested. Queued up a heater and it can be selected.

PR #966, opened 2016-08-28
[dep #971] graphical changes

Yurgeta:
updated some of the graphics, stockpile, walls, floor, redone o2 generator


Yurgeta:
not sure whats gone wrong here.. im new to this so any assistant to fix the conflict would be great..

Fatalexceptionz:
I resolved the conflict and put a PR through to @Yurgeta's branch.
Looks good to merge now.

abackwood:
- I like the airlock, but think the West-East one should be moved up a bit to leave some space on the bottom, placing it in the middle of the wall, like it is when North-South. Now it looks like it is right at the bottom.
- I'm not a big fan of the new O2 generator. It looks too much like a compressor. If you could add some element that really sells it as a generator that would be better IMO
- I had some doubts about the walls but they look good  in the screenshots and I like the cleaner look of them
- The rim of the stockpile should hug the outside of the tile a little more IMO, now it takes up too much of the tile. Are the stripes supposed to be paint on the floor marking the area? In that case you could also go with making the grey stripes transparent so you just have the white markings. Obviously the circles in the middle aren't very pretty but that'll be resolved when the neighbouring tile ting is improved :smile:

Mizipzor:
This is a tricky one. I appreciate your efforts, but we cant really have people run around replacing each others sprites. There would be some serious toe stepping. But there's also no clear distinction for what is place holder and what isn't.

@TheRiven You made the original oxygen machine, right? What do you think?

longtomjr:
I think we need to discuss this issue a bit more. See #971. I think some of the shading is not correct with the rest of the project so we will need to also reach some consensus on that.

It looks good, I really lke the airlock. We still need art for the airlock so if you can just PR the airlock it will be great @Yurgeta :)

vogonistic:
Waiting for some kind of feedback from the #971 discussion

Ajapeu:
@Yurgeta Might be good to come join the art channel on the discord and chat with folks to resolve any conflicts!

abackwood:
@Yurgeta Please resubmit the airlock art as its own PR. Also come talk on Discord or Issues about the stockpile art.
The whole package as is won't be accepted.

vogonistic:
Looking forward to the airlock PR. Closing this one.

Issue #197, opened 2016-08-18
What should we do for new game art?


Spool5520:
How should new game art be handled? I would like to work on some but am unsure what the style should be or how it should be decided on what gets put in and what doesn't. Any ideas? I have a style in mind but I'm sure there are other ideas.

UnknownSoldier1:
I'd be for either something more or less realistic or a pixelated look. A pixelated look would be a lot easier to make for non artists too.

Tranberry:
**On style this are my 2 cents:**
 Go for a simple to imitate style, my _personal experience_ with FOSS games is that when an 'ace' graphic designer/illustrator/whatever appears and his or her work gets adopted - the project either goes bananas in style (mixing the very poor with the very good) or it only gets some of the required art that is needed.

Myself I'm a pretty decent graphic dude but I don't have the time to mimic 'ace' style art for free, so a simpler but coherent style is my vote.

GamerInspired:
I think to work out what style we want for the game would be to make a few different varitions on a couple of things that are already in the game. This will help everyone get an idea what things might look like instead of just suggesting this or that (not that that is wrong or anything).

TheRiven:
I tend to favour the pixel look, mostly because its something I can just about work in :grinning:
But I would like to see what people can do before we settle on anything, we may find something that looks just right.

Tranberry:
@TheRiven got an example of your pixel style? There are many different styles under that umbrella. 

If you want to see my quickest style it would be like this #102 .

Spool5520:
Okay I like the ideas. I draw up some stuff when I get home and see what people think

Sommerbo:
Good pixelart is still hard to get right. But some decision needs to be made. Personally I would prefer pixelated art

TheRiven:
@Tranberry heh when I say I can just about work in it, I mean I can do a quick bit of pixel programmer art that serves as a half-reasonable example of the object :smile: 
But if you want an example you can look at my quick attempt at a power generator, as I threw that together so I had something to use.

Spool5520:
I personally think it's more important to have perspective be consistent more than the actual art itself for instance rimworld is all over the place with their perspective. I had in mind the classic Pokemon perspective I think it's called 3/4 view and I agree on the pixel art I think it would be easier to maintain consistency with pixel art. (Zelda Pokemon ect style pixel art)

Spool5520:
[...]

PR #969, opened 2016-08-28
[Un-conflicted!] Major changes to Sprite Management

Dormanil:
Allows for diagonally merging furniture and easier writing of the sprite's XML file.

I'd like some help getting the characters completely working again before this one can be merged in.
It looks as if the character's sprites are offset on some of the animation frames and they also seem to be using the wrong frames for their animation altogether sometimes. I have absolutely no clue as to why that happens, so if **YOU** can help me, I'd be much obliged.


Tranberry:
@Raketfart maybe you should see this?

Raketfart:
Hmmmm.... It seems like the xml has the character being 2 * pixelPerUnit high. So every y position should be even, or the character will be cut in half. So fixing y on sprites will probably fix this.

That said, I don't agree with having size and positions this way. It would not work if we wanted to combine all graphics in a huge spritesheet. It also would be harder to make spritesheets sizes in a power of two (512/1024/2048). I think we need to be able to define width, height and position independently of pixelPerUnit.

Dormanil:
Alright, at this point I can't say it is WIP any longer, because I fixed all the issues I knew existed, but maybe you surprise me and find something that I don't know about. 

The stockpile graphics were beautifully crafted by @Yurgeta and I know of one current issue with them, but these errors have to be dealt with artistically, and I am only good for coding, not for art. 

@Raketfart All these issues you mentioned are all induced because people keep submitting art that is either not in sizes of powers of 2 or not confirming to our tile-based-ness (i. e. sprites that are of sizes between two tiles worth). This makes it easier for sprite artists to write the XML files for their furniture, since they now can count tiles instead of having to deal with pesky pixels. Writing x="1", y="2" is way easier than x="256", y="512" -- even with these easy numbers.

Raketfart:
But if we imagine, at some point combining sprites together to reduce drawcalls. Then a sprite in 100 resolution has to be placed at 100 or 200 or 300. Another sprite with reolution 70 would be placed at 70, 140 or 210. So combining spritesheets would be a terrible puzzle with lots of empty space.

Dormanil:
As I say again, if we would actually decide on one common pixel per unit base, we would never have to deal with this issue.

A sprite at 32 resolution would still be placed at 32, 64, 96 and so on, and a sprite at 64 resolution (because common base) would be placed at 64, 128, 192 and so on. Differing to your example, we can fit 32 resolution sprites snugly in those places where 64's can go to.

Raketfart:
Yes, if we decide on a resolution base like that, it would be much better. We just have to account for a pixel spacing between sprites, then it wouldn't be a problem.
It seems like most sprites actually have those sizes anyway. 
Then I no longer have a problem with this. :)

Tranberry:
@longtomjr this is important ^

longtomjr:
Please look at this discussion about the Pixel Per Unit #197

Mizipzor:
Whats the reason for all the dead space in ` Assets/StreamingAssets/Images/Character/char1_spacesuit.png`?

Dormanil:
To have it be 70x140 pixel sprites, in other words 2 tiles high. Sprites should never *ever* be inbetween two tile sizes.

Raketfart:
I think the character is a fine example of a sprite that can be bigger than the fixed tile size.  You could also imagine the back of a chair sticking out of the top of a tile. 
To limit the artwork in this way would probably be bad.

Tranberry:
@Raketfart we are drawing and talking about this in `#art` on dsicord.

Mizipzor:
Surely a character can be any size as long as we correctly align his feet with whatever tile he is on?

Or, rather, no matter how much empty space above his head you add, the character is still about 1.5 tiles high.

longtomjr:
I think the Sprite size should stay the same, But we should just have empty space for sprites that take up more than one tile.

Dormanil:
I honestly rather have a character sprite be 64x128 and then not have to deal with odd-ball sizes ever again, but that is not my call.

longtomjr:
@Dormanil agreed. Even if the sprite has empty space at the top.

longtomjr:
OK, This has been standardized, see
https://github.com/TeamPorcupine/ProjectPorcupine/wiki/Standards-and-Conventions#pixels-per-unit

Raketfart:
Yes, I think we all agree that this is how we do it

Dormanil:
Since we are in a state of agreement, I think we are ready now.

Mizipzor:
So, uh, as someone that struggles that get a grip on the (many) changes, should this just be merged?

Assets/Scripts/Controllers/FurnitureSpriteController.cs
```diff
@@ -115,19 +113,40 @@ public Sprite GetSpriteForFurniture(Furniture furn)
...
-        spriteName += GetSuffixForNeighbour(furn, x, y - 1, "S");
-        spriteName += GetSuffixForNeighbour(furn, x - 1, y, "W");
+        string suffix = string.Empty;
 
-        // For example, if this object has all four neighbours of
+        suffix += GetSuffixForNeighbour(furn, x, y + 1, "N");
+        suffix += GetSuffixForNeighbour(furn, x + 1, y, "E");
+        suffix += GetSuffixForNeighbour(furn, x, y - 1, "S");
+        suffix += GetSuffixForNeighbour(furn, x - 1, y, "W");
+        if (suffix.Contains("N") && suffix.Contains("E"))
```
TomMalbran: I think that we can replace this calls with something like:
```
suffix += GetSuffixForNeighbour(furn, x + 1, y + 1, "ne");
suffix += GetSuffixForNeighbour(furn, x + 1, y - 1, "se");
suffix += GetSuffixForNeighbour(furn, x - 1, y - 1, "sw");
suffix += GetSuffixForNeighbour(furn, x - 1, y + 1, "nw");
```

vogonistic:
This PR has conflicts.

Assets/Scripts/Controllers/FurnitureSpriteController.cs
Dormanil: We can't. Because it would just make no sense to have things merge diagonally when there is no neighbour in the cardinal directions next to it.

TomMalbran: Right, maybe we could make a function?
```
private string GetSuffixForDiagonalNeighbour(string suffix, string coord1, string coord2, string Furniture furn, int x, int y)
{
    if (suffix.Contains(coord1) && suffix.Contains(coord2))
    {
        return GetSuffixForNeighbour(furn, x + 1, y + 1, coord1.ToLower() + coord2.ToLower());
    }

    return string.Empty;
}
```
Just to remove some repeated code.

Assets/Scripts/Controllers/FurnitureSpriteController.cs
```diff
@@ -73,18 +71,43 @@ public Sprite GetSpriteForFurniture(Furniture furn)
...
+        if (suffix.Contains("N") && suffix.Contains("W"))
+        {
+            suffix += GetSuffixForNeighbour(furn, x - 1, y + 1, "nw");
+        }
+
+        // For example, if this object has all eight neighbours of
         // the same type, then the string will look like:
-        //       Wall_NESW
+        //       Wall_NESWneseswnw
         return SpriteManager.current.GetSprite("Furniture", spriteName);
```
TomMalbran: This should be ```SpriteManager.current.GetSprite("Furniture", spriteName + suffix);```

Dormanil: Knew that I missed something. Thanks!

Assets/Scripts/Controllers/FurnitureSpriteController.cs
Dormanil: We still need to check the correct tiles though. If we did your function there, it would always check the tile to the northeast if there's a neighbour.

TomMalbran: Right the code has a copy-paste bug. should be:
```
private string GetSuffixForDiagonalNeighbour(string suffix, string coord1, string coord2, string Furniture furn, int x, int y)
{
    if (suffix.Contains(coord1) && suffix.Contains(coord2))
    {
        return GetSuffixForNeighbour(furn, x, y, coord1.ToLower() + coord2.ToLower());
    }

    return string.Empty;
}
```

Dormanil: I don't think you realise how `GetSuffixForNeighbour` works: It checks the tile at the given xy-Coordinates to see if it has exactly that furniture. If I did this code now, I'd be checking for the tile the furniture asking if it has neighbours is on, i. e. the same furniture I hand the method over as `furn`. Believe me, this would not work.

TomMalbran: I tested the code and it works. Here is a gist with the 3 functions: https://gist.github.com/TomMalbran/f43ba15b0b8b29f8af034f600707b378

GetSuffixForDiagonalNeighbour basically does what you did. It first check if there are neighbours at the given coords, and then if there is a neighbour at the given position. If you replace the variables in the code, you get the if that you created.

Dormanil: Okay, I am dumb and see now how it would work. Thanks for that, @TomMalbran. Do you want to PR it into my fork and get credited or shall I just do it myself?

TomMalbran: You can do it yourself. I don't mind.

Dormanil:
@Mizipzor, this only has so many changes, because I did not want to break every furniture's XML file to suddenly not work any more. The longer this is in limbo and the more changes to existing furniture or furniture additions are made, the bigger this will get.

Mizipzor:
After a quick vote in #art this should be merged, as it will only be more difficult to merge going forward, Though I would assume that this break some sprites in one way or another, you art people will have to band together for this one I think. :)

Dormanil:
And as I said, on #art in discord, I will try to help people moving other to this new format.

Mizipzor:
Minor compile error, will merge when fixed.

Dormanil:
Alright, merged in latest code from upstream and fixed what auto-merge borked up.

PR #976, opened 2016-08-28
Allow airlock doors to be pressure locked

koosemose:
Airlock doors can now be pressure locked (only open when the pressure is equal on both sides, accurate to 3 decimals).

This currently doesn't do much as there is no way to change the pressure in a room other than to open doors, or otherwise put a hole in a wall, but is the first step to allowing airlock doors to be used to make a proper airlock.

Default is for them to not be pressure locked (toggled from context menu) so as to not create any difficulties in using airlock doors before some form of pressure pump is added. For now airlock doors will continue to behave as before, unless the pressure lock is toggled on, then in most cases they just won't open (since there is no way to change the pressure in a room).

To test, a hole can be made in an outside wall, to dump all the atmosphere, then a pressure locked door will be able to open.

Additionally access to System.Math.Round (through ModUtils) was needed to allow rounding to a given decimal place (so as to avoid multiple steps to achieve the same effect. If a Gatekeeper would prefer not to use System.Math, I can change it to use Mathf.Round, and more steps, or Lua's Floor and multiple steps.

### DIFF
--- a/Assets/Scripts/Utilities/ModUtils.cs
+++ b/Assets/Scripts/Utilities/ModUtils.cs
@@ -26,6 +26,11 @@ public static class ModUtils
         return Mathf.FloorToInt(value);
     }
 
+    public static float Round(float value, int digits)
+    {
+        return Round(value, digits);
+    }
+
     public static void Log(object obj) 
     {
         Debug.Log(obj);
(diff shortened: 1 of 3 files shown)


vogonistic:
I'll merge this PR as is, but I'll also file an issue with improvements I think this needs to be fully useful.

Grenkin1988:
@koosemose Lock works, but people try to path trought it all the time. And just stand near it

PR #979, opened 2016-08-28
[small] Add null tile check to SpawnInventory

mikejbrown:
It's possible to generate `NullReferenceException`s by clicking off the map in dev spawn mode. This just fixes that.


vogonistic:
Merged!

PR #980, opened 2016-08-28
Moved most debug calls to UberLogger. 

Sommerbo:
Fixed a few typos
Not all `Debug.Log` where replaced due to UberLogger inability to show multiline messages.
SC passed.
And I screwed up commit message.

### DIFF
--- a/Assets/Scripts/Models/Temperature.cs
+++ b/Assets/Scripts/Models/Temperature.cs
@@ -134,7 +134,7 @@ public class Temperature
     /// </summary>
     /// <param name="x">X coordinates.</param>
     /// <param name="y">Y coordinates.</param>
-    /// <param name="temp">Temeprature to set at x,y.</param>
+    /// <param name="temp">Temperature to set at x,y.</param>
     public void SetTemperature(int x, int y, float temp)
     {
         if (IsWithinTemperatureBounds(temp))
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Models/Temperature.cs
```diff
@@ -212,7 +212,8 @@ public bool IsWithinTemperatureBounds(float temp)
         }
         else
         {
-            Debug.LogWarning(string.Format("Yep, something is wrong with your temperature: {0}.", temp));
+            // string.format not needed with UberLogger.
```
mikejbrown: Did not know that. Cool. :smile:

mikejbrown:
I didn't know you could remove `string.Format`, but beyond that I don't see what could possibly be breaking about the diff. I haven't generated the messages that used `string.Format` to check that, but it compiles and runs. :+1:

Assets/Scripts/Models/Temperature.cs
Mizipzor: While correct, this comment shouldnt be committed. :)

Assets/Scripts/Models/Temperature.cs
```diff
@@ -228,9 +229,9 @@ public bool IsWithinThermalDiffusivityBounds(float thermal_diff)
         {
             return true;
         }
-        else
+        else
```
Mizipzor: Trailing whitespace.

Sommerbo:
Fixed the whitespace

PR #984, opened 2016-08-28
Moved most debug calls to UberLogger. 

Sommerbo:
Grouped MouseOver* messages in one channel since they had similar function and nearly identical code.
Many files but small changes in each


vogonistic:
Merged!

PR #986, opened 2016-08-28
Fix character-generation overloads

WardBenjamin:
Characters now have random colors by default and overloads utilize each other so that there's no code duplication. This also fixes characters that were being generated without a color not having a random name, since that function just calls the other overload with a random color.


WardBenjamin:
@Mizipzor I blame you for merge conflicts :P

Mizipzor:
Pff, not me, I dont write any code, I just merge.

PR #992, opened 2016-08-28
Refactor FulfillableInventoryRequirements onto Job

### LINKED ISSUE #991: Each Job should have a "Can I complete this task given the current world state" function
Don't know what to call it, but it would be very helpful. For instance, is there enough steel in the world to do this job? If not, skip that task and do the next one. If so, drop the "global" count and take the job. I would guess that World would have to know anytime a resource is created or consumed, but there might be a better way. This would probably help to some extend with stockpiling as well.

mikejbrown:
Fixes #991 
StyleCop passed

Is now a public instance method. Call like `job.FulfillableInventoryRequirements()`.

### DIFF
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -617,7 +617,7 @@ public class Character : IXmlSerializable, ISelectable, IContextActionProvider
         }
         else
         {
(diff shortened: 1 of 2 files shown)


vogonistic:
Merged!

PR #994, opened 2016-08-28
[Conflict] World Tab

dustinwloring1988:
Here is a new world tab this way when the user clicks on world in the bottom menu it will do something.

I also add a name to the world it will just be Space followed by a number from 1 to 100 just something for now.


vogonistic:
This PR has a merge conflict. Remove `[conflict]` from the title when it's fixed. Don't forget to check for StyleCop errors.

sboigelot:
This is a very old PR with no update, I'll close it for now. @dustinwloring1988 feel free to re-create it if you update the branch ;)

PR #995, opened 2016-08-28
Adds Unit Test to Duplicate Key Comparer

zwrawr:
three tests fail , currently think its  the DuplicateKeyComparer being broken rather than the the unit test code being wrong. Would like someone to take a look and see if they agree

### DIFF
--- a/Assets/Editor/DuplicateKeyComparerTest.cs
+++ b/Assets/Editor/DuplicateKeyComparerTest.cs
@@ -14,7 +14,7 @@ public class DuplicateKeyComparerTest
     /* 
      * ==================================
      * Compare an object against it's self
-     * Objects should be considered duplicates
+     * Objects should be considered duplicates and so return value will be based of equalValueAtEnd
      * ==================================
      */
 
@@ -25,7 +25,7 @@ public class DuplicateKeyComparerTest
         int test_int = 1;
 
         DuplicateKeyComparer<int> dkc_int = new DuplicateKeyComparer<int>();
-        Assert.AreEqual(1, dkc_int.Compare(test_int, test_int));
+        Assert.Greater(dkc_int.Compare(test_int, test_int), 0);
     }
 
     [Test]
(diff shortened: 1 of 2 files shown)


kd7uiy:
Which tests are failing?

zwrawr:
these three

Compare_AgainstLargerValue_equalValueAtEnd_True

Compare_AgainstSmallerValue
Compare_AgainstSmallerValue_equalValueAtEnd_False

kd7uiy:
The value of a comparison isn't guaranteed to be 1 or -1, just positive or negative, or 0 if identical. You should really be checking to see if it is greater than or less than 0.

zwrawr:
I'll update the Asserts with that in mind , but i don't think that is the issue as it reports them returning either 1 or -1

> Compare_AgainstLargerValue_equalValueAtEnd_True [0.027s]
Expected: not -1
But was:  -1
at DuplicateKeyComparerTest.Compare_AgainstLargerValue_equalValueAtEnd_True () [0x0003e] in \ProjectPorcupine\Assets\Editor\DuplicateKeyComparerTest.cs:171

> Compare_AgainstSmallerValue [0.02s]
Expected: not 1
  But was:  1
at DuplicateKeyComparerTest.Compare_AgainstSmallerValue () [0x0003e] in \ProjectPorcupine\Assets\Editor\DuplicateKeyComparerTest.cs:190

>Compare_AgainstSmallerValue_equalValueAtEnd_False [0.022s]
Expected: not 1
  But was:  1
at DuplicateKeyComparerTest.Compare_AgainstSmallerValue_equalValueAtEnd_False () [0x0003e] in \ProjectPorcupine\Assets\Editor\DuplicateKeyComparerTest.cs:202

kd7uiy:
The True/False should only affect something in the event of an exact tie. Otherwise, the answer should be the same. So I think that's what the difference is.

kd7uiy:
The idea, BTW, behind the true/false is, if something is equal, do you want it placed at the beginning or the end of the list? I added that in from the one I originally used because it seemed better to stick an old task in at the end of the list, while the default behavior was to stick it in at the beginning.

zwrawr:
What would you expect those three methods to return ?

kd7uiy:
The second larger at the end should be negative, the first larger should be positive.

zwrawr:
Updated to fix issues , all tests pass

kd7uiy:
Awesome, then let's add it!

bjubes:
Tested this PR and it works. no style cop errors. all tests pass and check correct behavior. 
:+1:

PR #997, opened 2016-08-28
Implement PathfindingWeights

koosemose:
This is an extremely basic Pathfinding Weight for Tiles (which calls Furniture for it's pathfinding weight if it has furniture). At the moment the only thing used is movementCost, modifiers to the final weight (Accessed through the property PathfindingWeight on Tile), are set at 1. Pathfinding can get the total PathfindingCost (PathfindingWeight * movementCost) from the PathfindingCost Property.

The purpose is to get an extremely simple thing in place that allows those working on pathfinding to make use of a weight that can be further modified from movementCost, that is small enough to allow it to easily be merged. This will allow them to make use of it as soon as possible, and any further changes to how weight is calculated will be handled separately and should minimize conflicts with pathfinding PRs. PathfindingWeight (the multiplier to movementCost) can be modified through the property.

The system has been hooked into the current Pathfinding implementation, with no changes to behavior (specific weight for furnitures and tiles still need implemented, but this will allow future Pathfinding changes to not have to worry about what changes may be made to pathfinding weights)

At a minimum, in the future this will allow Tiles and Furnitures to have a pathfindingWeight read from xml, that will allow a separate modifier (so some tiles can be less desirable without being slower moving, such as a fire, or a door).

Further changes will depend on how costly they are calculation-wise when accessed at a high rate by the pathfinding system.

### DIFF
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -147,11 +147,24 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider, 
     public event Action<IPowerRelated> PowerValueChanged;
 
     // The effective pathfinding cost of pathing over this furniture
+    public float PathfindingCost
+    {
+        get
+        {
+            return PathfindingWeight * MovementCost;
+        }
+    }
+
     public float PathfindingWeight
     {
         get
         {
-            return pathfindingWeight * MovementCost;
+            return pathfindingWeight;
+        }
+
+        set
+        {
+            pathfindingWeight = value;
         }
     }
 
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -24,6 +24,9 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider, 
     // Prevent construction too close to the world's edge
     private const int MinEdgeDistance = 5;
 
+    // Base cost of pathfinding over this furniture, movement cost will modify the effective value
+    private float pathfindingWeight = 1f;
+
     // If the job causes some kind of object to be spawned, where will it appear?
     private Vector2 jobSpawnSpotOffset = Vector2.zero;
 
@@ -142,6 +145,15 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider, 
     public event Action<Furniture> Removed;
 
     public event Action<IPowerRelated> PowerValueChanged;
+
+    // The effective pathfinding cost of pathing over this furniture
+    public float PathfindingWeight
+    {
+        get
+        {
+            return pathfindingWeight * MovementCost;
+        }
+    }
 
     public Color Tint { get; private set; }
 
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -151,32 +151,22 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider, 
 
     public event Action<IPowerRelated> PowerValueChanged;
 
-    // PathfindingModifier is added to PathfindingCost to get the final amount
+    /// <summary>
+    /// Gets or sets the Furniture's pathfinding modifier which is added into the Tile's final PathfindingCost.
+    /// </summary>
     public float PathfindingModifier
     {
-        get
-        {
-            return pathfindingWeight;
-        }
-
-        set
-        {
-            pathfindingWeight = value;
-        }
+        get { return pathfindingWeight; }
+        set { pathfindingWeight = value; }
     }
 
-    // PathfindingWeight is multiplied by movement cost to get total PathfindingCost
+    /// <summary>
+    /// Gets or sets the Furniture's pathfinding weight which is multiplied into the Tile's final PathfindingCost.
+    /// </summary>
     public float PathfindingWeight
     {
-        get
-        {
-            return pathfindingWeight;
-        }
-
-        set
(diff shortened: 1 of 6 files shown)


vogonistic:
Should it be a a public parameter so that modders can change it for things that might change the value?

vogonistic:
I would also prefer if it's hooked up to the pathfinding graph generation so that we don't add code that is never called/used.

koosemose:
I'm not sure it should be public, as you would presumably be setting the base but getting it gives you the movement cost factored in. It doesn't seem proper that get and set would be nonsymmetrical.

Perhaps I should have 2 properties. A PathfindingWeight, which sets and gets the base, and PathfindingCost, which is what pathfinding would want.

vogonistic:
That makes sense. 👍

koosemose:
Here I go into pathfinding... if I don't return, it's all your fault.

koosemose:
And now PathfindingWeight is a get/set property, and and PathfindingCost is a get property, which is the one pathfinding will want to use.

koosemose:
Tile.PathfindingCost, now directly performs the calculations with numbers supplied from Furniture, and allows a value to be added (rather than multiplied) to the final cost. Should allow furnitures to simply change values only when something changes, rather than  performing potentially complex checks to return values depending on their current status. Tile.PathfindingCost still accessed in the same way.

TomMalbran:
Should we add the PathfindingWeight/PathfindingModifier to the Furnitures.xml and use it for doors?

koosemose:
Yes, that is the intention. I wanted to get the baseline implementation in as lightweight as possible so it could be merged and available for those working on Pathfinding as soon as possible. Further implementation (such as letting it be set through Xml) can come through a later PR.

TomMalbran:
Yes, I thought so, I had a similar idea to solve it when watching Quill's videos to solve it. I would still try to add the door part, since it wouldn't require many other changes, and the implementation would be complete, and others who need it can start using it and have at least one furniture as an example on how to use it.

koosemose:
For me personally I usually take a fair bit of time working through xml reading. I didn't want to delay things waiting on my own implementation.

koosemose:
Hmm... good point...

Assets/Scripts/Models/Furniture.cs
```diff
@@ -143,6 +151,34 @@ private Furniture(Furniture other)
 
     public event Action<IPowerRelated> PowerValueChanged;
 
+    // PathfindingModifier is added to PathfindingCost to get the final amount
```
Mizipzor: Change these to documentation comments and they should show up in code completion.
```
/// <summary>
///  PathfindingModifier is added to PathfindingCost to get the final amount.
/// </summary>
```

Assets/Scripts/Models/Furniture.cs
```diff
@@ -143,6 +151,34 @@ private Furniture(Furniture other)
...
+        get
+        {
+            return pathfindingWeight;
+        }
+
+        set
+        {
+            pathfindingWeight = value;
+        }
+    }
```
Mizipzor: Probably best to put these on one line each. Didnt StyleCop say that?
 ```
   public float PathfindingModifier
     {
         get {  return pathfindingWeight;  }
         set { pathfindingWeight = value;  }
     }
```

File: Assets/Scripts/Models/Furniture.cs [hunk 3]
```diff
@@ -404,11 +440,19 @@ public void ReadXmlPrototype(XmlReader readerParent)
             case "Description":
                 reader.Read();
                 description = reader.ReadContentAsString();
-                break;
+                    break;
```
Mizipzor: This indentation is a typo.

Assets/Scripts/Models/TileType.cs
```diff
@@ -62,6 +68,32 @@ public static TileType Floor
...
+        {
+            return pathfindingModifier;
+        }
+
+        set
+        {
+            pathfindingModifier = value;
+        }
+    }
+
```
Mizipzor: Oneline these as well.

Mizipzor:
Looks good, some very minor style issues. Ill merge when fixed.

Theres some duplication going on Between Tile and Furniture, but that can be refactored later.

koosemose:
The only duplication between them (that I'm aware of) is both having PathfindingWeight and PathfindingModifier, which is intentional, both have the potential to modify the final PathfindingCost. If it's something else, I can look at it while I'm working on the styling issues..

Assets/Scripts/Models/Furniture.cs
koosemose: Stylecop is perfectly fine with them being on separate lines. After having to make so many properties multiline on Stylecop Saturday, I just got in the habit of making all Properties like this to start with. But as long as Stylecop doesn't complain about one-lining them, I'll do it.

Assets/Scripts/Models/Furniture.cs
koosemose: Yeah, for some bizarre reason, sometimes when I paste into a line (such as pasting something that starts with a Carriage Return onto the end of a line) it will indent it another level.

Assets/Scripts/Models/Furniture.cs
koosemose: While I was aware of documentation comments, I was not aware they would show up in code completion, that is incredibly useful, especially in a project where people may end up having to use functions, properties, etc... that they're not familiar with.

koosemose:
I believe that should cover all of your styling and documentation concerns, if I missed any, let me know.

PR #998, opened 2016-08-28
Scheduler

### LINKED ISSUE #829: Scheduler
Right now it seems that cooldown logic is implemented in multiple places and it is possible, for instance, for Lua functions to get called hundreds of times a second with an enormous overhead (e.g. #796). Would people like to have a generic scheduler? Something like `GameEvent` (possibly extending `GameEvent`) only with the ability to be used by C# events as well.

Basically I'm talking about an object, probably owned by WorldController, which can register arbitrary C# and Lua events which can fire on a one shot, repeated `n` times or endlessly on an adjustable cooldown. It would track the cooldown individually for each event, handling time warp as well. It would also provide a central place to implement a policy on how often Lua calls could be scheduled, perhaps eventually even adjusting them adaptively based on performance. Maybe even some of the job code could be refactored into this (especially furniture jobs).

Thoughts?

mikejbrown:
Fixes #829

This is a generic scheduler implementation for timed events (i.e. any events not directly generated by user interaction). So far it isn't being used by anything yet except for gratuitous log-spamming debugging, but the framework is there ready to be used! Keen for feedback.

Outline
--------
* `ScheduledEvent` represents an event which can be fired on a repeated basis with a cooldown by the scheduler. It encapsulates a C# delegate `Action<ScheduledEvent>` called `OnFire`, which is automatically called as needed by its `Update` method. `ScheduledEvent` binds the delegate to a `string Name` and can be built through the prototype system or directly through a C# API. `ScheduledEvent` knows whether it is backed by a C# or Lua function (accessible through the `public EventType EventType { get; protected set; }` property), and binds the Lua function to a C# delegate when initialized from a prototype.
* `Scheduler` maintains the list of active `ScheduledEvent`s, dispatching them as appropriate through the `Update` method. It also has some convenience methods like `TimeToNextEvent` and is able to serialize its state to xml. The scheduler section of the save file looks like:

```
<Scheduler>
  <Event name="ping_log_lua" cooldown="2" timeToWait="1" repeatsForever="False" repeatsLeft="1" />
  <!-- ... more events ... -->
</Scheduler>
```
Event prototypes for Lua backed events are read by the `EventPrototypes : XmlPrototypes<ScheduledEvent>` class from the file `Events.xml` which looks like
```
<?xml version="1.0" encoding="UTF-8" ?>
<Events>
    <Event name="ping_log_lua" onFire="ping_log_lua"/> 
</Events>
```
`name` is the value used internally to track the prototypes. `onFire` is the name of the Lua function that gets called when the event fires. This function gets passed the `ScheduledEvent` object.

C# backed event prototypes must be added in C# code using `PrototypeManager.Event.Add()`.

The `ping_log` and `ping_log_lua` events are just there for testing purposes.

Feature Roadmap
===========

`ScheduledEvent` class
-----------------------
- [x] Logic for cooldowns and arbitrary number of event repeats
- [x] C# action handling
- [x] Lua action handling
- [x] Serialization save
- [x] and load (should be handled here?? `Scheduler` is the class that naturally knows about `EventPrototype`s, not `ScheduledEvent`... **being handled by `Scheduler`**)
- [x] Fix handling of large deltaTimes in update (e.g. if deltaTime >= 2 * cooldown do what? Fire multiple times?? **Yes**.)
- [x] `ScheduledEvent`s can be built from prototypes
- [x] Prototypes are managed by the new `Prototype` system

`Scheduler` class
-----------------
- [x] Register and unregister `ScheduledEvent`s
- [x] Handle dispatch to registered `ScheduledEvent`s
- [x] Read Lua events prototypes from XML file (now handled by the prototype system)
- [x] API for registering C# event prototypes (now handled by the prototype system)
- [x] Easy scheduling API for registered prototypes
- [x] Hook into the main `Update`
- [ ] Better optimizing auto-purging the event list (it currently happens every `Update`)? (_could probably remove [WIP] before this is done?_)
- [x] Serialization save and load
- [ ] Port of all cooldown/waiting/timed events in the game to the new system (_could probably remove [WIP] before this is done?_)

Anything missing above?

If anyone finds any issues or has any fixes I am willing to accept PRs on my branch https://github.com/mikejbrown/ProjectPorcupine/tree/feature/scheduler. Same terms as Contributing.md, though I strongly prefer test driven development (TDD) as much as possible. Every feature will absolutely have to have test coverage and pass StyleCop before [WIP] is removed. :smile:


Assets/Editor/ScheduledEventTest.cs
```diff
@@ -0,0 +1,182 @@
...
+        ScheduledEvent evt = new ScheduledEvent(
+            "test",
+            (e) => Debug.ULogChannel("ScheduledEventTest", "Event {0} fired", e.Name),
+            3.0f,
+            true,
+            1);
+
+        ScheduledEvent evt2 = new ScheduledEvent(evt);
+
+        Assert.That(evt.Name, Is.EqualTo("test"));
```
Grenkin1988: Why nobody uses:
```
Assert.IsNotNull()
Assert.AreEqual()
Assert.IsTrue()
```
and so on instead?

Assets/Scripts/Scheduler/ScheduledEvent.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// <param name="deltaTime">Delta time in seconds (note: game time, not real time).</param>
+        public void Update(float deltaTime)
+        {
+            this.TimeToWait -= deltaTime;
+
+            while (this.TimeToWait <= 0)
+            {
+                Fire();
+                this.TimeToWait += this.Cooldown;
+            }
```
henninglive: Does it make sense to fire the event multiple times per tick?

mikejbrown: The idea is to handle the case where deltaTime is large compared to the Cooldown of the event, so the event is scheduled to happen multiple times in the tick. The code above just makes sure the event fires when scheduled. This will probably be such a rare case as to not matter in practice unless there is a lag spike or an out of control mod (for example), in which case there are bigger problems going on than event scheduling.

So this is just handling an unlikely corner case. I could see an argument for ignoring the issue, and if people really want that logic gone I'll get rid of it.

Assets/Scripts/Scheduler/Scheduler.cs
```diff
@@ -0,0 +1,241 @@
...
+                if (instance == null)
+                {
+                    instance = new Scheduler();
+                }
+
+                return instance;
+            }
+        }
+
+        public List<ScheduledEvent> Events { get; protected set; }
```
henninglive: This allows for external modification of the event list. Is this intended?

mikejbrown: Hmm... not really. I suppose it should return a copy.

Grenkin1988: Copy would be - 
```
get
{
   return new ReadOnlyCollection(Events);
}
```

henninglive: If you return a copy, then you can't use the Events property internally in the Scheduler class when you modify the list.

Grenkin1988: If you do not care that anybody can add more to this List, than this is OK
If not - 
```
private List<ScheduledEvent> events;
public List<ScheduledEvent> Events 
{
   get
   {
      return new ReadOnlyCollection(events);
   }
}
```

Assets/Scripts/Scheduler/Scheduler.cs
```diff
@@ -0,0 +1,241 @@
...
+        /// <param name="evt">Event to deregister.</param>
+        public void DeregisterEvent(ScheduledEvent evt)
+        {
+            if (Events != null)
+            {
+                Events.Remove(evt);
+            }
+        }
+
+        public void PurgeEventList()
```
Grenkin1988: Rename to `RemoveFinishedEvents` or `ClearFinishedEvents`
By name `PurgeEventList` I decided that you are going to clear the whole list

Assets/Scripts/Scheduler/Scheduler.cs
```diff
@@ -0,0 +1,241 @@
...
+
+        public void Update(float deltaTime)
+        {
+            if (Events == null || Events.Count == 0)
+            {
+                return;
+            }
+
+            // Events may try to modify the event list, so cannot use a foreach on Events!
+            ScheduledEvent[] eventArray = Events.ToArray();
```
henninglive: Instead of modifying the Event list directly in `RegisterEvent()` and `DeregisterEvent()`. You could queue up the changes to the list, then apply them here, before the loop. This way you can avoid coping the Event list on every tick.

Assets/Scripts/Scheduler/Scheduler.cs
henninglive: Removal of finished events could be done by main loop in `update()`, on the next tick. This way you avoid the extra iteration.

Assets/Scripts/Scheduler/Scheduler.cs
mikejbrown: How is the latest push in your judgement?

Assets/Scripts/Scheduler/Scheduler.cs
```diff
@@ -44,7 +46,18 @@ public static Scheduler Current
...
 
-        public List<ScheduledEvent> Events { get; protected set; }
+        public ReadOnlyCollection<ScheduledEvent> Events
+        {
+            get
+            {
+                return new ReadOnlyCollection<ScheduledEvent>(events);
+            }
+
+            protected set
```
Grenkin1988: You think you will need subclass? And it is going to set this?

mikejbrown:
Now `Scheduler` defers the addition & removal of events from the events list until the next call of `Update()`. The `DeregisterEvent` method is now just a wrapper for `ScheduledEvent.Stop()`.

mikejbrown:
I believe this is all the basic functionality in place. The rest is just cleaning up and converting existing subsystems to use the new scheduler.

I hand rolled the xml handling to get it going, but I would like to make it fit better with the new `PrototypeManager` system for future. I tried that in what seemed (to me) the obvious way, but it went rather catastrophically pear shaped. So I'll tackle it later when I'm fresh, or someone who is more familiar with the prototype manager could take it on if they want.

With that caveat I believe it is now ready for review. Thoughts?

TomMalbran:
First of all you dont really need a class for ```EventPrototype``` a Scheduled Event Prototype is a ScheduledEvent, that will be used to create, multiple ScheduledEvents with the same data.

To use the PrototypeManager, you can then create a copy of any prototypes class, (i.e. FurniturePrototypes), you change the name of the class, replace ```<Furniture>``` for the prototype classname, the constructor base parameters (filename, global tag, individual tag (i.e: "Events.xml", "Events", "Event")), and the LoadPrototype. In LoadPrototype you read the xml (a that point the reader is at ```<Event>```) and create the prototype.

You can later use all the Prototypes using something like ```PrototypeManager.ScheduledEvent.Values``` to get all the values or ```PrototypeManager.ScheduledEvent.Keys``` to get the keys and ```PrototypeManager.ScheduledEvent.GetPrototype(key)``` to get the prototype.

I was checking at the code to see how it ca be done, but you are making a Lua call while creating the prototype and the idea of the prototype manager is that it is only xml data and doesn't know anything about Lua (those could even be loaded after). You can save the lua function name, and then call it when creating the Scheduled event.

mikejbrown:
@TomMalbran Thanks. That is almost exactly what I tried (basically the same except for deferring the binding of the lua functions). But something must have been wrong with the load order for the lua and and I ended up getting dozens of NullReferenceExceptions per frame that I just couldn't track down. :confused: I'll have another go when I'm fresh and defer the lua binding as you suggested. :smile:

TomMalbran:
The PrototypeManager is created when the WorldController is enabled, and your scheduler is created on demand, so the PrototypeManager is created before the Lua is read. I would just save the lua function name in the prototype and call the function when creating an event, or you can even modify the prototypes b setting the result of the lua function after you load the lua file.

mikejbrown:
It could still use a bit of refactoring but it is now using the new `PrototypeManager` system to handle xml loading. It is a little bit tricky to manage the load order but this appears to be working.

mikejbrown:
Think it might be ready to remove the [WIP] tag. Thoughts?

Assets/Scripts/Scheduler/EventPrototypes.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Xml;
+using Scheduler;
+
+public class EventPrototypes : XmlPrototypes<ScheduledEvent>
+{
+    public EventPrototypes() : base("Events.xml", "Events", "Event")
+    {
+    }
+
+    public void Add(string name, ScheduledEvent eventPrototype)
```
TomMalbran: ```SetPrototype``` does something similar to this function. If you don't want it to replace an item that already exists, or give a warning, I would adding an ```AddPrototype``` function to ```BasePrototypes``` so that it can be used by any class, or change ```SetPrototype```.

Assets/Scripts/Models/Prototypes/PrototypeManager.cs
```diff
@@ -20,6 +21,7 @@ public PrototypeManager()
         Need = new NeedPrototypes();
         Trader = new TraderPrototypes();
         Quest = new QuestPrototypes();
+        Event = new EventPrototypes();
```
TomMalbran: Should we call this SchedulerEvent and SchedulerEventPrototypes, since there are other events (GameEvents) and there can be more in the future.

Assets/Scripts/Scheduler/EventPrototypes.cs
```diff
@@ -0,0 +1,43 @@
+﻿#region License
```
TomMalbran: For a better organization, I think that is better to move this file to Models/Prototypes.

TomMalbran:
The PrototypeManager looks good, besides those comments.

mikejbrown:
@TomMalbran I believe that should address everything. Thanks for all the help. :smile:

TomMalbran:
@mikejbrown np, it looks good. The only thing is if the new ```Add``` function should be called that or ```AddPrototype```. But I am thinking that maybe we should rename the other functions to ```Has, Get, Set```, since you use the ```PrototypeManager```, it should be clear that you are working with a prototype. But that is for a different PR.

mikejbrown:
@TomMalbran Yes, I agree with you (shortening the other names), but I do like to keep the incidental effects of my PRs down. :smile:

Assets/Editor/ScheduledEventTest.cs
bjubes: IMO I think `Assert.That(x, Is.EqualTo(y));` reads just like English, so its easier to understand at a glance

Assets/Scripts/Controllers/WorldController.cs
```diff
@@ -244,6 +253,13 @@ private void CreateEmptyWorld()
 
         // Center the Camera
         Camera.main.transform.position = new Vector3(World.Width / 2, World.Height / 2, Camera.main.transform.position.z);
+
+        // FIXME: Get rid of this when done testing the scheduler!!!
+        // This is a Lua event which outputs a log message every two seconds on an infinite loop.
+        Scheduler.Scheduler.Current.ScheduleEvent("ping_log_lua", 2.0f, true, 0);
```
vogonistic: This should probably be removed before I merge. It's a bit spammy.

vogonistic:
No StyleCop warnings and tests well, but I don't want to merge something that outputs a log every 2 seconds.

mikejbrown:
@vogonistic Of course... done.

vogonistic:
Merged!

Assets/Editor/SchedulerEditorTest.cs
```diff
@@ -0,0 +1,311 @@
...
+{
+    private Scheduler.Scheduler scheduler;
+    private Action<ScheduledEvent> callback;
+
+    [SetUp]
+    public void Init()
+    {
+        if (PrototypeManager.SchedulerEvent == null)
+        {
+            new PrototypeManager();
```
Grenkin1988: This is very very very bad idea to have tests linked directly to file system. ( `FurniturePrototypes` )

vogonistic:
@mikejbrown Can you create a PR that resolves the problem @Grenkin1988 reported? We can't use `PrototypeManager` in our tests because it starts loading data from the `StreamingAssets` directory, which isn't available when running unit tests outside of Unity. Test shouldn't depend on file system or external resources.

Assets/Editor/SchedulerEditorTest.cs
```diff
@@ -0,0 +1,311 @@
...
+            "test",
+            (evt) => scheduler.DeregisterEvent(evt1),
+            5.0f,
+            false,
+            1);
+
+        // event which counts the events in the list at 7s
+        ScheduledEvent evt3 = new ScheduledEvent(
+            "test",
+            (evt) => (numEventsInList = scheduler.Events.Count),
```
Grenkin1988: this must be this ->  `(evt) => numEventsInList = scheduler.Events.Count,`
Without brackets

mikejbrown:
@vogonistic @Grenkin1988 see #1132

Issue #956, opened 2016-08-27
[WIP] UI Redesign


alexfriesen:
I started working on a new UI Design.

I am aiming for a slim sci-fi look.

This is the current state of the main menu. what do you think?

![pp_menu_closed](https://cloud.githubusercontent.com/assets/1307706/18029190/ae61c508-6c91-11e6-805c-97c5e2b42a3e.jpg)
![pp_menu_open](https://cloud.githubusercontent.com/assets/1307706/18029189/ae60f7b8-6c91-11e6-9879-9d296a8b8294.jpg)

vogonistic:
That is pretty slick!

Tranberry:
looks good, but how big are these buttons?

alexfriesen:
The current menu will be 90px high and the width depends on items in the menu ;)
A single button is about 98x75px.

Yurgeta:
this was something i was going to suggest / mock up in the near future because the amount of text buttons is growing fast its starting to clutter the screen a little so using picture representation for things now could save us some trouble later on.. maybe add a tool tip to the bottom so it gives the name above or something so people know what it is if they cant figure out the pictures

Mizipzor:
Doesnt it kinda clash with the pixel art? Its so smooth. Hard to tell.

Ajapeu:
@Mizipzor It might conflict a bit. That said... desperately needed, and getting to a more organized slick UI is really going to help when new features start coming in. Probably needs to happen now anyway. 

@alexfriesen This looks really nice.

Tranberry:
IMO that's pretty huge on any screen with less than 1080p resolution. It would be nice if it could scale..

alexfriesen:
Thanks for the feedback. 
I made the borders thicker and less transparent. 

With the thicker lines I would be possible to scale the menu down.

Here the reworked bottom menu
![pp_menu_main_closed](https://cloud.githubusercontent.com/assets/1307706/18034330/43f04786-6d3a-11e6-8911-f08a661f4a69.jpeg)
![pp_menu_main_open](https://cloud.githubusercontent.com/assets/1307706/18034329/43eced66-6d3a-11e6-9580-d0f0013432b6.jpeg)

The Options menu: 
![pp_menu_options](https://cloud.githubusercontent.com/assets/1307706/18034335/6bfd17ea-6d3a-11e6-87d3-e1ed6065d46e.jpeg)

And the build/furniture selection: 
[...]

PR #999, opened 2016-08-28
[Conflict] Ui redesign

alexfriesen:
I made some progress with the Options Dialog like explained in #956 

Here a preview 
pulllllllllllll


TomMalbran:
I like it. But I think that adding some transparency to the background color, might make it look better.

alexfriesen:
like that?
pull_trans

TomMalbran:
Yes. i like that better than before.

Tranberry:
looks good, as long as you don't go crazy with the transparency

Mizipzor:
That transparency has a huge impact on the looks, I love it. Normally you have to strike the tricky balance, you want the transparency effect to be clearly visible, but not so much that it makes text difficult to read. We dont have that problem because we framed our buttons. I think you can go harder on the background but keep the button a bit more opaque. Does that make sense? Maybe youre already doing that, hard to tell with the color.

alexfriesen:
In summary: 
 - settings are moved from the bottom menu to the options menu
 - the debug buttons are removed from the options menu (they didnt work anyway)

After I resolved the merge conflicts this might be ready to be merged.

alexfriesen:
PROGRESS!
pp_menu

sboigelot:
I love it, but be careful not to make too much change in this PR. At one point, decide to stop so we can merge it and start a new one.

Also, conflict ;)

TomMalbran:
Looking good so far. I would make the construction menu title look like the options one, or change both so that they look the same.

If we are going to merge this partial design which would be great and then continue with more changes, I think that we should still try to keep a category (Furniture/Floors) / buildables structure in construction for a better organization.

PR #1002, opened 2016-08-28
[Conflicts] Better Deconstruction

### LINKED ISSUE #180: Better Deconstruction
The idea is you deconstruct (or mine) stuff and get some (if not most) of the materials back. 
This might be great for Quills tutorial series.

Dan54:
fixes #180 

Deconstruction
+ Deconstruction is now a job that takes time.
+ Deconstruction now refunds some inventory
+ Building over furniture now creates a deconstruct job; upon completion, this creates a build job

Pathfinding
+ Wall movement cost has been updated
+ Pathfinding cost of 1000000 is now impassable
+ Added debug log for job enqueueing


vogonistic:
What's the reasoning for changing movement cost 0 (impassable) to an arbitrarily high number?

Dan54:
If walls were completely impassable, the pathing system was unable to path adjacent to walls (the pathfinder must reach its destination to create a path, even if you want to go adjacent to the goal).

Assets/Scripts/Controllers/BuildModeController.cs
```diff
@@ -246,6 +251,56 @@ public bool DoesBuildJobOverlapExistingBuildJob(Tile t, string furnitureType)
         return false;
     }
 
+    public void DeconstructFurnitureAt(Tile t, string nextBuild = null)
+    {
+        Job j;
```
Grenkin1988: Please no single letter variable names. This should be - job

Assets/Scripts/Controllers/BuildModeController.cs
```diff
@@ -246,6 +251,56 @@ public bool DoesBuildJobOverlapExistingBuildJob(Tile t, string furnitureType)
...
+        if (PrototypeManager.FurnitureJobDeconstruct.HasPrototype(t.Furniture.ObjectType))
+        {
+            // Make a clone of the job prototype
+            j = PrototypeManager.FurnitureJobDeconstruct.GetPrototype(t.Furniture.ObjectType).Clone();
+
+            // Assign the correct tile.
+            j.tile = t;
+            j.OnJobCompleted += (Job jo) =>
+            {
+                t.Furniture.Deconstruct();
```
Grenkin1988: This would lead to `j` not collected by GC you need to detach after you finished.

Assets/Scripts/Models/Prototypes/PrototypeManager.cs
```diff
@@ -28,6 +29,12 @@ public static BasePrototypes<Job> FurnitureJob
         protected set;
     }
 
+    public static BasePrototypes<Job> FurnitureJobDeconstruct
```
Grenkin1988: This can be one line
`public static BasePrototypes<Job> FurnitureJobDeconstruct { get; private set; }`
and why protected? Are you going to assign this property from derived class?

Dan54: No, it just matches the style of the file.

Assets/Scripts/Controllers/BuildModeController.cs
```diff
@@ -251,43 +251,43 @@ public bool DoesBuildJobOverlapExistingBuildJob(Tile t, string furnitureType)
...
         {
             // Make a clone of the job prototype
-            j = PrototypeManager.FurnitureJobDeconstruct.GetPrototype(t.Furniture.ObjectType).Clone();
+            job = PrototypeManager.FurnitureJobDeconstruct.GetPrototype(tile.Furniture.ObjectType).Clone();
 
             // Assign the correct tile.
-            j.tile = t;
-            j.OnJobCompleted += (Job jo) =>
+            job.tile = tile;
+            job.OnJobCompleted += (Job jo) =>
```
Grenkin1988: This would lead to `job` not collected by GC you need to detach from event after you finished.

NogginBops:
@Dan54 That solution vould make it so that the characters would be able to go through walls if no other path is available and it would set the character on a LONG journey through the tile.

vogonistic:
Do we not allow diagonal adjacency when building?

Dan54:
@NogginBops It's now applied as a `<PathfindingModifier />`.

NogginBops:
@Dan54 Still I think it's a bad idea to let characters walk through walls. 
The adjacency should probably be solved in another way than letting characters walk through walls. Like changing the algorithm. (The flags just nee to be passed down to the A*, I think)

It just doesn't feel good. And the case that a job is completely cut off from the rest of the world I think this should be handled with the room graph. #760

bjubes:
Also this lets characters presumably take on a job that they can't reach since they "can" get there in 1000000 seconds

Dan54:
Characters can still walk through walls, but the pathfinder will only tell them to at the start and end of the path.

kd7uiy:
This is related (Should fix?) #180

Assets/Scripts/Models/JobQueue.cs
```diff
@@ -35,6 +35,7 @@ public int GetCount()
 
     public void Enqueue(Job j)
     {
+        Debug.Log("Enqued Job" + (j.tile == null ? "." : (" at (" + j.tile.X + ", " + j.tile.Y + ").")));
```
Sommerbo: Better to use `Debug.ULogChannel` than regular `Debug.Log` particularity with something like this that is potentially quite spammy.
And it is spelled Enqueued.

vogonistic:
I want you to split this PR so that the pathfinding changes can be discussed separately and doesn't hold up a PR about deconstruction jobs.

I don't think that walls should have a movement or pathfinding cost. If we want the characters to stop on the tile before the job, then that's what we should implement.

vogonistic:
Can you add the words `fixes #180` to the description of this PR as well?

Dan54:
@vogonistic the pathfinding system needs to be updated for the change to work as the current system must reach its destination to create a path, even if you want to go adjacent to the goal. The only other way I can think of to get the pathfinding to work would be a `Tile.GetNeighbours()` call on every node.

vogonistic:
@Dan54 I'm discussing with @WardBenjamin. We'll either fix the current pathfinder to allow you to searching for adjacent tiles to the goal, or it'll be included in the next pathfinder which is almost done.

sboigelot:

Grenkin1988:
@Dan54 Could you please make PR for **Deconstruction** only?

mikejbrown:

sboigelot:
This PR has no sign of life since 16 days ago and is blocking other people to fix the issue. I'll close it for now. @Dan54 feel free to open it again if you work on the issue again and the issue is still open!

PR #1007, opened 2016-08-29
[Small]Cleaned up lingering Debug-calls. 

Sommerbo:
`Settings.cs` fail StyleCop, needs rule change. StyleCop message: 
SA1305 : CSharp.Naming : The variable name 'xDoc' begins with a prefix that looks like Hungarian notation. Remove the prefix or add it to the list of allowed prefixes.

### DIFF
--- a/Assets/Scripts/Utilities/Settings.cs
+++ b/Assets/Scripts/Utilities/Settings.cs
@@ -245,7 +245,7 @@ public static class Settings
         XmlDocument xDoc = new XmlDocument();
         xDoc.LoadXml(furnitureXmlText);
         Debug.ULogChannel("Settings", "Loaded settings");
-        Debug.Log(xDoc.InnerText); // Uber Logger doesn't handle multilines.
+        Debug.ULogChannel("Settings", xDoc.InnerText); 
 
         // Get the Settings node. Its children are the individual settings.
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Utilities/Settings.cs
```diff
@@ -245,7 +245,7 @@ private static void LoadSettings()
         XmlDocument xDoc = new XmlDocument();
         xDoc.LoadXml(furnitureXmlText);
         Debug.ULogChannel("Settings", "Loaded settings");
-        Debug.Log(xDoc.InnerText); // Uber Logger doesn't handle multilines.
+        Debug.ULogChannel("Settings", xDoc.InnerText);
```
bjubes: This was purposely kept as a Debug.Log because , as the comment says, you cant read the full output in uberlogger. Technically, you can still see it in the normal log since uber passes messages to it, but its not very helpful to have that line in uberlogger.

Sommerbo:
Actually that call only produces one line and a very messed up one at that: `101101en_USTrueTrueTrue20021103True`
The value of that message is limited. And please note that is from current upstream and Unity console. I think that it should be reworked or removed. But regardless it fits with in the current constraints of UberLogger.

PR #1012, opened 2016-08-29
[Small] Fixed Furniture and Job context menu

MANDAL0R3:
I have fixed the context menu in Furniture.cs where i had only shown the first job in the index(i know it was ugly and not good practice), now it will properly show all jobs that can manually prioritized by a selected character. Also in Tile.cs, for its context menu, i had hard coded the name of the job because at the time there was no good way to get the name. But now that jobs are selectable, i have changed it to show the name of the object that will be prioritized. 

Also in Furniture.cs, there was a foreach loop in the GetContextMenuActions() that used a 'var', i have removed the var.

NOTE: No in game change except for a pending build job showing its name in the context menu when a character is selected.


vogonistic:
Merged!

PR #1016, opened 2016-08-29
Fixes multiline problem in UberLogger on all platforms I hope.

Sommerbo:
Replaces newlines with space, UberLogger was supposed to work this way but wasn't due to `Enviroment.NewLine` wonkyness.

Added function that should handle both Unix and Windows, tested on windows.


Mizipzor:
Awesome. I think I would have preferred to have the logger increase the height if items that are multiline. But that might look weird if you expect them to be single line. Hm. Might be bad practice anyway to attempt to log blocks of text.

PR #1020, opened 2016-08-29
Fix repetitive logic in ConstructionMenu

bjubes:
#960 noted that their PR doesn't scale well. This aims to fix this by replacing repetitive `menu.SetActive(false)` calls with easier and clearer `DeactivateSubsExcept(selectedMenu)`.

While this changes very little now, a menu with five sub menus logic would be the followin in the current master inside each buttons delegate:
```
floorMenu.SetActive(false);
furnitureMenu.SetActive(false);
powerMenu.SetActive(false);
medicalMenu.SetActive(false);
```
This changes the logic to :
`DeactivateSubsExcept(miscMenu); `

### DIFF
--- a/Assets/Scripts/UI/InGameUI/ConstructionMenu.cs
+++ b/Assets/Scripts/UI/InGameUI/ConstructionMenu.cs
@@ -21,6 +21,13 @@ public class ConstructionMenu : MonoBehaviour
     public Button buttonDeconstruction;
 
     private BuildModeController bmc;
+    private GameObject[] furnitureSubs {
+        get { 
+            
+            // add every submenu here
+            return new GameObject[] {furnitureMenu, floorMenu};
+        }
+    }
 
     public void OnClickDeconstruct()
     {
@@ -30,13 +37,13 @@ public class ConstructionMenu : MonoBehaviour
 
     public void OnClickFloors()
     {
-        furnitureMenu.SetActive(false);
+        DeactivateSubsExcept(floorMenu);
         ToggleMenu(floorMenu);
     }
 
     public void OnClickFurniture()
     {
-        floorMenu.SetActive(false);
+        DeactivateSubsExcept(furnitureMenu);
         ToggleMenu(furnitureMenu);
     }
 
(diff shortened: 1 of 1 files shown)


Assets/Scripts/UI/InGameUI/ConstructionMenu.cs
```diff
@@ -53,6 +65,17 @@ public void ToggleMenu(GameObject menu)
         menu.SetActive(!menu.activeSelf);
     }
 
+    public void DeactivateSubsExcept(GameObject menu)
+    {
+        foreach (GameObject g in FurnitureSubs)
+        {
+            if (g != menu)
+            {
+                g.SetActive(false);
```
Grenkin1988: People please stop being lazy :-( What is `g`, is it Gravitational constant?

Assets/Scripts/UI/InGameUI/ConstructionMenu.cs
```diff
@@ -22,6 +22,18 @@ public class ConstructionMenu : MonoBehaviour
...
+        get 
+        {
+            return new GameObject[] 
+            {
+                // add every furniture submenu here
+                furnitureMenu, floorMenu
+            };
+        }
+    }
+
```
Grenkin1988: Add lazy loading here.
```
private GameObject[] FurnitureSubs 
{
   get 
   {
      if(furnitureSubs == null)
      {
         furnitureSubs  = new GameObject[] 
         {
            // add every furniture submenu here
            furnitureMenu, floorMenu
         };
      }
      return furnitureSubs;
    }
 }
private GameObject[] furnitureSubs;
```

Mizipzor: I think its better to just have it as an auto property and set it whenever `furnitureMenu` or `floorMenu` is set:

    private GameObject[] FurnitureSubs { get; private set; }

Grenkin1988: OK. Than it should be
```
private HashSet<GameObject> FurnitureSubs { get; private set; }
```

bjubes:
the problem is that you have to have separate variables for `furnitureMenu` and `floorMenu` so they can be assigned in the inspector so they have to be manually added to `furnitureSubs` 

I will lazy load it, but i cant make it an auto property because then unity cant make the inspector field.

Mizipzor:
Fair enough then. :)

PR #1021, opened 2016-08-29
Air Pump

abackwood:
This adds an air pump that can be built in place of a wall. It pumps air from one room to the other. The direction of pumping can be changed through the context menu.

The pump has constraints on pressure. It pumps only if the pressure in the source room is above a certain threshold and if the pressure in the target room is below a certain threshold. This makes is usable for airlocks.
The constraints can effectively be turned off by setting source constraint to 0 and/or target constraint to a very high number.

Below, a pump has been built to pump air out of the left room into the right, keeping too much air from going into space through the airlocks. In this case, it's not perfect because the doors will keep leaking, but that's the idea.
airpumpscreenshot

The `gas_throughput` parameter determines the amount of gas it moves per second.
The `source_pressure_limit` and `target_pressure_limit` parameters define the constraints.

Sprites are placeholder and should be replaced by someone with artistic sense


abackwood:
I run into an error with this. Everything works fine when I place the first pump, but when I place the second pump all room atmospheres jump to NaN. Does anyone know what might cause this?

MANDAL0R3:
If it helps, dividing 0.0f by 0.0f return NaN

gunthergun:
Can you also limit the air pump to let a specified amount exist ?

abackwood:
@MANDAL0R3 Thanks. Might be something in calculating the gas fraction in Room.

@gunthergun not yet but good idea :smile: Will add

MANDAL0R3:
Glad i can help :)

abackwood:
That's exactly what it was. Now the NaN error is fixed. Bug #1024 persists though, but it doesn't seem like that's specific to this PR

bjubes:
does this require power?

abackwood:
@bjubes in theory, yeah. I don't really know how the new power system works but I'll try and make it work. Otherwise it doesn't and someone else can fix that

koosemose:
Actually with pressurelocking, this makes for a pretty good airlock. It won't cycle properly, but if you just set it to pump all air out of the airlock, and pressurelock the outside door, then it will pump out whatever air gets in through the inner door, and the outer door won't open until it's empty again, so you don't lose any air.

Eventually of course we will need a proper cycling, so people have time to put on Spacesuits before the air is pumped out, but it's actually really good for now

abackwood:
The direction of pumping could be changed in code to achieve the cycling effect. One thing needed then is a check against the target pressure of the RECEIVING room, not the SOURCE room as it is now

vogonistic:
We need to figure out how to make a character that is pathfinding through an airlock toggle the switches and wait for the correct events.

abackwood:
Much of that is actually implicit as long as the airlock doesn't deadlock. The door won't open until the pressure is equalized, so characters will wait as they already do for doors. As long as the airlock makes sure that all characters that want to move through an airlock eventually get their turn we'll be okay.

Desn't mean the system will be efficient or elegant. Just functional.

koosemose:
My idea was that the act of trying to go through the door would be what toggles the switch. But the problem is that the airlock door needs to communicate with the pump to make it switch flow direction and target pressure.

abackwood:
IMO the missing link here is a sort of "room controller" that controls the furniturei in a sensible way. Basically a super-furniture. I don't have any ideas for how that can be done right now.

abackwood:
Found a bug with gas being created out of nowhere. Fixing now and taking the opportunity to expand the constraints to take into account minimum pressure in the source and maximum pressure in the target

vogonistic:
You have a merge conflict

Tranberry:
Would really like to see this feature in the project

abackwood:
Im going to fix the conflicts. Probably tonight, but life has been a bit full of other stuff.

koosemose:
Minor Stylecop issues in Room.cs ChangeGas and MoveGasTo.

Issue #607, opened 2016-08-22
Missing Artwork for Version 0.1


quill18:
The following art is required to be done/redone for the Version 0.1 milestone:

- [x] Character (Kenney's art is good, and CC0, but I think it would be best if all of our assets were "custom")
- [x] Metal Smelter
- [x] O2 Generator
- [x] Wall
- [x] Door
- [x] Floor

These furniture pieces do have art, but we'll need to verify that they all look consistent with our style:
- [ ] Cloning Pod
- [ ] Power Cell Press
- [x] Power Generator (I particularly like the look of this)

Inventory Resources need to be made consistent in look. Maybe that means we keep some, or maybe that means we replace them all:
- [ ] Copper Wire
- [ ] Raw Iron
- [ ] Steel Plate
- [ ] Ice
- [ ] Uranium

zwrawr:
Looks like Tranberry (TRBRY) is on the Character art
![](https://cdn.discordapp.com/attachments/216936858528120832/217274349114097674/spaceFarer.png)
*from the art channel in the [discord chat](discord.projectporcupine.com)*

Mizipzor:
Should we add a label for these meta/todo issues? I would imagine that each checkbos gets an issue link behind it that would contain the details. Though I absolutely agree with having the list here, rather than in the wiki or somewhere else.

JosephBywater:
Also, we want an Art style that is easy to replicate so as not to dissuade people from creating art because it looks too hard!

Sommerbo:
Um, no, @zwrawr I just pointed out that the airlock graphics is the wrong perspective. I am horrible at art. So I am not the one make any graphics.

JosephBywater:
I believe that it was actually @YourShadow.

Tranberry:
Oh my, that's embarrassing.

JosephBywater:
@Tranberry Why's that embarrassing?

Tranberry:
I guess it's like very badly written code on display. But I can handle it.

TheRiven:
What's that! you want some more Programmer pixel art! welllll ok then :laughing:  
![oresmelter-64bit](https://cloud.githubusercontent.com/assets/3747487/17859077/2c9cbce0-6880-11e6-9956-ba74b58582bc.png)

WardBenjamin:
@TheRiven wow dude, that looks fantastic. Do more plsthx 

Seriously though, looks great!

[...]

PR #1008, opened 2016-08-29
[Conflicts] Resource Spites Update (Uranium, Raw Iron, and Ice)

AaronBuckles:
The Uranium, Raw Iron, and Ice sprites have been updated.
The three contain a black line around them to keep with the consistent style.
The XML files have the correct Pixels per Unit, so the sprites have a high quality look in game.

PR #1028, opened 2016-08-29
[ART] Updated Iron and Steel sprites

TheRiven:
Time to stop being lazy and contribute again :sleeping: 

Replaced Iron ore and Steel sprites with ones that are more in keeping with the "current" art style.

ironset_1
ironset_2

Also changed the XML for the Steel sprite to work with the new size.

NOT included in this update is the new Iron ore mining "spot" sprite as this is not currently implemented into the game.
ironset_0

However if people feel it should be included in this PR rather then a later one please say so and I will add it in.

Referencing #607


TheRiven:
Apparently there is another PR for Iron and Steel (also Uranium which I have produced some sprites for but not yet put into a PR) #1008 

So if people would have a look at that one as well and see what they think that would be great.

Raketfart:
Looks great. I prefer this raw Iron!

Tranberry:
I'm biased towards your submission. I believe this fits best with the current direction of the game, luckily I'm in no authority to be a nepotist.

longtomjr:
I vouch for this. It replaces art that is needed. It is also great. I thank the heavens everyday for having this to lay my eyes upon. Also, glad @TheRiven stopped playing EU4 and committed this :P

Mizipzor:
Thats enough for me. Merged.

PR #1031, opened 2016-08-29
Fixes power system issue

Grenkin1988:
Initially power system concluded that Furniture objects are not the same but that was mistake.
Now each connection has it's unique key


Grenkin1988:
Sims that this is only my imagination

Issue #982, opened 2016-08-28
[Enhancement] Quillcorp Newspaper/Announcements


koosemose:
We need a Quillcorp newpaper (and possibly one or two other more generalized papers) that pops up or otherwise becomes periodically available. Even though it started as a joke in #art on Discord, it's a great way to introduce some of the flavor of the setting, along with some random humor. It could even be used to give advance warning of some events (Get news about increase in space pirates prior to space pirate attack, etc...)

Example (from #art):
@WardBenjamin , The CEO of Quillcorp, has announced that it main and only shareholder is still the main and only shareholder, Quill18.

Notice: Quillcorp has placed an embargo on "Chairs", any Quillcorp Basic Utility Station found in possession of this illegal contraband will be fined 2 million Quillbucks.

longtomjr:
I love this idea! I am imagining a Sim City Style ticker that just have a bunch of jokes and sometimes something important :) (Thinking of sim city 3000 specifically)

abackwood:
"Introducing the MagBlast 2200 magnetic rifle. 70% of the time, it hits *every* time."
"Thinktank finds something that could go wrong. Head of QuillCorp not availble for comment"

sboigelot:
A QuillCorp newspaper... for free? That's insane :dancer: 
A let's make it cost something :D (Auto deducted from your credit balance of course, atonement is mandatory)

longtomjr:
It comes for "free" with your monthly generous Quillcorp "Protection" fee :P

kd7uiy:
I suggest it be free, with in app purchases (Or "donations") that unlock certain features. Maybe unlock an economic section, or something.

kd7uiy:
If no one else is working on it, I'm going to work on a backend for this. The backend will allow for subscribing to updates, and periodically send out a news update to be read by some front end that someone will develop. 

I will set it up such that certain conditions either have to be met to include an item, or it is automatically unlocked. The update rate will be random, as configured by the XML file giving the text to include.

kd7uiy:
Put the start of this in #1033 , note that it's only a rather weak backend, but it's a start. I added the two examples given above. It should be fairly easy to extend from there, let me know if you have any issues!

kd7uiy:
Okay, so I've started the backend and submitted it, but there's some desire to have at least someone commit to doing the front end before they will accept the PR. Anyone want to commit to doing that?

kd7uiy:
[...]

PR #1033, opened 2016-08-29
Headline Generator- basic version

kd7uiy:
Created a Headline Generator, which will allow for creating newspaper type quotes. For now, the system is very generic, it will load an XML file of potential quotes, and randomly produce the quote to anyone listening. The Unit Test shows how to properly use it. Verified that it passes all StyleCop guidelines.


Assets/Scripts/Models/HeadlineGenerator.cs
```diff
@@ -0,0 +1,69 @@
...
+// ====================================================
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Xml;
+
+public class HeadlineGenerator 
+{
+    protected Action<string> updatedHeadline;
```
bjubes: you may want to make `updatedHeadline `an event, so 
`public event Action<string> updatedHeadline`
that way you can remove the Register and Unregister functions on line 55

kd7uiy:
Easy enough I suppose.

koosemose:
Not necessarily something for this PR, but capabilities to make this more than just a way to present occasional humorous or setting-building quotes that we should consider for the future. It would be useful to have a way to both call for a specific headline, and for a list/random option from a group of headlines.

So, if, for example, being fined for having chairs were an actual thing that could happen, the event system could call for that headline, before firing off an event that would fine you if you didn't get rid of all of your chairs before a given time period.

Or, the event system could fire an event that would increase the likelihood of pirate attacks for a timer period, and could request and show a headline from among several about either pirate attacks in the area or just a general increase in activity (so as to not always show the exact same thing).

It would also be nice if headlines had a way to sub in words. So with the pirate example, instead of just saying "Pirates have been seen operating in range of the system", it could be supplied with the name of one of a group of pirate groups (if we have multiples, rather than just generic pirates), and the name of your base and/or star system, so you would end up with something along the lines of "Pirates have been seen operating in range of the Telugela System, the Infamous Dead Birds suspected!" 

That way we can have both variation, some level of customization to the player's game, and could use the headline system for important information, possibly even things celebrating what the player has done ("Archangel Base, around Telugela IIIa, has been reinvigorated, setting off an Iridium rush!")

Assets/Editor/HeadlineGeneratorTest.cs
```diff
@@ -0,0 +1,52 @@
...
+public class HeadlineGeneratorTest
+{
+    private HeadlineGenerator gen;
+    private bool stringPrinted;
+
+    [SetUp]
+    public void SetUp()
+    {
+        string filePath = System.IO.Path.Combine(UnityEngine.Application.streamingAssetsPath, "Data");
+        filePath = System.IO.Path.Combine(filePath, "Headlines.xml");
```
Grenkin1988: It is very bad practice to have tests with files dependency

Assets/Scripts/Models/World.cs
```diff
@@ -158,18 +159,32 @@ public void UpdateCharacters(float deltaTime)
...
+        // Update Furniture
         foreach (Furniture f in furnitures)
         {
             f.Update(deltaTime);
         }
 
+        // Update HeadlineGenerator
+        foreach (HeadlineGenerator h in headlineGenerators)
+        {
+            h.Update(deltaTime);
```
Grenkin1988: `h` is height?

kd7uiy:
I have a way to add a saying, I could put in a "push" option that would automatically include that saying in to the mix.

h was for HeadlineGenerator, I'll change it to hg.

As for the file dependency, I'm going to create a "Test Case" folder in StreamingAssets, which is for this kind of test cases. 

I believe when these changes are added, it should be good to go. Still can use improvements, of course, but...

kd7uiy:
As this stands, it doesn't break anything, and serves as more of an API to do two things, continue working on the back end of the HeadlineGenerator, as well as setting a framework for the Front End. I suggest that if it meets coding standards to just pull it in.

vogonistic:
I'm personally not crazy about adding a subsystem that is never called.

Code that isn't connected and is never called on should probably wait until there is something that uses it, unless someone else is eagerly waiting for it. Are you writing the continuation of the backend and the front end or is someone else doing that?

kd7uiy:
I could write the front end, but I was kind of hoping someone else would do it, although that hasn't fully been determined yet.

vogonistic:
You have merge conflicts on this. Remove `[conflicts]` when it's fixed.

Have you figured out what to do about getting someone to build a front end for this?

kd7uiy:
I have resolved the conflicts, still no word on the front end. I'm beginning to thing I'll end up doing something, maybe this weekend.

Grenkin1988:
I will repeat once again. It is bad practice to have test that depend on file system.
You should test your functionality and not file system operations

```
private string testHadlineXml = @"
<Headlines minInterval=""5"" maxInterval=""10"">
	<Headline>The CEO of Quillcorp, has announced that it main and only shareholder is still the main and only shareholder, Quill18.</Headline>
	<Headline>Notice: Quillcorp has placed an embargo on ""Chairs"", any Quillcorp Basic Utility Station found in possession of this illegal contraband will be fined 2 million Quillbucks.</Headline>
</Headlines>";

    [SetUp]
    public void SetUp()
    {
        XmlDocument doc = new XmlDocument();
        doc.Load(new StringReader(testHadlineXml));
        gen = new HeadlineGenerator(doc.SelectSingleNode("Headlines"));
        gen.UpdatedHeadline += StringPrinted;
        stringPrinted = false;
    }
```
And no need for file.

Assets/Scripts/Models/HeadlineGenerator.cs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public event Action<string> UpdatedHeadline;
+
+    public void Update(float deltaTime)
+    {
+        time += deltaTime;
+        if (time > nextTime)
+        {
+            if (UpdatedHeadline != null)
```
Grenkin1988: Add method:
```
    private void OnUpdatedHeadline(string headline)
    {
        Action<string> handler = UpdatedHeadline;
        if (handler != null)
        {
            handler(headline);
        }
    }
```
and call in both times when you invoke event.

Grenkin1988:
Why you did not use this signature?
```
    private void OnUpdatedHeadline(string headline)
    {
        Action<string> handler = UpdatedHeadline;
        if (handler != null)
        {
            handler(headline);
        }
    }
```
do you know why is it done like this?

kd7uiy:
Why create a local variable when it isn't needed?

Grenkin1988:
It is default pattern that protects code with multithreading.

Assets/Editor/HeadlineGeneratorTest.cs
```diff
@@ -15,14 +14,17 @@ public class HeadlineGeneratorTest
...
+
     [SetUp]
     public void SetUp()
     {
-        string filePath = System.IO.Path.Combine(UnityEngine.Application.streamingAssetsPath, "TestCase");
-        filePath = System.IO.Path.Combine(filePath, "Headlines.xml");
-        string furnitureXmlText = System.IO.File.ReadAllText(filePath);
         XmlDocument doc = new XmlDocument();
-        doc.Load(new StringReader(furnitureXmlText));
+        doc.Load(testHadlineXml);
```
Grenkin1988: `.Load()` need fileName
`.LoadXml()` should be used.

Assets/Editor/HeadlineGeneratorTest.cs
```diff
@@ -0,0 +1,70 @@
...
+<Headlines minInterval=""5"" maxInterval=""10"">
+    <Headline>The CEO of Quillcorp, has announced that it main and only shareholder is still the main and only shareholder, Quill18.</Headline>
+    <Headline>Notice: Quillcorp has placed an embargo on ""Chairs"", any Quillcorp Basic Utility Station found in possession of this illegal contraband will be fined 2 million Quillbucks.</Headline>
+</Headlines>";
+
+    [SetUp]
+    public void SetUp()
+    {
+        XmlDocument doc = new XmlDocument();
+        doc.Load(testHadlineXml);
```
Grenkin1988: `.Load()` needs fileName
use `.LoadXml()`

Grenkin1988:
And still probably you should add [WIP]. Right now this code do nothing by itself.

kd7uiy:
capture

It's not pretty, but it does work.

kd7uiy:
Fix #982

Assets/Scripts/Models/HeadlineGenerator.cs
```diff
@@ -0,0 +1,81 @@
...
+
+public class HeadlineGenerator 
+{
+    private List<string> headlines = new List<string>();
+    private float time, nextTime, minInterval, maxInterval;
+
+    public HeadlineGenerator(XmlNode baseNode)
+    {
+        // TODO Consider default values for these. Also consider reasonable limits
+        minInterval = float.Parse(baseNode.Attributes.GetNamedItem("minInterval").Value);
```
bjubes: See if you can hook this up to the Scheduler instead so we can keep time based events all on one system

mikejbrown: If you want random delay times with Scheduler you can do it by making the event one shot and having the event callback queue up the next event. Maybe one day Scheduler will have a nicer API to automate this, but it is possible now.

kd7uiy:
Okay, I've now got it set to use the Scheduler.

bjubes:
Awsome. one last thing. just make the interval longer, say 5 minutes, and make a way to hide the headline once the player reads it and this can be merged

kd7uiy:
The dialog now goes away after 10 seconds, you can click it to make it go away quicker, and I changed the time of headlines popping up to between 50-100 seconds. It's pretty easy to configure anyways, it's included in the XML file.

bjubes:
thanks for fixing that

PR #1042, opened 2016-08-30
[ART] Updated Uranium "ore" sprite

TheRiven:
Replaced old Uranium ore sprite with one that better fits the current style and updated the XML to match.
uraniumset_1

Not included are sprites for future mining spot and processed uranium but they exist for the future.
uraniumset_0
uraniumset_2

Referencing #607


Tranberry:
At long last it's the holidays, don't forget to bring the tree - seriously I like these sprites!

Dormanil:
Looking good, Riven! I like those sprites.

TheRiven:
Having seen it in game I feel there is a bit of a clash with the Iron ore sprite due to size but I would like other peoples opinion.

Dormanil:
Well, you could just use the same box with different chunks of ore inside, then the size wouldn't be an issue.

longtomjr:
Aaaah. This is so good. The lord riven has provided us mere presents with a feast for our eyes
I endorse this. This is art that was needed aswell and it is done amazingly!
GJ @TheRiven !

TheRiven:
I have now updated the sprite to match the base size of the iron ore. It helped to actually compare them side by side which is what I should have done in the first place :grinning:

kd7uiy:
I like the first one, looks nice!

Raketfart:
:+1: So good!

vogonistic:
Merged!

PR #1047, opened 2016-08-30
[ART] new art for power cell and copper wire, fixed scaling problem with cell

dusho:
new art for power cell and copper wire, fixed scaling problem in power cell (.xml file was not loaded and pixelPerUnit was ignored)
before:
scr_cell_before
after:
scr_cell_copper_wire


abackwood:
In the future, it's preferred that you keep art PRs limited to one sprite change each: https://github.com/TeamPorcupine/ProjectPorcupine/wiki/Standards-and-Conventions#art-pr-guideline

I think the power cell is a big improvement - even beyond the scaling being fixed. It's clearer what it is now.
The copper wire could maybe be even more 'cartoony' to fit with the art style the other art seems to follow, but it's definitely an improvement in the right direction over what we have now.

I'm inclined to say yes to both :+1:

SquareMan:
I definitely like the power cell more now, it doesn't look like a flat object. I'm not sure about the coper wire though. I looks kinda squished, but not necessarily bad though.

dusho:
@abackwood I resorted to replace copper wire just to get rid of that 64kB 'photo' of coper wire that was there before.. and ye, I agree, it's not perfect and should be changed later on
sorry about putting 2 images at once.. didn't know about that rule..

Ajapeu:
@dusho new as of yesterday, so no biggie :P I like the new power cell!

longtomjr:
@dusho Also if you want to, feel free to join the discord. We discuss a lot for the art there :) Good work. Really like the cell and we can just keep the placeholder wire for now. :)

Tranberry:
A great improvement!

One tiny bit of down is the use of so many colors. not that it looks bad more that a less color heavy palette will improve the sprites cohesiveness. The wish is that the different sprites will work better together - _this would be no problem with one artist like in most indie games_. I believe this will help more people contribute art/sprites, but I might be wrong.

**Looks good**, keep up the hard work :+1:

ComputerOverlord:
The wire is in the wrong prespective

alexisfcote:
Was it decided if we wanted isometric perspective in the sprites? If yes that would mean no-rotation of furniture, items etc...
I didn't follow the discussions in the last week so maybe it is already decided, but last I checked we had a mix of both top down view and isometric in the sprites. If there was no discussion, maybe we need to talk about it before accepting isometric perspective vs top down view.

Mizipzor:
Indeed, seems we must declare the perspective; top down or with a slight angle.

dusho:
should I try to change it then?

Tranberry:
@alexisfcote @Mizipzor maybe move the discussion on how art should be drawn to the #1073 issue - which you might object too, it being about art _request_. But where would an aspiring artist look to contribute but in a request for new art?

svmnotn:
@dusho what is the state of this PR?

dusho:
committed new version of power cell and copper wire
img:
scr_cell_copper_wire_v2
I'm sure art will be replaced later on, maybe merge it just because it fixes problem with size of old power cell

Dormanil:
@dusho When was that latest 'after' picture made; before or after the change to the sprite's filtermode was made?

dusho:
uh.. not sure when was change to filtermode made.. PR is 6 days old and I haven't re-merged this with upstream as there were no conflicts

koosemose:
Here's a screenshot from this PR pulled into the most recent version at this time.

image

dusho:
looks weird with that 32x32 character.. 64x64 one is still in works, right.. also I saw some PR where walls had outlines as well.. 
so hopefully we are closing up on common look for things

Tranberry:
It also looks weird as the walls and floor have a bunch of colors while other stuff have much less

dusho:
will close this one and put art as temp placeholders in my other PR ( #1084 )

PR #1048, opened 2016-08-30
Connection is now cloned. Fixed #1038

### LINKED ISSUE #1038: [BUG] There is a bug in power system
Right now if you add 4 oxygen generators (-1) and one solar panel (+3)
They will work.
This is because right now system does not see 4 objects it see only 1. (`HashSet<Connection>`)
This is because I forgot to `Clone()` `Connection` from one `Furniture` to other

Grenkin1988:


Assets/Scripts/Models/Power/Connection.cs
```diff
@@ -100,6 +114,11 @@ public void WriteXml(XmlWriter writer)
             writer.WriteAttributeString(accumulatedPowerAttributeName, AccumulatedPower.ToString(CultureInfo.InvariantCulture));
         }
 
+        public object Clone()
```
TomMalbran: Why not make the return value a Connection, since that is exactly what this returns. You won't need to cast it after that change.

Grenkin1988: This is signature of ICloneable

vogonistic:
Merged!

PR #1049, opened 2016-08-30
Remove debug items spawned on world load

SquareMan:
Now that we have spawning of asteroids with resources, as well as mining furniture and ways to implement item gathering in-game, these debug resources can safely be removed and clean-up the code.


vogonistic:
Merged!

PR #1052, opened 2016-08-30
Fix metal smelter job construction

### LINKED ISSUE #973: [Bug] Iron Ore disappears
gif

Note, 50 was picked up, placed on a stack of 45, and 50 was left. It seems a bit off...

Also, if we prioritize jobs, it seems like that should be a lower priority, but that's a more complex issue...

SquareMan:
fixes #973 
The lua code didn't properly access desiredItems as a table. The Metal Smelter will now only request enough items to fill it's stack, as originally intended. It also won't consume excess items if a character happens to have too many.


vogonistic:
Merged!

Issue #1035, opened 2016-08-30
Basic character statistics


mbstraus:
Different astronauts should have different things they specialize in and are capable of.  For instance, one astronaut may be better at manual labor and moving off to mine asteroids and what-not, while another may be more suited to maintenance of machinery and is more capable of repairing broken things or manufacturing new machines.

I was thinking about the technology research feature that was on the list, and I think in order to really properly handle research, there should be stats on the PCs that dictate how well they are able to do things.

The basic plan here would be to have a system set up with a few stats, proof of concept sort of thing.  I guess as a base, the stats system could work in a similar fashion to Rim World in that each character would get a random distribution of their stat points, and maybe have a few that they are particularly good at.

To start, there could be a "Mining" stat, a "Hauling" stat, a "Construction" stat, and maybe a "Research" stat.  The stats themselves could then be incorporated later into the various systems to improve / decrease the speed that a particular job would be performed.

To make things dynamic, the stats can be read via XML / LUA in the same style as how some of the other things are read in.  Then, once a character is spawned, the stats are rolled randomly for that character and applied.  Various properties can be stored in the file that dictate what the stat governs, although how that would work exactly would require more thought, the hope being to avoid having to hard-code behavior related to the stats (making it easier to apply the stat to a particular action / job).

So basically, what I am proposing, is to build a framework into the character that will support basic statistics.  For the first iteration, these stats would just be values and wouldn't be applied to anything.  However, the framework would be established to make it easy to apply these stats to whatever, which may or may not include the ability to research.

Thoughts?

koosemose:
I'd personally like to see stats broken down into attributes and skills, as some things (unskilled labor) by definition isn't going to use a skill, such as hauling. Being good at hauling is a matter of three things (that I can think of), if you're strong enough to carry heavier loads (carry more items at once), how fast you are (so you can move what you can carry to where it's going faster), and how much stamina you have (so you're made less tired from carrying loads).

Skills can change at a relatively fast rate (akin to Rimworld's skills), whereas attributes either don't change at all, or slowly change by doing activities that make use of that attribute (and maybe increasing one attribute or not using an attribute often can potentially lower other attributes/the attribute not being used), or be raised only by activities specifically dedicated to that purpose (Strength training for strength, studying for Intelligence).

Note: any actual attribute names I used for example aren't necessarily suggestions for what attributes we have, they're just pulled from my background in Tabletop RPGs.

mbstraus:
[...]

PR #1055, opened 2016-08-31
Basic character statistics

mbstraus:
This adds simple stats to the characters and the character inspector pane.  There are 6 stats total that have been added, although these can be easily tweaked via the stats.xml file.  The stats are not actually used for anything at this point, they are purely just numbers tagged to the character.  The stats will also be saved and loaded.

Related to issue #1035 

image

Apologies if I did something wrong with setting up this pull request... first time.


TomMalbran:
This looks good. One thing I would change is the make the Stats array a dictionary, so that in the future you can do something like ```Character.GetStat("Strength")```. Since we don't know where this will be used I am not sure about which methods we need for them, and what information we actually need (Do we need to know the min and max or just the value. Can we change them, maybe using an xp approach).

And one other question: Do we need to set the min and max? Or maybe we can standardize them to 1 and 20, so that all the skills look the same.

mbstraus:
@TomMalbran 

Great suggestions!  I updated it to a dictionary and added a GetStat method (old Java habits and method names starting with lower case >.<).  I also removed the min and max values... probably not needed as it stands, and is easy enough to add back in if necessary.

Assets/Scripts/Models/Stat.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+    }
+
+    protected Stat(Stat other)
+    {
+        this.statType = other.statType;
+        this.Name = other.Name;
+    }
+
+    public void ReadXmlPrototype(XmlReader reader_parent)
```
Grenkin1988: parentReader

Assets/Scripts/Models/Stat.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        statType = reader_parent.GetAttribute("statType");
+
+        XmlReader reader = reader_parent.ReadSubtree();
+        List<string> luaActions = new List<string>();
```
Grenkin1988: `luaActions` where is it used?

Assets/Scripts/Models/Stat.cs
```diff
@@ -0,0 +1,58 @@
...
+using UnityEngine;
+
+public class Stat
+{
+
+    public string statType;
+    public string Name;
+    public int MinValue;
+    public int MaxValue;
+    public Character character;
```
Grenkin1988: Why `Stat` should know about `Character`?

Assets/StreamingAssets/Data/Stats.xml
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Stats>
+  <Stat statType="Strength">
+    <Name>Strength</Name>
+  </Stat>
```
Grenkin1988: Would more elegant to have
`<Stat statType="Strength" name="Strength" />`

Assets/Scripts/Models/Character.cs
```diff
@@ -452,6 +493,29 @@ private void LoadNeeds()
...
+            // TODO: Should there be any bias or any other algorithm applied here to make stats more interesting?
+            newStat.Value = UnityEngine.Random.Range(1, 20);
+            stats.Add(newStat.statType, newStat);
+        }
+        Debug.ULogChannel("Character", "Initialized " + stats.Count + " Stats.");
+    }
+
+    public Stat GetStat(string statType)
+    {
+        Stat s = null;
```
Grenkin1988: `Stat stat = null;`

Assets/Scripts/Models/Character.cs
```diff
@@ -397,6 +410,27 @@ public void ReadXml(XmlReader reader)
...
+        {
+            Debug.ULogWarningChannel("Character", "Did not find stats for character, will keep randomized values.");
+        }
+        else
+        {
+            string[] statsStringArray = reader.GetAttribute("stats").Split(':');
+            foreach (string statString in statsStringArray)
+            {
+                string[] keyValue = statString.Split(';');
+                Stat s = GetStat(keyValue[0]);
```
Grenkin1988: `Stat stat`

Assets/Scripts/Models/Character.cs
```diff
@@ -416,7 +450,14 @@ public string GetDescription()
...
 
-        return "A human astronaut." + needText;
+        StringBuilder statText = new StringBuilder();
+        foreach (KeyValuePair<string, Stat> s in stats)
+        {
+            // TODO: Localization
+            statText.Append("\n").Append(s.Value.Name).Append(": ").Append(s.Value.Value);
+        }
+
+        return "A human astronaut." + needText + statText;
```
Grenkin1988: Use `StringBuilder` for all statistics.

Assets/Scripts/Models/Character.cs
```diff
@@ -94,6 +98,7 @@ public Character(Tile tile, Color color)
         CurrTile = DestTile = nextTile = tile;
         characterColor = color;
         LoadNeeds();
+        LoadStats();
```
Grenkin1988: Extract methods call
```
LoadNeeds();
LoadStats();
```
to separate method.

mbstraus:
@Grenkin1988 

I think that commit should address your concerns above.

Assets/Scripts/Models/Character.cs
```diff
@@ -444,20 +447,19 @@ public string GetName()
...
-            needText += "\n" + LocalizationTable.GetLocalization(n.localisationID, n.DisplayAmount);
+            description.Append("\n").Append(LocalizationTable.GetLocalization(n.localisationID, n.DisplayAmount));
         }
 
-        StringBuilder statText = new StringBuilder();
         foreach (KeyValuePair<string, Stat> s in stats)
         {
             // TODO: Localization
-            statText.Append("\n").Append(s.Value.Name).Append(": ").Append(s.Value.Value);
+            description.Append("\n").Append(s.Value.Name).Append(": ").Append(s.Value.Value);
```
Grenkin1988: -> description.AppendLine(string.Format("{0}: {1}", s.Value.Name, s.Value.Value));

Assets/Scripts/Models/Character.cs
```diff
@@ -444,20 +447,19 @@ public string GetName()
...
+        StringBuilder description = new StringBuilder();
+        description.Append("A human astronaut.");
         foreach (Need n in needs)
         {
-            needText += "\n" + LocalizationTable.GetLocalization(n.localisationID, n.DisplayAmount);
+            description.Append("\n").Append(LocalizationTable.GetLocalization(n.localisationID, n.DisplayAmount));
         }
 
-        StringBuilder statText = new StringBuilder();
         foreach (KeyValuePair<string, Stat> s in stats)
```
Grenkin1988: If you need only `Values` you should iterate by `stats.Values`

Assets/Scripts/Models/Character.cs
```diff
@@ -357,6 +366,13 @@ public void WriteXml(XmlWriter writer)
         }
 
         writer.WriteAttributeString("needs", needString);
+
+        StringBuilder statsString = new StringBuilder();
+        foreach (KeyValuePair<string, Stat> stat in stats)
+        {
+            statsString.Append(stat.Value.statType).Append(";").Append(stat.Value.Value).Append(":");
+        }
+        writer.WriteAttributeString("stats", statsString.ToString());
```
TomMalbran: I know that it might be easier to make your own parser. But is better to follow the xml way and save the stats like:
```
<Stats>
  <Stat name="Strength" value="10" />
</Stats>
```
It will be easier to extend and read when we add more attributes to the stats like xp.

TomMalbran:
You should also fix the StyleCop errors, I can see a few by just reading the code.

Assets/Scripts/Models/Character.cs
mbstraus: I was following the same style that the Needs were using, thus why I went with that way.  I do agree that it is cleaner to save them as separate elements like that.. I can certainly make that change... but unless it is desired, i would prefer not to touch needs unless I have to, but it will be inconsistent if I don't.

TomMalbran: Oh. I didn't saw that the needs where done in that way. The needs code has several issues that need fixing. You shouldn't change the Needs in this PR since this PR is about Stats, but we should fix them in another PR.

mbstraus:
I think that should take care of things.  You guys are tough, I like it :)

vogonistic:
This branch has merge conflicts. Please remove [conflicts] from the title when fixed.

mbstraus:
Conflicts have been resolved.

vogonistic:
Merged!

vogonistic:
@mbstraus I missed the StyleCop errors in this PR. Can you fix them and submit a new PR?

mbstraus:
Sure, although I won't be able to till tomorrow.

mbstraus:
@vogonistic Looks like #1083 is being worked to take care of all of the StyleCop issues.

PR #1056, opened 2016-08-31
Fix param saving/loading

koosemose:
This is a redo of PR #819 .

It fixes how Parameters are saved and loaded, and changes the save format of Parameters match how they are written in the furniture prototypes, allowing the same Loading code to handle both.

This fixes a bug where the parameters of furniture (and any other object that uses Parameters) weren't properly loaded on Load Game, causing most furniture that uses Parameters to not work on Load.


vogonistic:
Merged!

PR #1057, opened 2016-08-31
Adding Beds

Ohda:
- Add a simple bed furniture. (Need 2 steel plates to create)
- Add Sleeping as a need.
- Add ph_ as the prefix for sprite's placeholder. 
- Add a default bed in the starting area.

There is still a lot of improvment needed in the Needs System.
Main problems are : 
- A Need require a furniture for our characters to complete it. If furniture doesn't exsit in the current world it will produce an error. (This is true for the beds but also for the O² generator).
- Caracters can use the same beds, at the same time.


Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -409,4 +409,30 @@
...
+    <BuildingJob jobTime="1">
+      <Inventory objectType="Steel Plate" amount="2" />
+    </BuildingJob>
+
+    <DeconstructJob>
+      <Inventory objectType="Steel Plate" amount="1" />
+    </DeconstructJob>
+
+    <Params>
+      <Param name="confort" value="0.5" />
```
koosemose: Little typo here.

Tranberry:
Is the sprite a place-holder or do you consider it final?

 I would really like to push for a file prefix(ph_) to place-holder sprites if this is a place-holder that is. But afaik there are _no_ consensus on that.

dusho:
I wouldn't name placeholder files differently.. it's much easier just to replace existing file in Unity
placeholders could be identified by image content (like mining drone)

Tranberry:
@dusho I really don't understand what you are trying to say :(

Ohda:
@Tranberry I will add (ph_). I really hope someone will do better !

dusho:
@Tranberry maybe I didn't get your point.. so you want sprites that should be replaced later named with prefix ph_ ? Then .xml file must be named ph_ as well.. right? But ideally when new art comes, you don't want to keep renaming .xml files as well..
That is why I mentioned that if it's a placeholder.. reflect it in image itself (same as is now for mining drone  image - where you have box with letters MINING DRONE)

Tranberry:
That is another solution, but I would really like to be able to search ph_*.png .

bjubes:
i agree `ph_` is very clear so we don't have to search and see if the author actually wants their art or just opened up MSPaint to give their PR an image. Its easy to rename xml files for sprites, since no one else is going to touch the files unless their also changing the image. because renaming in this scenario poses little chance for merge conflicts, it only adds clarity that the file is meant to be replaced

Ohda:
I just removed the WIP tag. 
Merging it or waiting for improvment of the need system, that'is the question here !

vogonistic:
There are still some StyleCop warnings in this commit. Please fix them before we merge.

Ohda:
Fixed

vogonistic:
Merged!

PR #1059, opened 2016-08-31
Ore Mining - Step 1

TheRiven:
Woo back to doing some code! :grinning: 

Adds in Iron ore and Uranium ore mining points/spots as well as sprites for them.
ironset_0
uraniumset_0

These mines are currently buildable by the player but this will be removed later once they have been integrated into the world gen. 

The player can right click on them and give the order to mine them. This generates a high priority job that causes a character to wonder over and "mine" the ore. Once the job is complete the mining point is removed and a stack of whatever ore was mined is placed there instead.

Problems to solve going forward:

- Currently multiple furnitures need to be created to handle each ore type, a single generic furniture would be preferable.
- Needs to not be buildable by the player! :laughing: 
- Needs to to be used by the world gen instead of placing stacks of ore around the place.

Now! feel free to point out my mistakes :smile:


TheRiven:
And to state the next steps:
Step 2 - Change world gen to use mining points
Step 3 - Make them not buildable by the Player
Step 4 - Generic Mining point furniture (hopefully)

Grenkin1988:
You can use params for this.
MiningPoint param "ore"=iron 
Something like this

TheRiven:
Hmm technically I already have that as it uses the `ore_type` to determine what to drop when its mined. So that's good :+1:

dusho:
can iron ore actually be silverish, gray, blueish ? instead of brown.. same goes for Raw Iron.png, I currently have it like this locally (just because I'm playing with general smelting...)
raw iron

NogginBops:
@dusho Why should raw iron be silverish/gray, any particular reason? It's more realistic with rusty brown.

dusho:
well.. when I was collecting rocks, iron ore had veins that were silverish. Rusty brown with greenish tint had copper as it reacts with air..
Also when checking games like Factorio, iron is gray, copper is brown, coal is black and so on..

vogonistic:
I would prefer the base of the art to be a generic asteroid rock type that we can join together a bit like the walls. Each asteroid we have now is really a slice through the asteroid where you could mine. We'd need one type where you can't see what it is and then when you mine an adjacent rock, we'd instead show a variation that shows the ore type (iron, copper, ..). Similar to how DF and Rimworld works.

Mizipzor:
@NogginBops Iron should not be brown or rusty. 

> [Rust](https://en.wikipedia.org/wiki/Rust) is an iron oxide, usually red oxide formed by the redox reaction of iron and oxygen in the presence of water or air moisture.

Thats kinda rare in space, dont ya think? ;)

TheRiven:
@vogonistic Yeah that's what I had in mind for the future as well, but I felt it would be better to start small and simply switch the system from generating loose inventories all over the asteroids to placing furniture that can be interacted with by the player instead. As this was a sprite I had already made I went with that. This can then be changed over in a later stage when we expand it into a wall based system like Rimworld/DF's.

@dusho @Mizipzor The sprite looks like that because when I presented it the consensus was that it would look better in a rusty orange colour. If enough people think otherwise then it can of course be changed. :+1:

vogonistic:
I'm ok with this as a step on the way. It's better than having scattered resources on the ground.

TheRiven:
Apparently making the world generator place mines instead of the ores was a somewhat easier change then I expected and it will still place resources that do not have a mine associated with them so ice still spawns.

So stage 2 is done although I would like to know if anyone has a better idea then just adding some Ifs in there as I have done.

I was thinking of using the resource objectTypes and just added them onto the end of a `mine_` string but to do this I would need to make sure the ore names all follow the same pattern first. Does this sound like a reasonable way forward for that? Is it even worth doing it now as the goal is to later make the mine (and eventually the wall) into a generic object and just set an `ore_type`?

Grenkin1988:
@TheRiven This looks good. But I would really recommend to do it generic from the beginning.
You would need just one furniture. mine
And have `ore_type` parameter
Than when you generate mines in in `WorldGenerator` you just need to put different `ore_type` for next mine.
For sprits - we alteady have methods `<GetSpriteName FunctionName="" />` you will need to generate name of sprite by `mine + _ + ore`

TheRiven:
Ok the mine is now generic as @Grenkin1988  suggested, it was not as much work as I thought it would be which is why I put it off.

I am however not happy with how I have it set up in the world gen, I feel that editing the prototype is not the right way to go about this, what do others think?

I could not see another way to change the ore_type before creating the object and I don't see a way to trigger the furniture changed event which would set the sprite again. Have I missed something? is there a sneaky function I have failed to see? keeping up with all the changes is tricky :smile:

Assets/Scripts/Models/WorldGenerator.cs
```diff
@@ -102,6 +102,16 @@ public static void Generate(World world, int seed)
 
                                 if (randomweight <= currentweight)
                                 {
+                                    
+                                    if (inv.objectType == "Raw Iron" || inv.objectType == "Uranium")
+                                    {
+                                        Furniture mine = PrototypeManager.Furniture.GetPrototype("mine");
```
Grenkin1988: add `.Clone();` to the end and this will not be prototype anymore.
And create overload function in World

```
public Furniture PlaceFurniture(string objectType, Tile t, bool doRoomFloodFill = true)
    {
        // TODO: This function assumes 1x1 tiles -- change this later!
        if (PrototypeManager.Furniture.HasPrototype(objectType) == false)
        {
            Debug.ULogErrorChannel("World", "furniturePrototypes doesn't contain a proto for key: " + objectType);
            return null;
        }

        Furniture furn = Furniture.PlaceInstance(PrototypeManager.Furniture.GetPrototype(objectType), t);

        return PlaceFurniture(furn, t, doRoomFloodFill );
    }

public Furniture PlaceFurniture( Furniture furniture, Tile t, bool doRoomFloodFill = true)
    {
        if (furn == null)
        {
            // Failed to place object -- most likely there was already something there.
            return null;
        }

        furn.Removed += OnFurnitureRemoved;
        furnitures.Add(furn);

        // Do we need to recalculate our rooms?
        if (doRoomFloodFill && furn.RoomEnclosure)
        {
            Room.DoRoomFloodFill(furn.Tile);
        }

        if (OnFurnitureCreated != null)
        {
            OnFurnitureCreated(furn);

            if (furn.MovementCost != 1)
            {
                // Since tiles return movement cost as their base cost multiplied
                // buy the furniture's movement cost, a furniture movement cost
                // of exactly 1 doesn't impact our pathfinding system, so we can
                // occasionally avoid invalidating pathfinding graphs.
                // InvalidateTileGraph();    // Reset the pathfinding system
                if (tileGraph != null)
                {
                    tileGraph.RegenerateGraphAtTile(t);
                }
            }
        }

        return furn;
    }
```

vogonistic:
Definitely need to clone the FurniturePrototype before it is placed.

TheRiven:
I would like to, but apparently the clone function is currently private. Is there a particular reason for this? Is it ok to make it public?

My assumption is that the whole point of having the clone function is so that you could do the sort of things I am now trying to do, am I mistaken?

Grenkin1988:
@TheRiven Yes, `Furniture.Clone();` should be `public`

TheRiven:
Thanks @Grenkin1988 Clone is now public and I have added the overload to deal with placing furniture that you already have a furniture object for.

How does it look?

ComputerOverlord:
I want mining into asteroids be like mining into mountains in rimworld

vogonistic:
I get this warning: `PlaceInstance -- Position Validity Function returned FALSE.` 20 or so times with this PR. Do you know what that's about?

TheRiven:
@vogonistic I think its because the Validity Function does not allow furniture to be placed to close to the edge of the map. It not actually causing an error as such, just warning us that its not allow to spawn anything within 5 tiles of the edge.

Also @ComputerOverlord  that is the next step on from this, I expect almost everything being done here will be needed for that as well, this is just a simpler implementation to start with.

TheRiven:
@vogonistic I have now tested my theory and by changing the `MinEdgeDistance` value to 0 in furniture you no longer get the error. 

The question is, what do we want to do about this as this will also cause problems when we switch over to using mineable walls later. I feel that this is outside the scope of this PR though, what do you think?

vogonistic:
Thanks for figuring that out.

I don't like having errors or warnings come up on a clean run. It's a terrible precedent to start ignoring them and makes testing a lot harder. 

Can you either add a flag so that we don't check validity on map creation or not spawn furniture close to the edge?

Assets/Scripts/Models/WorldGenerator.cs
```diff
@@ -102,16 +102,17 @@ public static void Generate(World world, int seed)
...
-                                        break;
-                                    }
 
                                         if (randomweight <= currentweight)
                                         {
+                                            if (inv.objectType == "Raw Iron" || inv.objectType == "Uranium")
+                                            {
+                                                Furniture mine = PrototypeManager.Furniture.GetPrototype("mine").Clone();
+                                                mine.Parameters["ore_type"].SetValue(inv.objectType.ToString());
+                                                world.PlaceFurniture(mine, t, false);
```
vogonistic: Could you wrap this block in a check like this to remove the errors?

```
Furniture minePrototype = PrototypeManager.Furniture.GetPrototype("mine");
if (minePrototype.IsValidPosition(t))
{
    // place mine here.
}
```

There seems to be something off with the indentation in this block as well. My MonoDevelop was very unhappy about it. You don't have to fix that, but it would be a plus point if you did.

TheRiven:
@vogonistic the solution I came up with is not the one you suggested but I do feel it is the better direction. Now the world gen will not generate anything within the minimum distance to the edge, this stops the errors and also means you don't get bits of asteroid with nothing on it.

This will also be needed for when we switch to filling the asteroid with walls as we would end up with gaps at the edge otherwise.

The code for the edge check is reused code from the furniture validation check and `minEdgeDistance` now has 2 separate versions, which should probably be merged somewhere central where we can define where the edge of the useable map is perhaps?

Assets/Scripts/Models/WorldGenerator.cs
```diff
@@ -83,44 +85,60 @@ public static void Generate(World world, int seed)
             {
                 for (int y = 0; y < height; y++)
                 {
-                        float noiseValue = Mathf.PerlinNoise((x + offsetX) / (width * asteroidNoiseScale * zScale), (y + offsetY) / (height * asteroidNoiseScale * zScale));
-                        if (noiseValue >= asteroidNoiseThreshhold && !IsStartArea(x, y, world))
+                    float noiseValue = Mathf.PerlinNoise((x + offsetX) / (width * asteroidNoiseScale * zScale), (y + offsetY) / (height * asteroidNoiseScale * zScale));
+                    if (noiseValue >= asteroidNoiseThreshhold && !IsStartArea(x, y, world))
+                    {
+                        Tile t = world.GetTileAt(x, y, z);
```
Grenkin1988: `Tile tile`

TheRiven:
@Grenkin1988 Fixed :+1:

bjubes:
still has conflicts

TheRiven:
@bjubes And fixed again! :+1:

koosemose:
All the issues seem to be taken care of, and I haven't found any other bugs from this. The remaining step will need further changes outside the scope of this PR (or are art related), Merged.

PR #1064, opened 2016-08-31
Time controller

janpeli:
Added time controller to have all time related code in one place. Cleaned up Keyboard controller and World controller.


TomMalbran:
I was thinking that we should so something like this, and I like it.

A few questions:
1. Should the TimeController keep track of the play/pause state of the game?
2. If we do 1, should we remove the IsModal thing and make the Dialogs call the TimeController to pause the game? This is so that we can get the real pause state only from the TimeController.
3. Should we make the TimeController a singleton?
4. Is this really a controller? it feels more like a structure that keeps data, more like a model that is used by other controllers. It probably should be a TimeManager, or something like that.

Mizipzor:
It is not a controller because its not attached to a GameObject, which is how we define a controller (afaik). Maybe move this to utilities?

TomMalbran:
Is not a Utility either since it stores data and doesn't provide utility functions. Is more a Model, but the Model definition is getting too wide, maybe we should split them in subgroups.

janpeli:
@TomMalbran 
1. I think it should. But I would like to have confirmation from other people on this.
2. I think that the is modal should go into timeController update method as an input parameter and pause the game based on that. IsModal is something else as paused. While paused you should be able to click on things while in modal probably not. I would keep the IsModal as it is because it would create spaghetti if called from different places at different times.
3. Depends on whether if we have multiple worlds we will have only one time or multiple times. Other controllers are also not singletons so I guess it is fine not being singleton.
4. I think we can rename it and maybe move the initialization to OnEnable.

janpeli:
Maybe later there should be some other object like "Calendar" that would store the total time since start of the game and this would be saved/loaded from save file.

vogonistic:
Merged!

TomMalbran:
I think that the TimeManager should keep track of the pause state, and we should keep the modal state, but isMode shouldn't be equal to isPaused. If a Dialog wants to pause the game when it pops up because is a modal, it can request it to the TimeManager. With that done we can remove IsPaused and TimeScale from the World controller.

vogonistic:
Is that ideas for a future PR or something that should have gone in this one?

TomMalbran:
it could have gone in this one. But since is merged it can be done in a separate PR.

Issue #1071, opened 2016-08-31
RoomChanged event


svmnotn:
We need an event to be fired whenever a room is changed, be it removed, added, or whatever else. So far the idea is having a `RoomChanged` event that would contain what happened, most likely represented by an enum so something like:
```c#
enum RoomChange {
   Created,
   Destroyed,
}
```

we could have different variants for `Split` and `Joined` or simply pass more information about the result of what has taken place. I will get around to doing this shortly so for now this will be the tracking issue, feel free to attempt to influence my PR before it is made. :D

vogonistic:
I think we need 4 events:
- Created
- Split
- Joined
- Destroyed

For `Split` we need to know who we split from and `Joined` we need to know who we joined into (the room that is left behind after the join).

Btw, this system is there to help with the upcoming Job System. If we have a job queue per room, we need these events to be able to migrate jobs to the new appropriate queues. The PR is already huge so we are trying to create PR's that might be useful for other things separately.

koosemose:
I also think a room should be alerted when its furniture changes... but that's only applicable if a room is keeping track of its furniture in the first place, which is somewhat of a separate matter.

vogonistic:
After a brief discussion on the Split and Joined events, we decided to have a discussion that is at least short, or medium long, on if it's worth having those extra events. They don't exist in underlying code and distinctions aren't always clear.

koosemose:
But would it be beneficial to have things done more in that way? Particularly Joined. For example, one notable issue is that currently when two rooms are joined one's atmosphere overwrites the other, that really needs changed (along with other considerations when joining rooms) Would that be enough to justify a Joined event, since it would be more of a distinction from creating a room?

vogonistic:
We just have to make sure we cover all the edge cases. With split events, we have the same problems right? But do we send the event to both of the rooms and keep a reference to the original room, even though all tiles have already been moved away to the new rooms?

vogonistic:
In the 4-event system, `Created` is entirely when it was created in space (or split from room 0) and `Destroyed` would only be when it is opened to space (or joined with room 0). Everything is really join and split events, but there is a question on who gets them and what objects do we need to keep.

svmnotn:
[...]

Issue #1000, opened 2016-08-28
Naming standard for events


Grenkin1988:
I would like to discuss this.
Why do we decided on such naming standard as – On…
From my practice, events always called just – Changed, Removed, Closed, Click.
Quite often methods attached to these events names as – OnClick, OnChanged…
There are some discussions about this:
[Events - naming convention and style](http://stackoverflow.com/questions/724085/events-naming-convention-and-style)
[Naming, declaring and defining delegates and events conventions [closed]](http://stackoverflow.com/questions/682635/naming-declaring-and-defining-delegates-and-events-conventions)

And even .NET classes do not have this On…
[Form Events](https://msdn.microsoft.com/en-us/library/system.windows.forms.form_events(v=vs.110).aspx)

Event is something that happened to object. It could `Changed`. And reaction to this could be `OnChanged`.

[Names of Type Members](https://msdn.microsoft.com/en-us/library/ms229012.aspx)
>Names of Events
Events always refer to some action, either one that is happening or one that has occurred. Therefore, as with methods, events are named with verbs, and verb tense is used to indicate the time when the event is raised.
✓ DO name events with a verb or a verb phrase.
Examples include Clicked, Painting, DroppedDown, and so on.
✓ DO give events names with a concept of before and after, using the present and past tenses.
For example, a close event that is raised before a window is closed would be called Closing, and one that is raised after the window is closed would be called Closed.
X DO NOT use "Before" or "After" prefixes or postfixes to indicate pre- and post-events. Use present and past tenses as just described.
✓ DO name event handlers (delegates used as types of events) with the "EventHandler" suffix, as shown in the following example:
public delegate void ClickedEventHandler(object sender, ClickedEventArgs e);
✓ DO use two parameters named sender and e in event handlers.
The sender parameter represents the object that raised the event. The sender parameter is typically of type object, even if it is possible to employ a more specific type.
✓ DO name event argument classes with the "EventArgs" suffix.

sboigelot:
Feel free to change, the naming convention you propose make sense :)

A side question is: when to use event and when to not use event. Events brings a lot of overhead, while sometime, if you just want to register one action and not N actions, you could just use an Action.

I would love to have your feedback on that too.

Grenkin1988:
@sboigelot As for me, this no big difference 1 subscriber or more. This is used to decouple objects from each other.
One of them is just saying - I did this, and it does not care whether anything is interested in this.
[...]

PR #1072, opened 2016-09-01
RoomEvents as per #1071

svmnotn:
I believe that this is now as complete as I can imagine it. At least until @vogonistic gives some insight into how the `Joined` event should behave. But it should be ok to merge now, and change later if any of the functionality needs to be altered.

### DIFF
--- a/Assets/Scripts/Models/Area/Room.cs
+++ b/Assets/Scripts/Models/Area/Room.cs
@@ -42,7 +42,7 @@ namespace ProjectPorcupine.Rooms
             }
         }
 
-        public int Tiles
+        public int TileCount
         {
             get
             {
@@ -96,7 +96,7 @@ namespace ProjectPorcupine.Rooms
                 tiles[i].Room = World.Current.RoomManager.OutsideRoom;
             }
 
-            tiles = new List<Tile>();
+            tiles.Clear();
         }
 
         public bool IsOutsideRoom()
@@ -121,7 +121,7 @@ namespace ProjectPorcupine.Rooms
             }
         }
 
-        public Tile[] FindExits()
+        public List<Tile> FindExits()
         {
             List<Tile> exits = new List<Tile>();
             foreach (Tile tile in tiles)
@@ -136,18 +136,18 @@ namespace ProjectPorcupine.Rooms
                             // We have found an exit
                             exits.Add(tile2);
                         }
-                    }      
+                    } 
                 }
             }
 
-            return exits.ToArray();
+            return exits;
         }
 
         public Dictionary<Tile, Room> GetNeighbours()
         {
             Dictionary<Tile, Room> neighboursRooms = new Dictionary<Tile, Room>();
 
-            Tile[] exits = this.FindExits();
+            List<Tile> exits = this.FindExits();
 
             foreach (Tile tile in exits)
             {
@@ -172,6 +172,7 @@ namespace ProjectPorcupine.Rooms
         }
 
         // Changes gas by an amount in preasure(in atm) multiplyed by number of tiles
+        // TODO check this method, it doesn't seem like the above comment is accurate.
         public void ChangeGas(string name, float amount)
         {
             if (IsOutsideRoom())
@@ -207,7 +208,7 @@ namespace ProjectPorcupine.Rooms
             }
         }
 
-        public string ChangedGases(string name)
+        public string ChangeInGas(string name)
         {
             if (deltaGas.ContainsKey(name))
             {
@@ -262,7 +263,7 @@ namespace ProjectPorcupine.Rooms
         {
             if (atmosphericGasses.ContainsKey(name))
             {
-                return atmosphericGasses[name] / Tiles;
+                return atmosphericGasses[name] / TileCount;
             }
 
             return 0;
@@ -286,7 +287,7 @@ namespace ProjectPorcupine.Rooms
 
             foreach (float pressure in atmosphericGasses.Values)
             {
-                totalPressure += pressure / Tiles;
+                totalPressure += pressure / TileCount;
             }
 
             return totalPressure;
@@ -359,7 +360,7 @@ namespace ProjectPorcupine.Rooms
         {
             foreach (string n in other.atmosphericGasses.Keys)
             {
-                this.atmosphericGasses[n] = other.atmosphericGasses[n] / sizeOfOtherRoom * this.Tiles;
+                this.atmosphericGasses[n] = other.atmosphericGasses[n] / sizeOfOtherRoom * this.TileCount;
             }
         }
 
--- a/Assets/Scripts/Models/Area/Room.cs
+++ b/Assets/Scripts/Models/Area/Room.cs
@@ -36,7 +36,15 @@ public class Room : IXmlSerializable
     {
         get
         {
-            return World.Current.GetRoomID(this);
+            return World.Current.RoomManager.GetRoomID(this);
+        }
+    }
+
+    public int Tiles
+    {
+        get
+        {
+            return tiles.Count;
         }
     }
 
@@ -48,117 +56,34 @@ public class Room : IXmlSerializable
         }
     }
 
-    public static void EqualiseGasByTile(Tile tile, float leakFactor)
+    public void AssignTile(Tile tile)
     {
-        List<Room> roomsDone = new List<Room>();
-        foreach (Tile t in tile.GetNeighbours())
-        {
-            // Skip tiles with a null room (i.e. outside).
-            // TODO: Verify that gas still leaks to the outside
-            // somehow.
-            if (t.Room == null)
-            {
-                continue;
-            }
-
-            if (roomsDone.Contains(t.Room) == false)
-            {
-                foreach (Room r in roomsDone)
-                {
-                    t.Room.EqualiseGas(r, leakFactor);
-                }
-
-                roomsDone.Add(t.Room);
-            }
-        }
-    }
-
-    public static void DoRoomFloodFill(Tile sourceTile, bool onlyIfOutside = false)
-    {
-        // SourceFurniture is the piece of furniture that may be
-        // splitting two existing rooms, or may be the final 
-        // enclosing piece to form a new room.
-        // Check the NESW neighbours of the furniture's tile
-        // and do flood fill from them.
-        World world = World.Current;
-
-        Room oldRoom = sourceTile.Room;
-
-        if (oldRoom != null)
-        {
-            // Save the size of old room before we start removing tiles.
-            // Needed for gas calculations.
-            int sizeOfOldRoom = oldRoom.GetSize();
-
-            // The source tile had a room, so this must be a new piece of furniture
-            // that is potentially dividing this old room into as many as four new rooms.
-
-            // Try building new rooms for each of our NESW directions.
-            foreach (Tile t in sourceTile.GetNeighbours())
-            {
-                if (t != null && t.Room != null && (onlyIfOutside == false || t.Room.IsOutsideRoom()))
-                {
-                    ActualFloodFill(t, oldRoom, sizeOfOldRoom);
-                }
-            }
-
-            sourceTile.Room = null;
-
-            oldRoom.tiles.Remove(sourceTile);
-
-            // If this piece of furniture was added to an existing room
-            // (which should always be true assuming with consider "outside" to be a big room)
-            // delete that room and assign all tiles within to be "outside" for now.
-            if (oldRoom.IsOutsideRoom() == false)
-            {
-                // At this point, oldRoom shouldn't have any more tiles left in it,
-                // so in practice this "DeleteRoom" should mostly only need
-                // to remove the room from the world's list.
-                if (oldRoom.tiles.Count > 0)
-                {
-                    Debug.ULogErrorChannel("Room", "'oldRoom' still has tiles assigned to it. This is clearly wrong.");
-                }
-
-                world.DeleteRoom(oldRoom);
-            }
-        }
-        else
-        {
-            // oldRoom is null, which means the source tile was probably a wall,
-            // though this MAY not be the case any longer (i.e. the wall was 
-            // probably deconstructed. So the only thing we have to try is
-            // to spawn ONE new room starting from the tile in question.
-
-            // You need to delete the surrounding rooms so a new room can be created
-            // This doesn't work for the gas calculations and needs to be fixed.
-            foreach (Tile t in sourceTile.GetNeighbours())
-            {
-                if (t != null && t.Room != null && (onlyIfOutside == false || t.Room.IsOutsideRoom()))
-                {
-                    world.DeleteRoom(t.Room);
-                }
-            }
-
-            ActualFloodFill(sourceTile, null, 0);
-        }
-    }
-
-    public void AssignTile(Tile t)
-    {
-        if (tiles.Contains(t))
+        if (tiles.Contains(tile))
         {
             // This tile already in this room.
             return;
         }
 
-        if (t.Room != null)
+        if (tile.Room != null)
         {
             // Belongs to some other room.
-            t.Room.tiles.Remove(t);
+            tile.Room.tiles.Remove(tile);
         }
 
-        t.Room = this;
-        tiles.Add(t);
+        tile.Room = this;
+        tiles.Add(tile);
+    }
+
+    public void UnassignTile(Tile tile)
+    {
+        if (tiles.Contains(tile))
+        {
+            // This tile in not in this room.
+            return;
+        }
+
+        tile.Room = null;
+        tiles.Remove(tile);
     }
 
     public void ReturnTilesToOutsideRoom()
@@ -166,7 +91,7 @@ public class Room : IXmlSerializable
         for (int i = 0; i < tiles.Count; i++)
         {
             // Assign to outside.
-            tiles[i].Room = World.Current.GetOutsideRoom();
+            tiles[i].Room = World.Current.RoomManager.OutsideRoom;
         }
 
         tiles = new List<Tile>();
@@ -174,12 +99,7 @@ public class Room : IXmlSerializable
 
     public bool IsOutsideRoom()
     {
-        return this == World.Current.GetOutsideRoom();
-    }
-
-    public int GetSize()
-    {
-        return tiles.Count();
+        return this == World.Current.RoomManager.OutsideRoom;
     }
 
     // Changes gas by amount in pressure (atm) per tile, evenly distributed over all gases present
@@ -227,21 +147,21 @@ public class Room : IXmlSerializable
 
         Tile[] exits = this.FindExits();
 
-        foreach (Tile t in exits)
+        foreach (Tile tile in exits)
         {
             // Loop over the exits to find a different room
-            Tile[] neighbours = t.GetNeighbours();
-            foreach (Tile tile2 in neighbours)
+            Tile[] neighbours = tile.GetNeighbours();
+            foreach (Tile neighbor in neighbours)
             {
-                if (tile2 == null || tile2.Room == null)
+                if (neighbor == null || neighbor.Room == null)
                 {
                     continue;
                 }
 
                 // We have found a room
-                if (tile2.Room != this)
+                if (neighbor.Room != this)
                 {
-                    neighboursRooms[tile2] = tile2.Room;
+                    neighboursRooms[neighbor] = neighbor.Room;
                 }
             }
         }
@@ -340,7 +260,7 @@ public class Room : IXmlSerializable
     {
         if (atmosphericGasses.ContainsKey(name))
         {
-            return atmosphericGasses[name] / GetSize();
+            return atmosphericGasses[name] / Tiles;
         }
 
         return 0;
@@ -353,26 +273,21 @@ public class Room : IXmlSerializable
             return 0;
         }
 
-        float t = 0;
+        float totalPressure = GetTotalGasPressure();
 
-        foreach (string n in atmosphericGasses.Keys)
-        {
-            t += atmosphericGasses[n];
-        }
-
-        return t == 0 ? 0 : atmosphericGasses[name] / t;
+        return totalPressure == 0 ? 0 : atmosphericGasses[name] / totalPressure;
     }
 
     public float GetTotalGasPressure()
     {
-        float t = 0;
+        float totalPressure = 0;
 
-        foreach (string n in atmosphericGasses.Keys)
+        foreach (float pressure in atmosphericGasses.Values)
         {
-            t += atmosphericGasses[n] / GetSize();
+            totalPressure += pressure / Tiles;
         }
 
-        return t;
+        return totalPressure;
     }
 
     public void MoveGasTo(Room room, float amount)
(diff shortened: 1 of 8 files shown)


Assets/Scripts/Models/Room.cs
```diff
@@ -19,18 +19,49 @@
...
+    }
+
     public Room()
     {
         tiles = new List<Tile>();
         atmosphericGasses = new Dictionary<string, float>();
         deltaGas = new Dictionary<string, string>();
     }
 
+    public static event Action<Room> OnRoomCreated;
```
Grenkin1988: `static` are you sure?

Grenkin1988:
Yes please add more events before we made final decision on event naming standard (#1000).

Assets/Scripts/Models/Room.cs
```diff
@@ -370,34 +424,34 @@ public void ReadXml(XmlReader reader)
         }
     }
 
-    protected static void ActualFloodFill(Tile tile, Room oldRoom, int sizeOfOldRoom)
+    protected static Room ActualFloodFill(Tile tile, Room oldRoom, int sizeOfOldRoom)
```
Grenkin1988: I think this method should not belong to `Room`.

TomMalbran:
I think that it might be a good time to create a RoomManager class. We can move the room list and all the room functions from the World into the RoomManager, and maybe the ActualFloodFill function from Room. Then we could add all the Room events into the RoomManager, which will make more sense to add them there as they are changes in the room structure, and not so much in the room itself.

Assets/Scripts/Models/Room.cs
svmnotn: I have it as `static` because otherwise you would have to set each of the events on a per `Room` basis, I don't think anyone actually wants to do that, therefore they are static, the other option is having something like a `RoomManager` of sorts that handles all of that, in which case these could be non-`static` but part of the instance of that manager.

Assets/Scripts/Models/Room.cs
svmnotn: What should it belong to then?

svmnotn:
@TomMalbran that is precisely what I came up with responding to @Grenkin1988, but I think it would be a bit more of an intrusive change than the current format, however I am willing to do that change if it everyone is willing to have all code related to rooms break a bit after the PR.

TomMalbran:
@svmnotn I think is the best solution and it should be done. We shouldn't do stuff in a wrong way because is less intrusive. If it breaks other PR, that is ok, when you submit a PR you need to be willing to deal the merge conflicts, specially when you do a PR that changes several files.

Assets/Scripts/Models/Room.cs
TomMalbran: I would move DoRoomFloodFill, ActualFloodFill and EqualiseGasByTile into the new RoomManager class, since all those deal with possible multiple rooms. Neither of those functions should be static. To call then you would do something like ```World.Current.RoomManager.DoRoomFloodFill```.

Assets/Scripts/Models/Room.cs
```diff
@@ -206,6 +65,18 @@ public void AssignTile(Tile t)
         tiles.Add(t);
     }
 
+    public void UnAssignTile(Tile t)
```
bjubes: don't capitalize the A in `UnassignTile`

Assets/Scripts/Models/RoomManager.cs
```diff
@@ -0,0 +1,320 @@
+﻿using System;
```
TomMalbran: Add the copyright header.

Assets/Scripts/Models/RoomManager.cs
```diff
@@ -0,0 +1,320 @@
...
+
+    public event Action<Room, Room> Split;
+
+    public event Action<Room, Room> Joined;
+
+    public event Action<Room> Removing;
+
+    public event Action<Room> Removed;
+
+    public List<Room> rooms;
```
TomMalbran: We should make this private.

Assets/Scripts/Models/World.cs
```diff
@@ -26,7 +26,7 @@ public class World : IXmlSerializable
     public readonly string GameVersion = "Someone_will_come_up_with_a_proper_naming_scheme_later";
     public List<Character> characters;
     public List<Furniture> furnitures;
-    public List<Room> rooms;
+    public RoomManager roomManager;
```
TomMalbran: We should make this a private set.

svmnotn: for this I followed the structure of the one below, the `InventoryManager`. But I do think an accessor would be better.

Assets/Scripts/Models/RoomManager.cs
svmnotn: it is needed in other places, and I didn't want to make a GetRooms function as it would either copy all the data(expensive-ish) or do the same thing...

Assets/Scripts/Models/Room.cs
svmnotn: heh copy/paste strong!

Assets/Scripts/Models/RoomManager.cs
Grenkin1988: Name if public property

Assets/Scripts/Models/World.cs
TomMalbran: We should change all of them eventually. We can do it slowly with the new ones.

Assets/Scripts/Models/RoomManager.cs
TomMalbran: The big problem is that we can't allow other to add or remove rooms from this list, they need to use the functions. Having it as a public property is not good.

An alternative is having a Count function and a GetRoom function and use a for which should not have any performance problem. It would be cleaner with a GetRooms, but it might be an issue if we need to copy them.

svmnotn: I could try to do something that yields every room on the list or something like that... it might work.. Stay tuned!

Assets/Scripts/Models/RoomManager.cs
```diff
@@ -0,0 +1,328 @@
...
+    {
+        rooms = new List<Room>();
+    }
+
+    public Room GetOutsideRoom()
+    {
+        return rooms[0];
+    }
+
+    public int GetRoomID(Room r)
```
Grenkin1988: No single letter variables. Yes...yes... there are a lot of them in code. But if you change code - make it even better.

Assets/Scripts/Models/RoomManager.cs
```diff
@@ -39,50 +39,52 @@ public RoomManager()
...
     {
-        return rooms.IndexOf(r);
+        return rooms.IndexOf(room);
     }
 
-    public Room GetRoomFromID(int i)
+    public Room GetRoomFromID(int index)
     {
-        if (i < 0 || i > rooms.Count - 1)
+        if (index < 0 || index > rooms.Count - 1)
```
Grenkin1988: It was absolutely OK to have `i` :-)

svmnotn: pfft! no more single letters!

Assets/Scripts/Models/RoomManager.cs
```diff
@@ -0,0 +1,354 @@
...
+// file LICENSE, which is part of this source code package, for details.
+// ====================================================
+#endregion
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+public class RoomManager: IEnumerable<Room>
+{
+    public event Action<Room> Adding;
```
Grenkin1988: SA1201 : CSharp.Ordering : All events must be placed after all constructors.

Assets/Scripts/Models/RoomManager.cs
```diff
@@ -0,0 +1,354 @@
...
+        get { 
+            return rooms.Count;
+        }
+    }
+
+    public Room GetOutsideRoom()
+    {
+        // TODO maybe find some other way of knowing what the outside is?
+        // or perhaps have it not just be a 0 here, maybe a const?
+        return rooms[0];
```
TomMalbran: Maybe the RoomManager should create the outside room in the constructor instead of assuming that it always have a Room at index 0 and that that one is the outside one.

Assets/Scripts/Models/RoomManager.cs
TomMalbran: index is good here since is not a for loop

Assets/Scripts/Models/RoomManager.cs
svmnotn: That is probably the best way, but where is this assumption first made? (where are we actually adding that first room that becomes the outside?)

Grenkin1988: Here:
```
RoomManager = new RoomManager();
RoomManager.AddRoom(new Room()); // Create the outside?
```

svmnotn: Is that the only place? I thought I might be missing one or two..

svmnotn:
@TomMalbran @Grenkin1988 any help with the Join event would be greatly appreciated :D

vogonistic:
This looks like a fantastic improvement. Other than join using stale data, what else is it you want to improve in it?

svmnotn:
@vogonistic just the join thing... I haven't figured out a way of solving that...

TomMalbran:
Do we actually need the Joined event? It might be cool, but having Created and Removed events might work just fine.

svmnotn:
I think we do, since then all the special atmo handling could be moved to only happen on join/split events. I'm just not quite sure how to do it... If I can't think of anything today I'll just make it a TODO and switch the PR to not WIP and let someone else take care of that.

TomMalbran:
The best way to check if rooms were joined is to check if the tiles of the old rooms are part of the new room. Note that upto 4 rooms could be joined by destroying a wall. Not sure if the joined event should list them all or just the 2.

Now if we can assume that you destroy only one wall at a time, and since you can't walk in between 2 diagonal walls, then there might be a simpler way. The way I am thinking is that after destroying only 1 wall, 2 rooms will be joined only if there are walls north and south of the tile and not at the east and west, or if there are walls east and west of the tile and not at the north and south, or if the wall is a diagonal with no walls north, south, east or west. 3 rooms can be joined into one if the wall has 1 wall north, south, east or west, and 2 diagonals on the oposite side. And 4 rooms can be joined if the tile has a wall on each corner and none in the south, north, east west.

These would be the combinations I am talking about. You are destroying the middle Wall and checking the neighbours and the result is the amount of rooms.

```
W: Wall
_: Floor/Empty
*: Anything

2 rooms
*W*    *_*
_W_    WWW
*W*    *_*

3 rooms
W_W    *_W    *W*    W_*
_W_    WW_    _W_    _WW
*W*    *_W    W_W    W_*

4 rooms
W_W
_W_
W_W
```
I might be missing some cases.

svmnotn:
@TomMalbran I think those are the cases, but that is basically what is currently done, the issue is that once the new room is created, all the old rooms (potentially up to 4) have already been removed. And I can't delay the removal because the floodFill depends on the tile not... oh wait... can I do that? ... Time to check!

TomMalbran:
I am not sure how it will work, I haven't checked the algorithm in a while. You kind of need to know something of the state before flood filling.

svmnotn:
I think I will look into remaking the flood fill, as currently something got broken somewhere...

mikejbrown:

sboigelot:
@svmnotn what is the status of this PR? Do you still need help? If yes, can you give us more information about what you need?

svmnotn:
@sboigelot I just got time again to work on it, did a major rebase as it was quite behind. I just need to know what the expectations are for the `Joined` event? Because as it stands right now it uses rooms which are no longer managed by the RoomManager and as such their IDs cant be read, also their contents have been transferred to the new room, so I'm not sure exactly what is needed. I guess for now this could be merged after I add a few comments detailing such behavior and then we can go back and change it later?

sboigelot:
I'm not familiar enougth with this part of the code

ping @TomMalbran  & @Grenkin1988 what do you think about it?

svmnotn:
@sboigelot from what @TomMalbran commented on discord, I think this is good to merge unless @vogonistic has any ideas about how the `Joined` event should behave.

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,429 @@
...
+        /// </summary>
+        public event Action<Room> Removed;
+
+        #endregion
+
+        /// <summary>
+        /// Gets the amount of rooms.
+        /// </summary>
+        /// <value>The number of rooms managed.</value>
+        public int Rooms
```
TomMalbran: I think this should be Count. It will make it more clear.

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,429 @@
...
+            { 
+                return rooms.Count;
+            }
+        }
+
+        /// <summary>
+        /// Gets the outside room.
+        /// </summary>
+        /// <value>The outside room.</value>
+        public Room OutsideRoom
```
TomMalbran: I think we should start using ```public Room OutsideRoom { get; private set; }``` in this cases, since we don't need 4 lines for it.

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,429 @@
...
+        {
+            return rooms.IndexOf(room);
+        }
+
+        /// <summary>
+        /// Gets the room from an ID.
+        /// </summary>
+        /// <returns>The room based on the ID.</returns>
+        /// <param name="index">The ID of the room. Possibly the index.</param>
+        public Room GetRoomFromID(int index)
```
TomMalbran: Should we change this to using public ```Room this[int index]```? Since we are using it like that in several places?

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,429 @@
...
+// This program comes with ABSOLUTELY NO WARRANTY; This is free software, 
+// and you are welcome to redistribute it under certain conditions; See 
+// file LICENSE, which is part of this source code package, for details.
+// ====================================================
+#endregion
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace ProjectPorcupine
```
TomMalbran: Should the Namespace be ```ProjectPorcupine.Area``` or ```ProjectPorcupine.Room``` ?

Assets/Scripts/Models/Area/World.cs
```diff
@@ -124,60 +123,14 @@ public World()
...
-    }
-
-    public void AddRoom(Room r)
-    {
-        rooms.Add(r);
-        roomGraph = null;
-
-        Debug.ULogChannel("Rooms", "creating room:" + r.ID);
-    }
+    public ProjectPorcupine.RoomManager RoomManager { get; private set; }
```
TomMalbran: I like it better when we add the namespace to using.

sboigelot: Yes this is important.

Assets/Scripts/Models/Area/RoomManager.cs
vogonistic: The naming conventions encourage namespaces to be plural unless it doesn't make sense. I vote for `ProjectPorcupine.Rooms`

vogonistic:
Let's not hold up this PR any longer. We can figure out the Joined event when we have a strong need for it.

I think @TomMalbran has some really good code changes and that we should merge as soon as they are implemented.

svmnotn:
... stupid git thingy...

Assets/Scripts/Models/Area/World.cs
svmnotn: Done, did the same to the `PowerNetwork` so as to show some standardization.

Assets/Scripts/Models/Area/RoomManager.cs
svmnotn: I didn't think StyleCop would like it... It did

Assets/Scripts/Models/Area/RoomManager.cs
svmnotn: Done, I ... naming is hard...

Assets/Scripts/Models/Area/RoomManager.cs
svmnotn: I prefer `ProjectPorcupine.Rooms` too, although that is not the folder name... it might be good to make a separate folder? not sure here.

Assets/Scripts/Models/Area/Room.cs
```diff
@@ -15,546 +15,360 @@
...
+                return World.Current.RoomManager.GetRoomID(this);
+            }
         }
-    }
 
-    public static void EqualiseGasByTile(Tile tile, float leakFactor)
-    {
-        List<Room> roomsDone = new List<Room>();
-        foreach (Tile t in tile.GetNeighbours())
+        public int Tiles
```
sboigelot: This should be TileCount

Assets/Scripts/Models/Area/Room.cs
```diff
@@ -15,546 +15,360 @@
...
+            {
+                // Assign to outside.
+                tiles[i].Room = World.Current.RoomManager.OutsideRoom;
+            }
 
-        if (t.Room != null)
-        {
-            // Belongs to some other room.
-            t.Room.tiles.Remove(t);
+            tiles = new List<Tile>();
```
sboigelot: tiles.Clear() ? not much changes but assign less GC mess

Assets/Scripts/Models/Area/Room.cs
```diff
@@ -15,546 +15,360 @@
...
+                ChangeGas(name, amount * fraction);
+            }
         }
-    }
 
-    public Tile[] FindExits()
-    {
-        List<Tile> exits = new List<Tile>();
-        foreach (Tile tile in tiles)
+        public Tile[] FindExits()
```
sboigelot: why using an array and not a IEnumerable or a List<T>? If you build it as a List<T> wouldn't it be easier to keep it as is?

Assets/Scripts/Models/Area/Room.cs
```diff
@@ -15,546 +15,360 @@
...
+            {
+                atmosphericGasses[name] = 0;
+            }
         }
-    }
 
-    public string ChangedGases(string name)
-    {
-        if (deltaGas.ContainsKey(name))
+        public string ChangedGases(string name)
```
sboigelot: this method name do not describe the method body, can you have a more meaningful name?

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,427 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace ProjectPorcupine.Rooms
+{
+    public class RoomManager : IEnumerable<Room>
+    {
+        /// <summary>
+        /// A all the currently managed rooms.
```
sboigelot: comment should say
A list of all the currently managed rooms

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,427 @@
...
+            rooms = new List<Room>();
+
+            // Create the outside room.
+            OutsideRoom = new Room();
+
+            // Add the outside rooom to the list just in case.
+            rooms.Add(OutsideRoom);
+        }
+
+        #region Events
```
sboigelot: no region, I think

sboigelot: even though I personalty like them in this case

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,427 @@
...
+        /// Gets the outside room.
+        /// </summary>
+        /// <value>The outside room.</value>
+        public Room OutsideRoom { get; private set; }
+
+        /// <summary>
+        /// Gets a managed room from an ID.
+        /// </summary>
+        /// <param name="index">The ID of the room.</param>
+        public Room this [int index]
```
sboigelot: no space between this and [
You would kill stylecop on visual studio

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,427 @@
...
+            return rooms.IndexOf(room);
+        }
+
+        /// <summary>
+        /// Add a room to the manager.
+        /// </summary>
+        /// <param name="room">A room to be managed.</param>
+        public void Add(Room room)
+        {
+            World.Current.roomGraph = null;
```
sboigelot: shouldn't this be the world registering to the adding & removing event? this is not the responsability of the room manager

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,427 @@
...
+        {
+            if (room.IsOutsideRoom())
+            {
+                Debug.ULogErrorChannel("World", "Tried to delete the outside room.");
+                return;
+            }
+
+            Debug.ULogChannel("Rooms", "Removing room:" + room.ID);
+
+            World.Current.roomGraph = null;
```
sboigelot: same comment as above

Assets/Scripts/Models/Area/RoomManager.cs
```diff
@@ -0,0 +1,427 @@
...
+
+        /// <summary>
+        /// Remove a room from the manager.
+        /// </summary>
+        /// <param name="room">The room to remove.</param>
+        public void Remove(Room room)
+        {
+            if (room.IsOutsideRoom())
+            {
+                Debug.ULogErrorChannel("World", "Tried to delete the outside room.");
```
sboigelot: I would remove this error message if this is a expected case

Assets/Scripts/Models/Area/Room.cs
svmnotn: This is not a method I changed, what happened was a bit of a whitespace change perhaps combined with a different source or something? I think git is just being weird. However I will change the name, just as a cleanup.

svmnotn:
@sboigelot fixed the comments

Assets/Scripts/Models/Area/RoomManager.cs
svmnotn: I think it is not an expected case

Assets/Scripts/Models/Area/RoomManager.cs
svmnotn: fixed?

Assets/Scripts/Models/Area/RoomManager.cs
svmnotn: stylecop on mono was fine with this...

Assets/Scripts/Models/Area/RoomManager.cs
svmnotn: I liked it too... oh well

svmnotn: Maybe we should look to see if more regions are better... we could maybe do an Issue to find what the "sweet" spot for this might be

Assets/Scripts/Models/Area/RoomManager.cs
svmnotn: fixed

Assets/Scripts/Models/Area/Room.cs
svmnotn: Again, this is how it was, the only reason git sees a difference is because of some weirdness but I didn't really change anything there. However, I have changed it to List as it creates one either way

Assets/Scripts/Models/Area/Room.cs
svmnotn: fixed

Assets/Scripts/Models/Area/Room.cs
svmnotn: fixed

Assets/Scripts/Models/Area/RoomManager.cs
sboigelot: Yes on mono it won't crash, trust me on this :) VS will die, it's a bug in Stylecop ;)

sboigelot:
testing now

Assets/Scripts/Models/Area/RoomManager.cs
TomMalbran: You get this error everytime you delete a wall in the outside room, or that is adjacent to the outside room. So it seems like there is an error on the code or this can happen and we shouldn't log an error.

PR #1074, opened 2016-09-01
Made changes to CONTRIBUTING.md following up on #1000

Grenkin1988:


Mizipzor:
I fear that might blunder might have caused confusion in the other thread. Should we maybe formally define with exampled as I did, or at least try to do. The `delegate`, the `event` and the `eventhandler`? Or does that bloat the text. Maybe link to a wiki page that goes into the details?

Grenkin1988:
@Mizipzor you mean add this: [Names of Type Members](https://msdn.microsoft.com/en-us/library/ms229012(v=vs.110).aspx) ?

vogonistic:
I'm merging without another link to the naming conventions. There already is one in the file.

PR #1080, opened 2016-09-01
[small] Always check against NullReferenceException when using tile.GetNeighbours!

mikejbrown:
(Don't know if this bug has been reported yet.)
To reproduce the bug build floor tiles along the edge of the map. This fixes it.

### DIFF
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -513,7 +513,7 @@ public class Character : IXmlSerializable, ISelectable, IContextActionProvider
         // If the dest tile does not have neighbours that are walkable it's very likely that they can't be walked to
         if (DestTile != null)
         {
-            if (DestTile.GetNeighbours().Any((tile) => { return tile.MovementCost > 0; }) == false)
+            if (DestTile.GetNeighbours().Any((tile) => (tile != null && tile.MovementCost > 0)) == false)
             {
                 Debug.ULogChannel("Character", "No neighbouring floor tiles! Abandoning job.");
                 AbandonJob(false);
--- a/Assets/Scripts/Models/Character.cs
+++ b/Assets/Scripts/Models/Character.cs
@@ -290,7 +290,7 @@ public class Character : IXmlSerializable, ISelectable, IContextActionProvider
         DestTile = MyJob.tile;
 
         // If the dest tile does not have neighbours that are walkable it's very likable that they can't be walked to.
-        if (DestTile.GetNeighbours().Any((tile) => { return tile.MovementCost > 0; }) == false)
+        if (DestTile.GetNeighbours().Any((tile) => (tile != null && tile.MovementCost > 0)) == false)
         {
             Debug.ULogChannel("Character", "No neighbouring floor tiles! Abandoning job.");
             AbandonJob(false);
--- a/Assets/Scripts/Models/Room.cs
+++ b/Assets/Scripts/Models/Room.cs
@@ -87,7 +87,7 @@ public class Room : IXmlSerializable
             // Try building new rooms for each of our NESW directions.
             foreach (Tile t in sourceTile.GetNeighbours())
             {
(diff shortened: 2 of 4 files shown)


vogonistic:
I thought we added a safe zone around the edges of the universe? Odd.

mikejbrown:
Hmmm... it's probably worth doing a global search of the codebase to see if there are any more of these. _Any_ time you use Tile.GetNeighbours() you have to do a check because it is not guaranteed to return non-null values!

mikejbrown:
I think that's all of them now.

Grenkin1988:
I think more obvious is to change
```
    /// <summary>
    /// Gets the neighbours.
    /// </summary>
    /// <returns>The neighbours.</returns>
    /// <param name="diagOkay">Is diagonal movement okay?.</param>
    public Tile[] GetNeighbours(bool diagOkay = false)
    {
        Tile[] tiles = diagOkay == false ? new Tile[4] : new Tile[8];

        Tile tileAt = World.Current.GetTileAt(X, Y + 1);
        tiles[0] = tileAt; // Could be null, but that's okay.
        tileAt = World.Current.GetTileAt(X + 1, Y);
        tiles[1] = tileAt; // Could be null, but that's okay.
        tileAt = World.Current.GetTileAt(X, Y - 1);
        tiles[2] = tileAt; // Could be null, but that's okay.
        tileAt = World.Current.GetTileAt(X - 1, Y);
        tiles[3] = tileAt; // Could be null, but that's okay.

        if (diagOkay == true)
        {
            tileAt = World.Current.GetTileAt(X + 1, Y + 1);
            tiles[4] = tileAt; // Could be null, but that's okay.
            tileAt = World.Current.GetTileAt(X + 1, Y - 1);
            tiles[5] = tileAt; // Could be null, but that's okay.
            tileAt = World.Current.GetTileAt(X - 1, Y - 1);
            tiles[6] = tileAt; // Could be null, but that's okay.
            tileAt = World.Current.GetTileAt(X - 1, Y + 1);
            tiles[7] = tileAt; // Could be null, but that's okay.
        }

        return tiles.Where(tile => tile != null).ToArray();
    }
```

koosemose:
I believe it was furniture that was blocked from building within 5 (or so) of the border, apparently tiles need the same treatment.

vogonistic:
@Grenkin1988 I think when that was originally implemented, the order was special. No idea if anyone depends on it.
@koosemose Ah, that's what we did.

Grenkin1988:
getneighboursusage
Here is `GetNeighbours()` usage
Check someone if order matters.
Is not, that why we need to return `null`

mikejbrown:
I remember in one of the videos quill was debating whether or not it should return nulls. He decided in the end that a constant mapping of indices => directions was more important.

vogonistic:
@Grenkin1988 It's not clear if you said that you've checked all uses of them or if you think someone else should.

Grenkin1988:
@vogonistic I've wrote
> Check someone if order matters.

mikejbrown:
Hmm... I don't think it matters any more. Anyone remember what the ordering was used for?

bjubes:
from @Grenkin1988 's screenshot it it seems the majority/ all of the uses of getNeighbors is to iterate over the list, so removing nulls will be cleaner and save an extra loop

mikejbrown:
I've made the change. It seems to work fine. Will anyone vouch that nothing breaks?

DeShrike:
I think order matter because at some point in the code the opposite neighbour is accessed by some culculation on  the indexes: 
Something like :
var northNeightbor = neightbours[ (ix + 2) % 4 ]

mikejbrown:
@DeShrike Yeah, I remember something like that being in the original code as well, but going through the current references to GetNeighbours I couldn't find anything like that still in the code. It's possible I missed something though so would appreciate a check.

vogonistic:
Merged!

PR #1083, opened 2016-09-02
Fixed all build and StyleCop warnings

vogonistic:
- Fixed all build warnings for unused variables
 - Fixed all StyleCop warnings
 - Fixed a bug where `PathfindingModifier` wasn't used
 - Changed `MovementCost` to be a property with a private setter
 - Committed a lot of auto-corrections that MonoDevelop does with the settings in the CONTRIBUTING.md file.


Assets/Scripts/Controllers/KeyboardController.cs
```diff
@@ -26,7 +25,6 @@ public class KeyboardController
     // Use this for initialization.
     public KeyboardController(BuildModeController buildModeController, WorldController worldController)
```
TomMalbran: Since you aren't using ```buildModeController``` you don't need it as a parameter.

vogonistic: Odd that there isn't a StyleCop warning for an unused parameter in a non-override or interface function.

Assets/Scripts/Models/Character.cs
```diff
@@ -295,7 +290,10 @@ public void PrioritizeJob(Job job)
         DestTile = MyJob.tile;
 
         // If the dest tile does not have neighbours that are walkable it's very likable that they can't be walked to.
-        if (DestTile.GetNeighbours().Any((tile) => { return tile.MovementCost > 0; }) == false)
+        if (DestTile.GetNeighbours().Any((tile) =>
```
TomMalbran: Should we make a ```HasWalkableNeighbours()``` method in Tile and use it here, since this line is used twice in this file.

Assets/Scripts/Models/Character.cs
```diff
@@ -295,7 +290,10 @@ public void PrioritizeJob(Job job)
         DestTile = MyJob.tile;
 
         // If the dest tile does not have neighbours that are walkable it's very likable that they can't be walked to.
-        if (DestTile.GetNeighbours().Any((tile) => { return tile.MovementCost > 0; }) == false)
```
Grenkin1988: `if (DestTile.GetNeighbours().Any(tile =>tile.MovementCost > 0) == false)` is better

Assets/Scripts/Models/Furniture.cs
```diff
@@ -791,8 +794,8 @@ private Furniture Clone()
     private bool DefaultIsValidPosition(Tile t)
```
Grenkin1988: t => tile

Assets/Scripts/Models/Tile.cs
```diff
@@ -410,7 +409,7 @@ public IEnumerable<ContextMenuAction> GetContextMenuActions(ContextMenu contextM
             {
                 Text = "Cancel Job",
                 RequireCharacterSelected = false,
-                Action = (cm, c) => 
+                Action = (cm, c) =>
```
Grenkin1988: Could we move this actions to methods and assign methods instead of lambda expressions?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -422,136 +422,136 @@ public void ReadXmlPrototype(XmlReader readerParent)
...
+                    replaceableFurniture.Add(reader.GetAttribute("typeTag").ToString());
+                    break;
                 case "DragType":
-                reader.Read();
-                DragType = reader.ReadContentAsString();
-                break;
+                    reader.Read();
+                    DragType = reader.ReadContentAsString();
+                    break;
                 case "BuildingJob":
```
TomMalbran: We should make a function for this case and reduce a bit the length of this function.

Assets/Scripts/Models/Character.cs
vogonistic: Sounds good. Adding it now.

Assets/Scripts/Models/Tile.cs
```diff
@@ -112,15 +112,14 @@ public float PathfindingCost
 
     public int Y { get; private set; }
 
-    public float MovementModifier;
+    public float MovementModifier { get; private set; }
```
TomMalbran: The idea of it being public set, was to be able to change it from Lua.

Assets/Scripts/Models/Character.cs
vogonistic: I'm using your suggestion on syntax, but refactoring it into `Tile.HasWalkableNeighbours()`

Assets/Scripts/Models/Furniture.cs
vogonistic: The scope of this PR is primarily to get rid of build warnings, StyleCop warnings and at least some of the MonoDevelop auto-corrections. I'm fixing this since I've already touched most lines in this function.

Grenkin1988: + 100 to karma :-)

Assets/Scripts/Models/Tile.cs
vogonistic: We could, but I don't want to expand the scope of this PR too far. It's already big in terms of touched lines.

Assets/Scripts/Models/Furniture.cs
vogonistic: I agree, but I don't think it's a good fit for this PR.

Assets/Scripts/Models/Tile.cs
vogonistic: I think we need a system of setting conditions or statuses instead of directly modifying the movementmodifier, but I'm changing it back to public until we have that in place.

Assets/Scripts/Models/Tile.cs
Grenkin1988: OK

Grenkin1988:
Could we add  `//TODO:`  `//FIME:`  where we think is good place for refactoring for future PRs?

Assets/Scripts/Models/Tile.cs
TomMalbran: I agree if those conditions can be modable too. Once we do that we might not even need the MovementModifier, since the Modifier will come from the Tile Condition.

vogonistic:
I'm not sure a TODO or FIXME helps very much. I've yet to hear someone choosing to refactor something based on a TODO in the code. Almost everything is based on adding features. 

There are currently 76 TODO and 29 FIXME in the Assets directory.

Grenkin1988:
I was going to look through them after finishing with `Power` :-)

Assets/Scripts/Models/Tile.cs
vogonistic: We haven't done any real design on it, but I'm thinking you add the condition `Bloody` to the tile so there can be a specific job to clean it up later. But while it's set, the movement is modified for that Tile.

koosemose:
No unexpected crashes, and I can't see any evidence of any functional changes due to unexpected side effects of the PR.

Assets/Scripts/Controllers/SpawnInventoryController.cs
```diff
@@ -105,9 +105,9 @@ private void CreateInventoryButtons()
             button.colors = colorBlock;
             string localName = invName;
             button.onClick.AddListener(
-                () => { OnButtonClick(localName); });
+                () => OnButtonClick(localName));
```
TomMalbran: This would look better as ```button.onClick.AddListener(() => OnButtonClick(localName));``` or as a 3 lines.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -759,7 +759,10 @@ public IEnumerable<ContextMenuAction> GetContextMenuActions(ContextMenu contextM
                     {
                         Text = "Prioritize " + Name,
                         RequireCharacterSelected = true,
-                        Action = (ca, c) => { c.PrioritizeJob(jobs[0]); }
+                        Action = (ca, c) =>
```
TomMalbran: Should we change this to: ```Action = (ca, c) => c.PrioritizeJob(jobs[0]);```?

vogonistic:
Merging my own PR.

Issue #604, opened 2016-08-22
Producer/Consumer pattern for furniture


Grenkin1988:
Recently I was think of some generic pattern for our furniture so it could handle complicated stuff:
-   Some furniture could produce multiple resources
-   Some furniture could consume multiple resources
-   Some furniture could produce multiple things while consuming another

My thought was to have something like:

```
<Consume>
    <Item name=”” amount=””/>
</Consume>
<Produce>
    <Item name=”” amount=””/>
</ Produce >
```

Some thoughts were here #602
What do you think about this?

Xeni91:
I think it is good but we need to implement a time aspekt in this. Consumption and Production can be after every operation, but when some furniture needs a constant supply of power/gas/oxygen we should be able to set the intervall.

Grenkin1988:
Right now we need to gather all the requirements for this, and decide whether this is good idea at all

mikejbrown:
So `<Item name=”” amount=”” rate=""/>` instead?

How about this calling arrangement:

- Producer OnUpdate makes X units of resource Y & calls a callback to a consumer registered for listening to that resource (scheduling could be an issue if there are >1 consumers?), essentially giving an _offer_ to the consumer...
- The Consumer, in its callback, decides whether or not it wants to use that resource & how much. Then it calls a function on the producer saying "I'm taking Z much of Y now" so...
- The Producer can now update its state, offer any remaining resource to the next consumer etc...

Just quickly sketching ideas here. Thoughts?

dusho:
It definitely makes sense to make some generic factory. From the looks at Metal Smelter it should cover most of LUA code ( together with that stacking code - it shouldn't be in LUA at all)
Would this cover all that is needed?:
```
<Factory processingSpeed="">
	<Input>
		<FactoryItem name="" amount="" position=""></FactoryItem>
		<FactoryItem name="" amount="" position=""></FactoryItem>
	</Input>
	<Output>
		<FactoryItem name="" amount="" position=""></FactoryItem>
		<FactoryItem name="" amount="" position=""></FactoryItem>
	</Output>
</Factory>
```
where position would be spawn points (required position for input, spawn position for output), and processingSpeed would be rate of taking all items from Input and producing Output. If Output is full (can't stack anymore) or Input is not fulfilled, factory will not produce anything.

Grenkin1988:
I would definitely go for making derived class for `Factory`

RodRitter:
[...]

PR #1084, opened 2016-09-02
Attempt at generic factory furniture (consume/produce)

dusho:
Working in generic consume/produce (discussed in #604)
Can do:
- added optional Workshop (FurnitureWorkshop.cs) to Furniture .xml
- furniture can handle multiple production chains (recipes)
- can have multiple input/output slots
- added context menu if there is more than 1 available production chain
- load/save seems to be working (Inventory.locked is not persisted, after load, items from input are taken away - new issue, I guess)

There is new temporary Metal Furnace (that works as Metal Smelter) just to try things out. Idea is to get rid of most of repeated code in LUA for furniture that consumes and produces something
New art for Power cell press, input: 3x uranium, 2x steel plate, output: 1x power cell


Assets/Editor/FactoryInfoTest.cs
```diff
@@ -0,0 +1,48 @@
...
+public class FactoryInfoTest
+{
+    [Test]
+    public void TestSerialization()
+    {
+        FactoryInfo fi = new FactoryInfo();        
+        fi.PossibleProductions = new System.Collections.Generic.List<FactoryInfo.ProductionChain>();
+        //fi.InputSlots = new List<FactoryInfo.IntVector2>() { new FactoryInfo.IntVector2(1, 0) };
+        //fi.OutputSlots = new List<FactoryInfo.IntVector2>() { new FactoryInfo.IntVector2(2, 0) };
+        var chain1 = new FactoryInfo.ProductionChain() { Name="Iron smelting", ProcessingTime = 4.0f };
```
vogonistic: This test file doesn't look formatted according to the rules in the CONTRIBUTING.md file. It's pretty hard to read.

Assets/Scripts/Models/Furniture.Factory.cs
```diff
@@ -0,0 +1,185 @@
...
+        // - add param to reflect factory can provide output (has output inside)
+        //   - as output will be produced after time, it is possible that output spot can be ocupied meanwhile
+        // - process for specified time
+        // - if output slot is free, provide output (if not, keep output 'inside' factory)
+
+        var curSetupChainName = furnParameters[CUR_PRODUCTION_CHAIN_PARAM_NAME].ToString();
+
+        if (!string.IsNullOrEmpty(curSetupChainName))
+        {
+            FactoryInfo.ProductionChain cChain = null;
```
vogonistic: You should avoid prefixes on variables. If you want it really short, maybe `prodChain`? I'd name it `productionChain`.

Assets/Scripts/Models/Furniture.Factory.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+            // if there is no processing in progress
+            if (furnParameters[CUR_PROCESSED_INV_PARAM_NAME].ToInt() == 0)
+            {
+                // there is nothing being processed now                
+                List<KeyValuePair<Tile, int>> flaggedForTaking = new List<KeyValuePair<Tile, int>>();
+                foreach (var reqInputItem in cChain.Input)
+                {
+                    // check input slots for req. item:                        
+                    Tile tt = World.Current.GetTileAt(Tile.X + reqInputItem.SlotPosX, Tile.Y + reqInputItem.SlotPosY);
```
vogonistic: Avoid naming variables `tt` or similar. This should probably be `inputTile` or `inputSlotTile`.

Assets/Scripts/Models/Furniture.Factory.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            // create possible jobs for factory (hauling input)
+            // - if input slot is empty pick whatever material factory can process (that has highest priority?)
+            // - if input slot already contains some inventory, gather more from that type until stack is full
+
+
+
+            // TODO: can cause problems if there are more inputs and 1 is hanging
+            if (JobCount() > 0)
```
vogonistic: This should have brackets `{}` and probably a warning or error log if it shouldn't happen.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -20,7 +20,7 @@
 /// InstalledObjects are things like walls, doors, and furniture (e.g. a sofa).
 /// </summary>
 [MoonSharpUserData]
-public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider
+public partial class Furniture : IXmlSerializable, ISelectable, IContextActionProvider
```
vogonistic: This project is about getting people involved with game programming, C# and Unity through a tile-based base building game. Quill has stayed away from several concept that he thought would be off-putting to new users and `partial` might very well be one of them. I don't know of it being used anywhere else.

In order to start using it, we probably need to discuss the potential value.

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -277,6 +277,45 @@
...
+
+    <BuildingJob jobTime="0" />
+
+    <JobSpotOffset X="1" Y="0" />
+
+	<FactoryInfo>
+		<PossibleProductions>
+			<ProductionChain name="Iron smelting" processingTime="4">
+				<Input>
+					<Item objectType="Raw Iron" amount="3" slotPosX="0" slotPosY="0" />
```
vogonistic: `slotPosX` and `slotPosY` doesn't belong on each individual `Item`. It should probably be defined for the furniture as a whole.

Any particular reason it converts 3 for three? Is the idea that you can convert up to 48 iron for 48 steel? Or does it only take 3 items in the input?

vogonistic:
+1 on adding Copper, but that could probably have been added in a small separate PR which I'd taken in a heartbeat.

In general I like the consume/produce pattern. This looks like a pretty good improvement to the current way of doing conversions on top of that.

Assets/StreamingAssets/Data/Furniture.xml
dusho: Ye, I had it like before, but this setup enables you to have different positions of output slots for different production chain and mainly this way, it will be kept always in sync - number of items in input and slots.. (like modder can specify 2 slots in input, but places actually 3 item types as requirement for production - this can cause some headache I wanted to prevent)

vogonistic: Can you give me an example of a machine that would take input from one side for one conversion and somewhere else for another? I think it's unnecessary duplication of information. If you have a good example, but it's still a rare one, there should be a default value for the furniture and overrides other tags.

I like having the values to support conversion ratio, like it's specified on copper. (although, I'd use 1:2).

dusho: example: you have one machine that creates some chips (or something), it has 2 production chains defined:
- crappy chip (requires 2 inputs - copper wire 1x, silicon wafer 1x)
- super cool chip (requires 3 inputs - copper wire 1x, silicon wafer 2x, diamond 1x)
Not sure I get the conversion ratios.. it's explicitly set by amounts on input and output (factorio style)

vogonistic: We could also store the material inside the furniture, like in Factorio, and not have to bother with which side of the machine you input the material in. If we start having multiple inputs and store everything on different tiles, we are going to start having huge machines pretty quickly.

dusho: sure.. but what does it mean 'inside furniture' cause you don't want to hide the input materials from the user.. it will be still placed next to machine (defined input slots)
also we are using fast little dudes to place things instead of fixed belts (like in Factorio), so having fixed input slots is ok, I think
flexible input gives you more choices and in this case prevents some setup errors, and needing to specify input slot position for every item isn't such a big extra
though it can still have default input and output slots defined for furniture.. have overrides (slotPosX and Y) as you mentioned .. 
It may be harder in have generic input slots and there may be some funky situations.. like you have two input slots, you need copper and steel.. 2 dudes will accidentally pick copper and try to feed the machine.. they will both place copper (using those 2 slots), then machine will not work.. you need to sync some job with furniture that you have to clean up one of the slots for steel..

svmnotn:
@dusho this PR has some conflicts, please fix them

Assets/Scripts/Models/Furniture.cs
bjubes: although its worth noting that by having `partial` people will learn what the keyword means and how splitting up classes can be useful in a large project.

Assets/StreamingAssets/Data/Furniture.xml
vogonistic: Factorial holds the material that has been fed to the machine inside it and you have to click on it to see how much it's holding. Same thing with material dropped of at a building site in Rimworld. I think that is a better solution. 

The reason I asked about 3 inputs is because we'd need 4 tiles to put and get material from plus the sprite of the machine, making them very big. 

It's a personal preference and something that I think is worth discussing separately.

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -277,6 +277,45 @@
...
+    <PathfindingModifier>1.5</PathfindingModifier>
+    <PathfindingWeight>2</PathfindingWeight>
+    <Width>3</Width>
+    <Height>3</Height>
+
+    <BuildingJob jobTime="0" />
+
+    <JobSpotOffset X="1" Y="0" />
+
+	<FactoryInfo>
```
zwrawr: Your xml has a mix of spaces and tabs, pick one or the other :)

dusho:
made some changes.. still having problem with context menu Action not changing my params (some problem with being a copy or something and not updating real furniture params)
Added new temp art here
image

Assets/Scripts/Models/Furniture.cs
sboigelot: hem I'm not a fan either of partial, if you have to partial a class you can probably split it in two for separation of concerns.

dusho:
new commit, got rid of that partial class.. factory overall renamed to workshop.. to have class FurnitureWorkshop (instead of FurnitureFactory that can be confused with design pattern)
still having issue with context menu.. last is triggered no matter which one of workshop's context items is selected

dusho:
merged with upstream
added copper plate and renamed raw_copper to Raw Copper for consistency with other resources
there seems to be funky behavior of jobs when items are not available (job is put into waitingJobQueue and then never put into active jobQueue, even if items are available) - not related to this commit

Assets/StreamingAssets/Images/Inventory/Raw Copper.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Sprites>
+	<Sprite name="Raw Copper"     x="0" y="0" w="1" h="1" pixelPerUnit="64" />
+</Sprites>
```
Grenkin1988: Add one line at the end

Assets/StreamingAssets/Images/Inventory/Power Cell.xml
```diff
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="UTF-8" ?>
 <Sprites>
-	<Sprite name="Power Cell"     x="0" y="0" w="1" h="1" pixelPerUnit="128" />
+	<Sprite name="Power Cell"     x="0" y="0" w="1" h="1" pixelPerUnit="64" />
 </Sprites>
```
Grenkin1988: Add one line at the end

Assets/StreamingAssets/Images/Inventory/Copper Plate.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Sprites>
+	<Sprite name="Copper Plate"     x="0" y="0" w="1" h="1" pixelPerUnit="64" />
+</Sprites>
```
Grenkin1988: Add one line at the end

Assets/StreamingAssets/Images/Furniture/Metal Furnace.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Sprites>
+	<Sprite name="Metal Furnace"     			      x="0" 	y="0"  w="3" h="3" pixelPerUnit="32" />
+	<Sprite name="Metal_Furnace_Running_1"     	x="0" 	y="3" w="3" h="3" pixelPerUnit="32" />
+	<Sprite name="Metal_Furnace_Running_2"     	x="3" 	y="3" w="3" h="3" pixelPerUnit="32" />
+	<Sprite name="Metal_Furnace_Running_3"     	x="6" y="3" w="3" h="3" pixelPerUnit="32" />
+	<Sprite name="Metal_Furnace_Running_4"     	x="9" y="3" w="3" h="3" pixelPerUnit="32" />
+	<Sprite name="Metal_Furnace_Running_5"     	x="12" y="3" w="3" h="3" pixelPerUnit="32" />
+</Sprites>
```
Grenkin1988: Add one line at the end

Assets/Scripts/Models/FurnitureWorkshop.cs
```diff
@@ -0,0 +1,343 @@
...
+using System.Text;
+using System.Xml;
+using System.Xml.Serialization;
+using UnityEngine;
+
+[Serializable]
+[XmlRoot("Workshop")]
+public class FurnitureWorkshop
+{
+    #region Serialization
```
Grenkin1988: It is bad practice to use `region` to hide something

Grenkin1988:
`FurnitureWorkshop` has StyleCop errors

Assets/Scripts/Models/FurnitureWorkshop.cs
```diff
@@ -0,0 +1,343 @@
...
+                string prodChainName = chain.Name;
+                WorkshopMenuActions.Add(new WorkshopContextMenu()
+                {
+                    ProductionChainName = prodChainName,
+                    Function = (furn, prod) =>
+                    {
+                        var oldProd = furn.Parameters[CUR_PRODUCTION_CHAIN_PARAM_NAME];
+                        var isNotProcessing = furn.Parameters[CUR_PROCESSED_INV_PARAM_NAME].ToInt() == 0;
+                        // if selected production really changes and nothing is being processed now
+                        if (!prod.Equals(oldProd) && isNotProcessing)
```
Grenkin1988: `!prod.Equals(oldProd.Value)` - this is more obvious.

Assets/Scripts/Models/FurnitureWorkshop.cs
```diff
@@ -0,0 +1,343 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Xml;
+using System.Xml.Serialization;
+using UnityEngine;
```
Grenkin1988: Remove unused directives

Assets/Scripts/Models/FurnitureWorkshop.cs
```diff
@@ -0,0 +1,343 @@
...
+        // - add param to reflect factory can provide output (has output inside)
+        //   - as output will be produced after time, it is possible that output spot can be ocupied meanwhile
+        // - process for specified time
+        // - if output slot is free, provide output (if not, keep output 'inside' factory)
+
+        var curSetupChainName = furnParameters[CUR_PRODUCTION_CHAIN_PARAM_NAME].ToString();
+
+        if (!string.IsNullOrEmpty(curSetupChainName))
+        {
+            FurnitureWorkshop.ProductionChain prodChain = GetProductionChainInfo(curSetupChainName);
```
Grenkin1988: This can be -> `ProductionChain prodChain = GetProductionChainInfo(curSetupChainName);`

Assets/Scripts/Models/FurnitureWorkshop.cs
```diff
@@ -0,0 +1,343 @@
...
+        }
+    }
+
+    private void HaulingJobForInputs(FurnitureWorkshop.ProductionChain prodChain)
+    {
+        // for all inputs in production chain
+        foreach (var reqInputItem in prodChain.Input)
+        {
+            // if there is no hauling job for input object type, create one
+            var existingHaulingJob = furniture.jobs.Any(x => x.inventoryRequirements.ContainsKey(reqInputItem.ObjectType));
```
Grenkin1988: How about this:
```
Job furnitureJob;
string requiredType = reqInputItem.ObjectType;
bool existingHaulingJob = furniture.FindJobByPredicate(job => job.inventoryRequirements.ContainsKey(requiredType), out furnitureJob);
```
?

Aand add:
```
public bool FindJobByPredicate(Func<Job, bool> predicate, out Job job)
    {
    job = jobs.FirstOrDefault(predicate);
    return job != null;
}
```
to `Furniture`

Assets/Scripts/Models/FurnitureWorkshop.cs
```diff
@@ -0,0 +1,343 @@
...
+    [XmlIgnore]
+    public List<WorkshopContextMenu> WorkshopMenuActions { get; protected set; }
+
+    private Furniture furniture;
+
+    protected Parameter furnParameters { get { return furniture.Parameters; } }
+
+    public string GetDescription()
+    {
+        StringBuilder sb = new StringBuilder();
```
Grenkin1988: Like this:
```
StringBuilder sb = new StringBuilder();
string prodChain = furnParameters[CUR_PRODUCTION_CHAIN_PARAM_NAME].ToString();
sb.AppendLine(!string.IsNullOrEmpty(prodChain) ? string.Format("Production: {0}", prodChain) : "No selected production");
return sb.ToString();
```
?

Assets/Scripts/Models/FurnitureWorkshop.cs
```diff
@@ -0,0 +1,343 @@
...
+        if (PossibleProductions.Count > 1)
+        {
+            furnParameters.AddParameter(new Parameter(CUR_PRODUCTION_CHAIN_PARAM_NAME, null));
+            foreach (var chain in PossibleProductions)
+            {
+                string prodChainName = chain.Name;
+                WorkshopMenuActions.Add(new WorkshopContextMenu()
+                {
+                    ProductionChainName = prodChainName,
+                    Function = (furn, prod) =>
```
Grenkin1988: I think more readable would be change lambda by simple method. ->
```
WorkshopMenuActions.Add(new WorkshopContextMenu
                {
                    ProductionChainName = chain.Name,
                    Function = ChangeCurrentProductionChain
                });
```
and
```
private static void ChangeCurrentProductionChain(Furniture furniture, string productName)
    {
        Parameter currentProdaction = furniture.Parameters[CUR_PRODUCTION_CHAIN_PARAM_NAME];
        bool isNotProcessing = furniture.Parameters[CUR_PROCESSED_INV_PARAM_NAME].ToInt() == 0;
        // if selected production really changes and nothing is being processed now
        if (productName.Equals(currentProdaction.Value) || !isNotProcessing)
        {
            return;
        }

        furniture.CancelJobs();
        furniture.Parameters[CUR_PRODUCTION_CHAIN_PARAM_NAME].SetValue(productName);
    }
```

Assets/Scripts/Models/FurnitureWorkshop.cs
```diff
@@ -0,0 +1,343 @@
...
+                && tt.Inventory.StackSize >= reqInputItem.Amount)
+            {
+                flaggedForTaking.Add(new KeyValuePair<Tile, int>(tt, reqInputItem.Amount));
+            }
+        }
+
+        return flaggedForTaking;
+    }
+
+    private FurnitureWorkshop.ProductionChain GetProductionChainInfo(string curSetupChainName)
```
Grenkin1988: ->
```
private ProductionChain GetProductionChainByName(string productionChainName)
{
    return PossibleProductions.FirstOrDefault(chain => chain.Name.Equals(productionChainName));
}
```

Assets/Scripts/Models/FurnitureWorkshop.cs
dusho: I think I incorporated all comments, but not sure about this one..
That method in Furniture would be linq returning object (FirstOrDefault) and then checking if it's null.. It's basically what Any does
Your proposed code is exactly (if not more) wordy than original one.
I think it's cool just to have collection and let people do appropriate linqs on that..

Grenkin1988: **Main** think is than `jobs` stays private in `Furniture`

dusho: is there a reason to keep it private there? exposing collection enables linq queries for various needs (even for modders?). People will be adding their queries to furniture all the time.. is that desired?
is it that with getter on collections you are actually able to add items to it?
there is another PR #1201 that exposes only indexer to jobs

Grenkin1988: Exposing collection could lead to adding new elements or clearing it without appropriate logic.

dusho: ye.. you were faster than my edit.. I agree that public getters on collection is asking for problems in C#

dusho:
incorporated comments from Grenkin1988 (thanks for great review!)
fixed StyleCop warnings

Assets/Scripts/Models/FurnitureWorkshop.cs
```diff
@@ -279,65 +294,66 @@ private List<TileObjectTypeAmount> CheckForInventoryAtOutput(FurnitureWorkshop.P
...
-}
+    [Serializable]
+    public class ProductionChain
+    {
+        [XmlAttribute("name")]
+        public string Name { get; set; }
+        [XmlAttribute("processingTime")]
+        public float ProcessingTime { get; set; }
 
-public class WorkshopContextMenu
```
Grenkin1988: Oops!

Grenkin1988:
You removed `WorkshopContextMenu`

dusho:
sorry.. forgot to add newly created file. fixed

sboigelot:

sboigelot:
This branch break something, the character do not find job anymore: do not happen on master:

```
FilippoLeon did not find a job.
UnityEngine.Debug:Log(Object)
UberLogger.Logger:ForwardToUnity(Object, LogSeverity, Object, Object[]) (at Assets/UberLogger/UberLogger.cs:504)
Debug:ULogChannel(String, String, Object[]) (at Assets/UberLogger/Debug.cs:233)
Character:GetNewJob() (at Assets/Scripts/Models/Character.cs:608)
Character:Update_DoJob(Single) (at Assets/Scripts/Models/Character.cs:707)
Character:Update(Single) (at Assets/Scripts/Models/Character.cs:346)
World:TickEveryFrame(Single) (at Assets/Scripts/Models/World.cs:167)
WorldController:Update() (at Assets/Scripts/Controllers/WorldController.cs:162)

```

dusho:
can you describe the testcase? cause nothing different should happen for existing furniture (as it uses LUA code for update). But new code kicks in only for power cell press.. and metal furnace + there is trick with furnace that you have to pick what to smelt there in context menu (right click) , iron ore -> iron plate, copper ore -> copper plate

sboigelot:
@dusho, I just tried it again. The test case is: start the game, open the unity console and check the console :)

image


The character stay stuck at the center tile and never does anything

sboigelot:
Sorry, my bad, I'm silly, we have to order them to mine now, they won't stockpile ore by themself. This is confusing

dusho:
well actually something is indeed broken.. in my branch they are not picking to ice at all.. in master it works
I'm confused, as I did not change stockpile lua code.. and changes in Furniture.cs are related only to Workshop code.. I don't get it.. feels like some automerge messed up something somewhere, but I can't tell where

Assets/StreamingAssets/Data/Inventory.xml
```diff
@@ -1,9 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <Inventories>
     <Inventory type="Raw Iron" maxStackSize="50" basePrice="0.8" category="inv_cat_raw"/>
-    <Inventory type="Ice" maxStackSize="10" basePrice="1.2" category="inv_cat_raw"/>
```
sboigelot: this may be the issue about ice? just quick reading

dusho:
ok fixed.. 	ice got removed from inventory.xml somehow

Assets/StreamingAssets/Data/Inventory.xml
dusho: ye.. wish I saw your comment sooner :)

PR #1085, opened 2016-09-02
Added power consumption to the rest of the machines.

Tritaris:
Now all machines consume power. If power is lost, machines that create jobs will cancel them and won't remake them til power is restored. This way the characters can't use the machines when there is no power.

### DIFF
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -342,6 +342,16 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider
 
     public void Update(float deltaTime)
     {
+        if (HasPower() == false && PowerConnection.IsPowerConsumer)
+        {
+            if ( JobCount() > 0)
+            {
+                CancelJobs();
+            }
+
+            return;
+        }
+
         // TODO: some weird thing happens
         if (EventActions != null)
         {
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -342,7 +342,7 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider
 
     public void Update(float deltaTime)
     {
-        if (HasPower() == false && PowerConnection.IsPowerConsumer)
+        if (PowerConnection.IsPowerConsumer && HasPower() == false)
         {
             if ( JobCount() > 0)
             {
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -342,9 +342,9 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider
 
     public void Update(float deltaTime)
     {
-        if (PowerConnection.IsPowerConsumer && HasPower() == false)
+        if (PowerConnection != null && PowerConnection.IsPowerConsumer && HasPower() == false)
         {
-            if ( JobCount() > 0)
+            if (JobCount() > 0)
             {
                 CancelJobs();
             }
--- a/Assets/StreamingAssets/Data/Furniture.xml
+++ b/Assets/StreamingAssets/Data/Furniture.xml
@@ -176,7 +176,7 @@
     <PathfindingWeight>2</PathfindingWeight>
     <Width>3</Width>
     <Height>3</Height>
-	  <PowerConnection inputRate="1" outputRate="" capacity=""/>
+    <PowerConnection inputRate="1" outputRate="" capacity=""/>
 
     <BuildingJob jobTime="0" />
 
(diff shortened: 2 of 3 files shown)


TomMalbran:
I don't like that we need to check for the power in the update function and exit sooner, since that is the default for most of the machines. For machines that work all the time, you just use the power to determine which action to do.

Maybe we should define a Furniture property ```OnlyWorksWithPower``` so that when is true, the furniture update function we can exit sooner. The Lua function would then be called only when the Furniture is powered. We would also need to add 2 new events to the furniture that we can hook from Lua ```PowerdOn``` and ```PoweredOff``` so that if needed we can do stuff like removing the jobs.

Grenkin1988:
@TomMalbran `Furniture` has `public event Action<Furniture> IsOperatingChanged;` and `HasPower()` should work quiet fast I think.

TomMalbran:
Actually, if for the furniture where we set ```OnlyWorksWithPower = true``` we will always cancel the jobs when the furniture is powered off, and there is no reason to have the jobs still running if the furniture has power, we can just always do that. If not we could add a ```CancelJobsOnPowerOff``` property to do what it says.

@Grenkin1988 That is cool, but I was thinking more of a way to have a Lua function that is called when the furniture is powered off or on. But I am now thinking that for this it might not be required.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -342,6 +342,16 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
...
     public void Update(float deltaTime)
     {
+        if (HasPower() == false && PowerConnection.IsPowerConsumer)
+        {
+            if ( JobCount() > 0)
+            {
+                CancelJobs();
+            }
+
+            return;
```
Grenkin1988: Some `OnUpdate` still need to be called probably

TomMalbran:
It looks better, but I think that we should allow certain furniture to work without power by adding a property on the furniture ```WorksWithoutPower``` or for the opposite ```OnlyWorksWithPower = RequiresPowerToWork```. If the furniture works without power we don't check for power in the update function and call the lua, if required.

Grenkin1988:
@TomMalbran That is what `PowerConnection.IsPowerConsumer` do

TomMalbran:
It feels too messy to have to check on every lua function if there is power and return early. Like 90% of the furniture that requires power won't operate without, so it will be better to have a special case for that 10% and deal with the other 90% in the code without calling the lua function.

What we need to know is if the furniture can work without power, call the lua function, that function can check if the furniture has power and deal with the required cases. If the furniture doesn't work without power, and there is no power we remove the jobs (the jobs would only be created when there is power and we don't when those will be done), and bail early. If it has power we call the lua function, which can create the jobs again if required.

Grenkin1988:
@TomMalbran Agree. But `CanWorksWithoutPower` should be part of `Furniture` not `Connection` and it needs to read from .xml.
But `CanWorksWithoutPower` could be part of other PR

TomMalbran:
@Grenkin1988 Yes, it was I mean, ```CanWorksWithoutPower``` is part of the Furniture and set from xml.

In a way it might be part of a different PR since it will not be required here, and it will not be used for any furniture. But if we don't do it, we might not remember to add it and come with a sloppy solution. Either way is fine, but I prefer the complete solution, that we know that will be needed eventually.

Tritaris:
Fixes #1106

bjubes:
If possible, please remove all the extra tabs from the Lua files.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -342,6 +342,16 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
 
     public void Update(float deltaTime)
     {
+        if (HasPower() == false && PowerConnection.IsPowerConsumer)
```
vogonistic: I think this would be a little bit faster for failures if you check `PowerConnection.IsPowerConsumer` first

vogonistic:
Can you fix the tabs on the lines you changed in the Furniture.xml and Furniture.lua?

You can see them if you go to `Files changed` and add `?ts=1` to the url.

Tritaris:
@vogonistic Should be good to merge now.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -342,7 +342,7 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
 
     public void Update(float deltaTime)
     {
-        if (HasPower() == false && PowerConnection.IsPowerConsumer)
+        if (PowerConnection.IsPowerConsumer && HasPower() == false)
```
vogonistic: This doesn't work as well as expected because PowerConnection can be null. I very quickly get 150 warnings. You can add a check for `PowerConnection != null` or change it back.

vogonistic:
Merged!

PR #1092, opened 2016-09-02
Resolved TODO in Settings

Grenkin1988:
Resolved:
`// TODO: Make generic getSettingAs that infers return type from default value.`


vogonistic:
You have merge conflicts.

Grenkin1988:
@vogonistic Resolved conflicts

Assets/Scripts/Utilities/Settings.cs
```diff
@@ -37,31 +34,34 @@ public static class Settings
...
         {
-            settingsDict.Add(key, defaultValue);
-            SaveSettings();
+            Debug.ULogErrorChannel("Settings", "Settings Dictionary was not loaded!");
             return defaultValue;
         }
-        else
+
+        string setting = GetSetting(key, defaultValue);
+        if (setting != null && setting != defaultValue)
```
vogonistic: You should remove the check if `setting != defaultValue`. If it is the same as default value, there is no need to add it and you can safely return the value.

vogonistic: As it is, it causes the localization not to load properly for me which generates a whole string of errors.

Grenkin1988:
@vogonistic Fixed

Issue #1075, opened 2016-09-01
[Feature Todo list] Trading System


sboigelot:
Here is a list of task for the current simple trading system (rimworld like)

I'm adding it here as a reference to keep track of the progress in the feature. If you wish to handle one of the sub-tasks, don't hesitate to start on it, just post a comment here so we don't make concurrent work

- [ ] Add Currencies
- [ ] Make currencies visible somewhere on the UI
- [ ] Create a trade UI
- [ ] Disallow trade to be concluded if player/trader do not have enough money or stock
- [ ] Trade ship visit at random times (game event?)
- [ ] Trade ship visit frequency increase according to [base value]/[time played]/[renown]
- [ ] Make "trade with trader" a queuable job by selecting a character and context menu on a trade ship on a landing pad (remove auto opening of trading UI when trade ship land) (Waiting for new job system)
- [ ] Remove Context Menu Item on Landing pad to "Call a trade ship" this is just a test action
- [ ] Have a base price on item prototypes
- [ ] Have a category on item prototypes
- [ ] Allow trader prototype to have a stock prototype based on item category (keep the version based on item object type as well)
- [ ] Register TradeShip on the World (and have a trade Controller if needed)
- [ ] Save / Load TradeShip currently visiting

mikejbrown:
>Trade ship visit at random times (game event?)

This would be a perfect use case for the new Scheduler system #998 if/when that gets merged.

sboigelot:
@mikejbrown I'll have a look at it as soon as I can if no-one else pick it up before ;)

buggy213:
Do you think it would be a good idea to have some way to convert between currencies?

TomMalbran:
A few other suggestions for later:

1. Since we have a ship that lands, I think that it will make a lot of sense that when the trade is complete, we would need to move the resources we sold into the ship. Make them instantly disappear from a stockpile feels unrealistic. The trader could come with some characters that could help move the items, and they will leave once all items are in the ship. Same can happen when unloading, we could define and unloading stockpile close to the landing pad and the items we bought would be move from the ship into the stockpile.

2. We could also add a console to call other traders if we need to. We can play here with how it may work, which other traders we could call and when? do we need to pay them extra to come? are they prices more expensive since we demand them?

3. What if we could request a trader to bring certain items (which ones depend on the trader type). The amount can be random, and even whether they always bring them too, but with a high probability of having them. This items would have a higher price.

[...]

Issue #776, opened 2016-08-25
Economy (with connected story element)


koosemose:
With a game like this the ability to sell things can be very important (especially if we later implement some of the talked about ideas of having some resources possibly being rarer on different playthroughs). However, if you desperately need some resource, having to wait for a trade ship can be devastating, but being able to always sell things straight up would be too easy.

While thinking about this, it occurred to me that in the suggested storyline, we have Quill Corp as an exploitative corporation, and we could make use of that in designing the economy. If in the early game there is an easily available (maybe even part of the starter base) "Trade Station" that is solely for selling to Quill Corp, and them being exploitative they offer lower prices for your resources and things cost more, with the implication that in their "basic utility station" they have a way to easily transport the goods (so a landing pad isn't needed for it), and Mid to Late game you would want to build the more involved Landing Pad and anything else involved with that system (maybe a communications console as well), that would allow you to deal with other people (other corporations and/or free traders) who would generally offer you better prices, but only be available some of the time (similar to RimWorld's traders that come by every so often). 

I'm not sure if others will find it an interesting mechanic, but we could further reinforce the exploitative corporation idea by having trading with Quill Corp use a separate currency (A scrip like currency), the primary effect being that even late game it may be desirable to still do some amount of trade with Quill Corp so you have an emergency fund in case you need something suddenly and can't wait for a trader.

We could even go so far as having "faction standing" with different corporations and traders (though it might be better and simpler to have standing with "Free Traders" rather than individual traders.) Which could affect their prices and how often they show up, and possibly even delivery speed (maybe with a high enough faction standing you can actual call for a trader to come, so you don't have to wait as long, but still taking some time to arrive).

Dan54:
I think the "faction standing" could be made so you gain standing by trading and completing faction missions, but gaining standing with one faction makes you lose standing with another. Free trader standing could give you more frequent traders. QuillCorp standing gets you better prices, but prices are always significantly lower than free traders.

A similar system could be used for Space Pirates vs Space Police.

sboigelot:
Great idea :)

I'm also having the question about what is the currency in the game. I will need this answer to finish implementing the "free trader" trade.

In rimworld they use an Inventory item, Silver I think. But should we go that way? It always seemed strange to me. If we are in a star-trek like future, I assume we should have electronic currency.

koosemose:
[...]

Issue #1034, opened 2016-08-30
[Discussion] Switching our data format language.


koosemose:
**PLEASE VOTE YOUR CHOICE IN STRAWPOLL**
Data Format Poll : http://www.strawpoll.me/11156965

Surprisingly, we've never had a discussion directly about if we want to continue using XML as our data format or switch to something else. We've only discussed it in passing ( in #25 and #21 ). Our primary other candidate would be JSON as Unity already has some degree of support for it and further support could be achieved through [SimpleJSON](http://wiki.unity3d.com/index.php/SimpleJSON).

We have added outside libraries in the past when we needed to, so I don't think that should be a primary concern. Quills only recently stated concern was that we use one language for it, and that someone would need to volunteer to make the needed changes, stated in #25 (I don't claim that this is the entirety of his opinion on the matter, only that it is the only semi-recent statement on the matter). I know I, for one, would be willing to assist on making the changeover, if that's what's decided, and I'm confident that if we choose to do it, others will help, so I also don't think that should be a primary concern.

With JSON (via SimpleJSON) we wouldn't need to loop over the data structure, which can be error prone, such as if you skip reading a line and try to readToNextSibling from the wrong depth, you instead just check if the data you want exists and use it to populate the objects you need populated.

If interest is shown in switching to JSON we can do further investigation into what exactly it will mean for our data loading and how it will work for us.

Thoughts and opinions for or against?

Edit: An informal count of what people's preferences seem to be as of right now. This is only to give people a general idea of how people are feeling, and there are several I couldn't discern their opinion if any on the matter.

| JSON | XML | YAML |
| --- | --- | --- |
| @koosemose | @crafty-geek | @NogginBops |
| @vogonistic | @Grenkin1988 | @mikejbrown |
| @Dormanil | @gunthergun |  |
| @Mizipzor | @sboigelot |  |
| @MANDAL0R3 | @dusho |  |
| @TomMalbran | @powli |  |
| @mikejbrown | @kd7uiy |  |
| @crafty-geek |  |  |
| @Ohda |  |  |
| @mbstraus |  |  |

vogonistic:
I would love to switch to JSON!

kd7uiy:
[...]

PR #1097, opened 2016-09-02
[updated] TradeShip visit using ScheduleEvents

sboigelot:
feature #1075 

using the new schedule events from #776

This PR allow trader to come each 5 minutes as soon as the player build a landing pad. The time between visit will be variable and implemented in a separate PR.

Also add a option "DevModeOnly" to the ContextMenuAction of furnitures, to allow contextual menu action to only be visible in developer mode. This allow to still call in a trader from a landing pad while testing, but not while playing.

Created a TradeController to centralize the code about trading.


bjubes:
if the devModeOnly tag is omitted from the XML, does it default to false?

sboigelot:
@bjubes no, I don't think the reader can handle it for now and I didn't wanted to change too much as long as the discussion about XML vs JSON is still hot: #1034

bjubes:
well GetSettingAsBool is passed a default of false in your code, so if omitting tags is supported then it will default to false. If omitting tags isn't supported then that's a different problem entirely.

sboigelot:
@bjubes The context menu action builder in the furniture support it, the Xml reader of furniture prototype doesn't

vogonistic:
This PR has conflicts.

vogonistic:
I really think that DevModeOnly should have a default value of false. If the parser can't handle it, then we ought to create an issue to fix that, but it wouldn't be good if everyone needs to put in a tag for something that should only apply to pretty much no contextual actions.

sboigelot:
Merge conflict fixed + @vogonistic & @bjubes now the xml parsing of contextual menu action allow the developer to forget the DevModeOnly, the default value will be false.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -766,12 +767,16 @@ public IEnumerable<ContextMenuAction> GetContextMenuActions(ContextMenu contextM
 
         foreach (ContextMenuLuaAction contextMenuLuaAction in contextMenuLuaActions)
         {
-            yield return new ContextMenuAction
+            if (!contextMenuLuaAction.DevModeOnly ||
+                Settings.GetSettingAsBool("DialogBoxSettings_developerModeToggle", false))
```
vogonistic: This is a compile error now. The new prototype is just `GetSetting` and it infers the type form the default, so you can remove `AsBool`.

sboigelot: I will update the branch

sboigelot: done :)

sboigelot:
@vogonistic I updated the code to the new API of the Settings

bjubes:
Merged!

PR #1102, opened 2016-09-03
Jobs created by dragging over several tiles now added to the queue in…

### LINKED ISSUE #1100: Tasks should be added as a "Snake" [Easy]
When building square shapes, like a huge patch of floors, the build pattern is really inefficient. The tiles should be added as a "snake", for even rows they should go from the bottom up, for odd rows from the top down. This would make building such shapes vastly more efficient. I know the job system is going to be re-written, but it helps to queue the jobs in the right order first, which this will reduce movement time considerably.
gif

jaselito:
… a snake pattern.

resolve #1100

### DIFF
--- a/Assets/Scripts/Controllers/MouseController.cs
+++ b/Assets/Scripts/Controllers/MouseController.cs
@@ -404,7 +404,12 @@ public class MouseController
     {
         for (int x = dragParams.StartX; x <= dragParams.EndX; x++)
         {
-            for (int y = dragParams.StartY; y <= dragParams.EndY; y++)
+            // Variables for the for-loop over the y-coordinates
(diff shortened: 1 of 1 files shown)


crafty-geek:
Consider adding a comment (or appending to the comment on line 407) to explain that those vars facilitate the efficient snake pattern.  it's a little hard to intuit that from reading the code as written.

NogginBops:
@jaselito A screenshot would be appreciated.

bjubes:
Just tested this, it works great!
snake

jaselito:
Made the comment clearer as requested.  Thanks for the screencap @bjubes !

vogonistic:
Merged!

PR #1111, opened 2016-09-03
Restrict furniture functionality based on atmosphere

mbstraus:
This is for the base building feature in issue #49, for Atmosphere requirement for some buildings.

This adds some new XML tags to the Furniture.xml file to define required atmospheric gases for placement of a particular piece of furniture.  For now, it is mostly just "O2" (values can be tweaked as necessary).  I also incorporated this into the actual placement of the furniture, so that if a piece of furniture requires O2, for instance, then it will prevent you from placing it outside (or, more formally, place it on a tile where there is no O2 present).

Some things to note on some changes I had to make...
* I had to add an extra bool for the furniture placement methods, so that it could ignore placement validation in certain situations.  Since the atmosphere may not be defined at the time of placement of a particular item, if I checked it during load time, some of the furniture wasn't getting placed.  This may have a performance benefit, as when loading a world, everything should be valid (I guess unless there is save file corruption or tampering?).  May need to find a more elegant solution to this in the future.
* To make things consistent, I kept with using XmlReader.  Didn't want to mess with things too much there.

I tested out the save / load, new world, and placement of things that shouldn't be allowed to be placed outside.  Probably needs some extra eyes to make sure I didn't miss anything or whatever.


koosemose:
Regarding more elegant solution, I think in the future we may want to change a lot of it to being "atmosphere required to function, since someone could technically build a smelter in an atmosphered room, and then destroy a wall, and it's functioning just fine because it passed the original placement tests.

TomMalbran:
I think that this is a bad idea which will make building the base really tedious. I make a new room, and then I will have to wait for the gas amount to be correct to start placing furniture. Is not a good solution.

The good solution is to make it so that furniture that requires certain gases to work, will work only when those gases are present. I can place the furniture anywhere I want, but it might not start working instantly.

dusho:
Ye.. I'm with @koosemose on that.. Atmosphere shouldn't be precondition for placing the furniture, but for actually functioning of furniture

mbstraus:
Yeah, good call.  I updated it so that it should prevent operation rather than placement.

mbstraus:
I did, however, leave in the ignore code for placement position on load, as that should be unnecessary to do during load time.  If you guys want me to revert that, I can.

quill18:
Agreed that placement should be unrestricted -- but requiring an atmosphere to function for some devices makes sense.

Maybe include a red "O2" icon if furniture is missing required atmosphere?

mbstraus:
Hmm, yeah, that would be nice for that.  On initial pass on doing something like that (putting the red "O2" icon over the furniture), that seems like a larger change that probably deserves its own PR.  Perhaps I'm just not all that comfortable with the sprite controller and all of that, though.  I did add the environment status of the furniture as part of the text shown when you select the item though, as a way of observing the status of it.

Here's when the environment is bad:
image

Here's after the room has been enclosed:
image

Tritaris:
@mbstraus I can do the red "O2" for you if you want, I made the power status icon code.

mbstraus:
@Tritaris  Sure, have at it :)

Tritaris:
@mbstraus I can not able to run this PR. It gives me a NullReferenceException for the mouseController.

mbstraus:
@Tritaris Ah whoops, never re-tested after merging in the Z-axis thing.  Just had to add that to the equals method for TIle.  Should be working now.

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -193,6 +192,9 @@
 
     <LocalizationCode>furn_mining_drone_station</LocalizationCode>
     <UnlocalizedDescription>furn_mining_drone_station_desc</UnlocalizedDescription>
+    <RequiredAtmosphere>
+      <Gas hasValue="true" name="O2" />
```
TomMalbran: The order of the attributes should be name first hasValue second.

Also, we should specify what amount it requires. Having almost cero should be the same as having 0.

And, we shouldn't need to add the ```RequiredAtmosphere``` when the furniture requires no gases.

TomMalbran:
I think that this should work closer to #1085. In the update function of furniture we check if the furniture requires atmos, if it does, we check that the requirements are valid, if they aren't we remove the jobs and return early.

Assets/StreamingAssets/Data/Furniture.xml
mbstraus: The order of the attributes makes no difference what-so-ever.  Is there a particular reason why you want it this way?  I don't feel like that is a worthwhile change to make.

I can certainly add that in... I actually had that originally.

You don't need to supply RequiredAtmosphere, it is completely optional.  In fact, there are some examples of that in the file already.  The "hasValue" attribute is meant to be used for when you SHOULDN'T have that type of atmosphere.  If it doesn't matter at all, then you leave it off.

mbstraus:
I updated this to mirror the power system setup.  I also added a min requirement to the heater (just the heater).  If left off, it will default to 0.  It is using the gas amount, not the pressure, which is not currently displayed on the UI, but the flag is updating as I would expect.

mbstraus:
I thought I would give this project a whirl to see what it was like, and I think after a bit of time on it, I am going to graciously bow out.  If you want to take this PR as is, then feel free to merge it, otherwise I am going to close this out... I don't really have any more time to spend on this.

mbstraus:
Don't feel like dealing with any merge conflicts, I am closing this pull request.

PR #1112, opened 2016-09-03
Power System fix

Grenkin1988:
Now that was really embarrassing :-(


bjubes:
I'm confused as to why you used `ProjectPorcupine.Power` in your code after including it in a using statement. Shouldn't you just use on or the other? or is another namespace have a power class that is causing a conflict?

Grenkin1988:
Have no idea why, but compiler refuse to use `System` without full path in `PowerSystemTest`

JupiterJoe:
The compiler is confused between two things (System namespace and System class) that have exactly the same name.

bjubes:
yeah, in retrospect `system` is a terrible name for a class

bjubes:
maybe rename it to `powerSystem` or `powerNetwork`?

quill18:
Agreed that a rename to PowerNetwork would likely be best.

quill18:
Actually, the namespace should probably also be PowerNetwork.

Grenkin1988:
Renamed to `PowerNetwork`

bjubes:
If possible could you make your diff cleaner for PowerGridTest.cs. Its not essential if you can't, but currently git says you deleted and added 280 lines

Grenkin1988:
@bjubes I moved it to `namespace ProjectPorcupine.PowerNetwork.Test` so I think this is just git gets insane

bjubes:
Merged!

PR #1115, opened 2016-09-03
[WIP] One character / Bed

### LINKED ISSUE #1099: Sharing a single bed, especially when there's multiple
Somehow I suspect that isn't the intended behavior. I'm actually not sure how many there are, but when the time comes for sleep, the nearest unoccupied bed should be chosen, not just the nearest.

gif

Ohda:
Fixes #1099.
Caracters can now check if there is a bed availble. 
If you still see 2 characters on a bed, this is because they are waiting for a job.

Eddit : Added a first job to leave the bed when finish sleeping.


bjubes:
The many characters in a bed bug is still occurring on your branch. It seems to only happen where there are two or more beds in the world.
snake

Ohda:
You're sure your guys have something to do?

bjubes:
heres another go at it. This time I watched the other workers, and they both were both getting uranium for the power generator.
http://i.imgur.com/NW4ubTl.gif
unfortunately, selecting a character doesnt show his current job, so I can't know for sure that the second character had nothing to do, but it seems unlikely unless this is a bug in the job system.

Ohda:
The easy way to be sure is to check the console if there is any 'character did not find  a job' message.

bjubes:
Im going to assume that the character did not find a job, either way this PR fixes the bug for the most part, but there are still exceptions when a character has nothing else to do.

sboigelot:
Maybe we should automatically force a character that finish using the bed to move 1 or two tiles. So they leave the bed

Assets/Scripts/Models/FurnitureActions.cs
```diff
@@ -48,23 +48,11 @@ public static void LoadModsScripts(DirectoryInfo[] mods)
         }
     }
 
-    public static void CallFunctionsWithFurniture<T>(string[] functionNames, T furn, float deltaTime)
+    public static void CallFunctionsWithFurniture<T>(string[] functionNames, params object[] args)
```
Grenkin1988: Why this method is generic? What `<T>` suppose to do?

TomMalbran:
Should we generalize this code? Instead of using params and lu functions, we can use furniture properties and prevent the use inside c#. A lua callback can be used if required. I am thinking that something like this can be used on other furniture, like chairs, sofas... If we go with the lua approach, we will have code duplication.

BTW. If we move this feature completely to the furniture, it can be good to implement it as a component instead of directly inside the furniture.

Assets/Scripts/Models/Character.cs
```diff
@@ -591,11 +591,14 @@ private void GetNewJob()
             }
         }
 
-        if (needPercent > 50 && needPercent < 100 && need.RestoreNeedFurn != null)
+        if (needPercent > 50 && needPercent <= 100 && need.RestoreNeedFurn != null)
         {
-            if (World.Current.CountFurnitureType(need.RestoreNeedFurn.ObjectType) > 0)
+            IEnumerable<Furniture> furnitures = World.Current.furnitures.Where(furniture => furniture.ObjectType == need.RestoreNeedFurn.ObjectType && furniture.IsUsable() == true);
+            if (furnitures.Count() > 0)
```
Grenkin1988: `furnitures.Count() > 0` should be replaced by `furnitures.Any()`

Assets/Scripts/Models/Character.cs
```diff
@@ -591,11 +591,14 @@ private void GetNewJob()
             }
         }
 
-        if (needPercent > 50 && needPercent < 100 && need.RestoreNeedFurn != null)
+        if (needPercent > 50 && needPercent <= 100 && need.RestoreNeedFurn != null)
         {
-            if (World.Current.CountFurnitureType(need.RestoreNeedFurn.ObjectType) > 0)
+            IEnumerable<Furniture> furnitures = World.Current.furnitures.Where(furniture => furniture.ObjectType == need.RestoreNeedFurn.ObjectType && furniture.IsUsable() == true);
```
Grenkin1988: `IEnumerable` is just a query. It is executed each time you try to iterate with it.
Usually it is bad when you have to iterate through `IEnumerable` several times. This is because list of object could change between iterations and this would lead to situation when you iterate with two different lists.
If you need to use `IEnumerable` several time it is good idea to execute by `.ToList()` `.ToArray()` call

Assets/Scripts/Models/Character.cs
```diff
@@ -1081,4 +1077,22 @@ private void OnInventoryCreated(Inventory inv)
             }
         }
     }
-}
\ No newline at end of file
+
+    private Furniture GetClosestFurniture(IEnumerable<Furniture> furnitures)
+    {
+        int bestLength = 0;
```
Grenkin1988: `int bestLength = int.MaxValue;`

Assets/Scripts/Models/Character.cs
```diff
@@ -1081,4 +1077,22 @@ private void OnInventoryCreated(Inventory inv)
...
+        Furniture destFurniture = new Furniture();
+
+        foreach (Furniture furniture in furnitures)
+        {
+            Path_AStar path = new Path_AStar(World.Current, currTile, furniture.Tile);
+            if (bestLength == 0 || path.Length() < bestLength)
```
Grenkin1988:
```
if (path.Length() >= bestLength)
    {
        continue;
    }

bestLength = path.Length();
destFurniture = furniture;
```

Grenkin1988:
@Ohda This is cosmetic that I found. ^
Regarding parameters and lua use from this.
I would say that we need to use C# here as much as possible. Definitely we would have `Furniture` like bed in future.
Bu you should ask more people about this.

Assets/Scripts/Models/Character.cs
```diff
@@ -1077,20 +1077,21 @@ private void OnInventoryCreated(Inventory inv)
             }
         }
     }
-
-    private Furniture GetClosestFurniture(IEnumerable<Furniture> furnitures)
+    
+    private Furniture GetClosestFurniture(List<Furniture> furnitures)
```
Grenkin1988: Here you could except IEnumerable as before

Assets/Scripts/Models/Character.cs
```diff
@@ -593,8 +593,8 @@ private void GetNewJob()
 
         if (needPercent > 50 && needPercent <= 100 && need.RestoreNeedFurn != null)
         {
-            IEnumerable<Furniture> furnitures = World.Current.furnitures.Where(furniture => furniture.ObjectType == need.RestoreNeedFurn.ObjectType && furniture.IsUsable() == true);
-            if (furnitures.Count() > 0)
+            List<Furniture> furnitures = World.Current.furnitures.Where(furniture => furniture.ObjectType == need.RestoreNeedFurn.ObjectType && furniture.IsUsable() == true).ToList();
+            if (furnitures.Any())
```
Grenkin1988: If this is `List<>` it is better to use `.Count` property of list instead of LINQ method.

Ohda:
I don't find solutions to handle thoses behaviours properly for now. I close the PR.
Feel free to try things. It will be easier to wait for the new job and pathfinding system.

PR #1118, opened 2016-09-04
Fixes NullReferenceException in Metal Smelter

Tritaris:
Fixes 'NullReferenceException: Object reference not set to an instance of an object` . When the metal smelter tries to spawn the steel plates it can't find the location. Variable was misspelled.


vogonistic:
Merged! Good catch.

PR #1119, opened 2016-09-04
[HotFix] Fixes #1117

### LINKED ISSUE #1117: [Bug] After loading a save, cannot place any furniture
This looks to be only a problem after loading a save.  Regardless of where you try to place a new piece of furniture, the system prevents you, citing that there is no floor tile underneath.

image

That was off of the branch that I'm currently working in, which I didn't touch anything with regards to this... however I switched to my master and got it up to date, and I am actually just straight up unable to load anything.  It just starts spewing null reference exceptions after loading up a save.  Is anyone else seeing this or am I just going crazy?

koosemose:
Z and Depth not Properly saved - Fixed
Cached Tiletypes causing innacurate comparisons - Fixed

### DIFF
--- a/Assets/Scripts/Models/TileType.cs
+++ b/Assets/Scripts/Models/TileType.cs
@@ -52,12 +52,12 @@ public class TileType : IXmlSerializable
     // Just two util functions to not break every link to TileType.(Empty | Floor)
     public static TileType Empty 
     {
-        get { return tileTypeDictionary["Empty"]; } 
+        get { return empty ?? (empty = tileTypeDictionary["Empty"]); } 
     }
 
     public static TileType Floor 
     {
-        get { return tileTypeDictionary["Floor"]; } 
+        get { return floor ?? (floor = tileTypeDictionary["Floor"]); }
     }
 
     public string Type { get; protected set; }
--- a/Assets/Scripts/Models/TileType.cs
+++ b/Assets/Scripts/Models/TileType.cs
@@ -52,12 +52,12 @@ public class TileType : IXmlSerializable
     // Just two util functions to not break every link to TileType.(Empty | Floor)
     public static TileType Empty 
     {
-        get { return empty ?? (empty = tileTypeDictionary["Empty"]); } 
+        get { return tileTypeDictionary["Empty"]; } 
     }
 
     public static TileType Floor 
     {
-        get { return floor ?? (floor = tileTypeDictionary["Floor"]); }
+        get { return tileTypeDictionary["Floor"]; } 
     }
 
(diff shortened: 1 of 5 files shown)


koosemose:
I have tried to resolve this in a way that doesn't roll back #1093 , but cannot figure out how to do so, aside from completely sidestepping accessing where it is cached.

Grenkin1988:
Do not get why you need to undo cache of tyleType

koosemose:
Because it prevents loading of save files.

It causes this line ` if (t2.Type != TileType.Floor && tileTypeBuildPermissions.Contains(t2.Type) == false)` (828 in furniture.cs) to fail incorrectly. When t2.Type is TileType.Floor, != still returns true. I attempted to change it to !t2.Type.Equals(TileType.Floor) but it still doesn't return what you would expect. It seems to only do this when loading from a save file. I am uncertain how to fix this without rolling back the change.

I tagged your merge in the comment to call you attention, so you can attempt to find a proper fix, but I just wanted to get a fix that would allow loading out as soon as possible.

Grenkin1988:
I will look at this in one hour

koosemose:
If you work on a fix on master, be aware that there is a second bug that also causes saves to not load properly, that's a few places where Z coordinate is not written to save file, and depth as well. Feel free to reference them and include in your own fix if needed to debug, and/or submit PR covering that as well.

Grenkin1988:
@koosemose I've fixed `TileType` issue. What should I do? New PR?

koosemose:
Probably best, I am not at my computer to include in this fix in any way, and won't be for 8+ hours.

mattpott:
Maybe do a pull request to his own branch which would get he can merged.

koosemose:
I may be able to accept a PR from mobile, if it can be done shortly. Otherwise I won't be able to until much later

Grenkin1988:
@koosemose https://github.com/koosemose/ProjectPorcupine/pull/1

Assets/Scripts/Models/TileType.cs
```diff
@@ -57,7 +57,7 @@ public static TileType Empty
 
     public static TileType Floor 
     {
-        get { return floor ?? (floor = tileTypeDictionary["Floor"]); }
+        get { return floor ?? (floor = tileTypeDictionary["Floor"]); }
```
NogginBops: There is a trailing white space here. It should probably be removed. :)

Grenkin1988: I am in progress of cleanup of TileType.

vogonistic:
Merged!

PR #1121, opened 2016-09-04
FilterMode Point for SpriteManager-loaded sprites

Dormanil:
Set the FilterMode to Point for textures loaded in SpriteManager to make them look nicer (not washed out) when zoomed in.

PSA: I didn't change the default ppu for files without XML, because I honestly think that we should never have decided to do that, especially when everyone in favour of 32x32 (majority of the Sprite Crew on Discord) was not even there.


Raketfart:
:+1:

sboigelot:
much crispier :D nice!

PR #1123, opened 2016-09-04
Fixed Wall sprites

Dormanil:
I mixed up some sprites and made one that was entirely wrong, but this little fix should make them okay again.


Dormanil:
Before: [!snippet1](http://puu.sh/qZC3W/ece8c0f958.jpg)
After: [!snippet2](http://puu.sh/qZCtZ/777397ba26.jpg)

TomMalbran:
I see a couple of lines in between the walls like on row 3 column 12, and on row 6 column 12 (each tile is a cell). Is that an issue with the sprite or with rendering?

bjubes:
Im gonna hold of on this one for now due to @TomMalbran comment and #1142

Dormanil:
That is a rendering issue, the sprite itself is clean from those lines. 

And the fixes to the XML file and to the spritesheet are favourable regardless of #1142. Imagine someone implements the changes discussed in #1142 but doesn't notice the errors I fixed? Then we have nicer looking sprites indeed, but the error in XML and Sprite is still there. It's just safer to fix it now than to hope that the next contributor notices the errors.

TomMalbran:
If this is a rendering issue, then is fine for now. We should still merge it to fix the XML issue. If needed we can do a new issue/pr for the other stuff.

mikejbrown:
Going ahead and merging for the fixes. The sprite can be changed later.

PR #1124, opened 2016-09-04
TileType class refactoring

Grenkin1988:
Made some refactoring for TileType


bjubes:
why did u change the spacing in Tiles.xml, or was that accidental?

Grenkin1988:
A can undo this change to Tiles.xml, but that is how file looks like after autoformat.

bjubes:
I don't think we have a standard for amount of spaces to indent in XML, so your probably fine

Assets/Scripts/Models/TileType.cs
```diff
@@ -6,87 +6,78 @@
...
     /// </summary>
-    public float PathfindingModifier
-    {
-        get { return pathfindingModifier; }
-        set { pathfindingModifier = value; }
-    }
+    public float PathfindingModifier { get; set; }
 
-    // TODO!
     public bool LinksToNeighbours { get; protected set; }
```
TomMalbran: Should you use private here, and in the next few properties? Is changed above from protected.

Assets/Scripts/Models/TileType.cs
```diff
@@ -207,104 +172,64 @@ public override int GetHashCode()
...
+        XmlReader reader = parentReader.ReadSubtree();
         while (reader.Read())
         {
             switch (reader.Name)
             {
                 case "Name":
-                    reader.Read();
-                    Name = reader.ReadContentAsString();
-                    break;
+                reader.Read();
```
TomMalbran: I thought that we were doing swicthes like:

```
switch (reader.Name)
{
    case "Name":
        // Code goes here
        break;
    case "Description":
        // ...
        break;
}
```

It was already changed in furniture from a different style.

PR #1125, opened 2016-09-04
Fix folding in Travis log

bjubes:
fixes the order of echos in travis so the log folds on headings rather than the line after

before:
before
After:
after

any messages about failures still are shown after the Unit Tests section and will not be folded.


bjubes:
If possible, merge #1138 first as this is a much smaller change so if they conflict, fixing this PR will be much easier.

vogonistic:
Merged!

PR #1127, opened 2016-09-04
[Updated] Keyboard mapping

sboigelot:
I'm one of those guys using a different keyboard layout then the rest of the world :)

Yes I'm using AZERTY, not even the classic one, but the French-Belgium-Point variation. A sub variation of a sub variation of something different then QWERTY

As it, the current implementation is not allowing me to move around with WASD or my localized version QZSD.

This will open the gate for player customization of the key used in the game.

For now, I do not save/load the keyboard layout from the settings, as I'm waiting for the discussion about data format to land.

Controllers are responsible to map action to inputs names
The keyboard manager map inputs names to keycodes (or later will load this list from a file)


TomMalbran:
I think that at this point we should start thinking of a KeyboardManager that holds all the shortcuts. We can then add shortcuts from other modules (the KeyboardManager shouldn't know about all the modules), where you pass a name, the shortcut and a callback (i.e. for ```KeyboardManager.RegisterShortcut("IncreaseTimeScale", [ KeyCode.Plus, KeyCode.KeypadPlus ], IncreaseTimeScale)```). The KeyboardController can then ask the KeyboardManager to execute a shortcut with the input it received.

The mapping could be loaded from an xml file (different from the settings one) from the KeyboardManager and it will replace the registered keys.

We probably should switch KeyCode for strings so that we can load them from an xml.

sboigelot:
@TomMalbran I implemented your suggestion. Except for the load/save from xml for now as I'm waiting for the output of #1034

Save/load will be another PR

TomMalbran:
@sboigelot This looks good. One question though. How will you handle the default shortcuts vs the user ones?

In the way I see it, there should be an easy way for us and for modders to set the Keyboard input combination for the shortcut. We can then have a ui where the user can map the shortcuts to a different Keyboard input combination or to none.

This could probably be done in 2 ways. We pass the Keyboard input combination when registering a shortcut, or each mod will have an xml file with all the shortcuts. and the Keyboard Manager will load have to load those. Maybe using the first option is easier?

We would then have a different file with the Keyboard input combination changed by the user. This list will be loaded after and the Keyboard input combination will be changed.

sboigelot:
@TomMalbran for now this works great, we can always improve it according to our needs. I'm trying to keep my PR smaller in order to not loose a lot of work if the PR is rejected.

This is the basic functionality, and it's working. If it's accepted we will all be free to build on top of it.

bjubes:
Awesome! I'm glad were working towards remapping keys since that's one of the main features of every game that unity has no support for.

PR #1129, opened 2016-09-04
UberLogger can detect that he called from Unit Tests

Grenkin1988:
Some UnitTest are failing when running them from IDE (VisualStudio) with exception:
```
System.TypeInitializationException: initializer type "UberLogger.Logger" threw an exception.
  ----> System.Security.SecurityException: ECall methods must be packaged in a system module.
   in UberLogger.Logger.Log (String channel, Object source, LogSeverity severity, Object message, Object [] par)
   in Debug.ULogChannel (String channel, String message, Object [] par) in D: \ Development \ GitHubRepository \ ProjectPorcupine \ Assets \ UberLogger \ Debug.cs: line 127
   in ScheduledEventTest <Init> b__2_0 (ScheduledEvent evt) in D:. \ Development \ GitHubRepository \ ProjectPorcupine \ Assets \ Editor \ ScheduledEventTest.cs: line 26
   in Scheduler.ScheduledEvent.Fire () in D: \ Development \ GitHubRepository \ ProjectPorcupine \ Assets \ Scripts \ Scheduler \ ScheduledEvent.cs: line 160
   in Scheduler.ScheduledEvent.Update (Single deltaTime) in D: \ Development \ GitHubRepository \ ProjectPorcupine \ Assets \ Scripts \ Scheduler \ ScheduledEvent.cs: line 145
   in ScheduledEventTest.CooldownEventRunTest () in D: \ Development \ GitHubRepository \ ProjectPorcupine \ Assets \ Editor \ ScheduledEventTest.cs: line 141
--SecurityException
   in UnityEngine.Application.add_logMessageReceived (LogCallback value)
   in UberLogger.Logger..cctor () in D: \ Development \ GitHubRepository \ ProjectPorcupine \ Assets \ UberLogger \ UberLogger.cs: line 205
```
This "fix" will prevent call to `Log` if running from UTest.

PR #1132, opened 2016-09-04
Fix SchedulerEditorTests using PrototypeManager

mikejbrown:
Fixes issue mentioned by @Grenkin1988 @vogonistic in #998.

It is now possible to use Scheduler without PrototypeManager for unit testing purposes. I did have to make `LuaUtilities.LoadScript` public for this to work. Any objections?

Apologies for slow response... I went offline for the weekend and this happened. :cry:

### DIFF
--- a/Assets/Editor/SchedulerEditorTest.cs
+++ b/Assets/Editor/SchedulerEditorTest.cs
@@ -7,6 +7,7 @@
 // ====================================================
 #endregion
 using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.IO;
 using System.Text;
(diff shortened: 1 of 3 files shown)


TomMalbran:
Maybe a better solution would be to change the PrototypeManger so that it won't load the files when creating it. Eventually the ModManager will load all the files, but for now the PrototypeManager can do it.

My idea is in the ```XmlPrototypes``` constructor remove ```LoadPrototypesFromFile();``` and then create a method in PrototypesManager like:
```
    public void LoadPrototypesFromFile()
    {
        Furniture.LoadPrototypesFromFile();
        Inventory.LoadPrototypesFromFile();
        Need.LoadPrototypesFromFile();
        Trader.LoadPrototypesFromFile();
        Quest.LoadPrototypesFromFile();
        Stat.LoadPrototypesFromFile();
        SchedulerEvent.LoadPrototypesFromFile();
    }
```

You can then use ```PrototypeManager.SchedulerEvent.Add()``` to add the prototypes.

TomMalbran:
Or have the function ```LoadModPrototypesFromFile``` load all the files.

I could make a PR with this change so that you can still use the PrototoypeManager and add your prototoypes manually.

Grenkin1988:
ULog should not be used from tests.
It makes tests to fail when you run them from IDE. (Not from Unity)
Here I am fixing this issue #1129  (fixing - **disable Log from test run**)

mikejbrown:
Self-closed. See discussion on #1146.

PR #1133, opened 2016-09-04
Mod/Math/Lua Utils Unit tests

zwrawr:
Adds some unit tests for these three classes.
All tests currently pass


bjubes:
The more unit tests the merrier!

PR #1134, opened 2016-09-04
[Conflicts] Renamed EventAction to EventActions - makes a lot more sense.

longtomjr:


TomMalbran:
I am moving all the actions into an action manager, so that we don't need to keep repeating code in the Actions classes. So this change will not be needed soon.

koosemose:
@TomMalbran EventAction(s) isn't the same as the other Actions Classes, it's for events to be fired in the various Object classes (not sure what to refer to them as a group, but Inventory, Furniture, Tile, and so on, though it is currently written to only accept Furniture as it's reference to pass on to Lua and needs made generic rather than all the other's having to roll their own events.

TomMalbran:
@koosemose Right. I was thinking of GameEventActions.
Make a separate issue to make this class a generic one. That issue should be solved in its own PR and not on other that deal with a different problem. Since it is used only for Furniture it makes sense that it works like that, if you need it for something else, then you can make it generic.

I am ok with the EventAction name, even though it can several actions.

Grenkin1988:
Actually class is already `EventActions`

PR #1136, opened 2016-09-04
[Stylecop violations] [Documentation]The complete Documentation for Furniture.cs

longtomjr:
This also contains  #1134

This is not perfect, but it is a hell of an improvement


Assets/Scripts/Models/Furniture.cs
```diff
@@ -337,6 +415,10 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
         return obj;
     }
 
+    /// <summary>
+    /// The update function used by unity. Updates every frame
+    /// </summary>
+    /// <param name="deltaTime">The time since the last update was called.</param>
     public void Update(float deltaTime)
```
Grenkin1988: This Update method is not used by Unity.

longtomjr: Aaaarg fudge. I relized that, i was just busy with multiple stuff. Feel
free to edit it. I am in bed now, if not fixed, i will fix it tomorrow
morning.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -701,43 +843,75 @@ public void Deconstruct()
...
+    /// </summary>
+    /// <param name="typeTag">Tag to check for</param>
+    /// <returns>True if furniture has specified tag</returns>
     public bool HasTypeTag(string typeTag)
     {
         return typeTags.Contains(typeTag);
     }
 
+    /// <summary>
+    /// Returns the name of the furniture.
```
Grenkin1988: This is not true. It returns `LocalizationCode` as I see.

longtomjr:
Thanks for pointing out this stuff. If someone else does not fix this, i will by morning

Assets/Scripts/Models/Furniture.cs
```diff
@@ -244,28 +291,53 @@ public List<string> ReplaceableFurniture
...
+    /// </summary>
     public int Height { get; private set; }
 
+    /// <summary>
+    /// Gets the code used for Localization of the furniture.
+    /// </summary>
     public string LocalizationCode { get; private set; }
 
+    /// <summary>
+    /// Gets the description of the furniture if there is no localization support.
```
Grenkin1988: Here is `UnlocalizedDescription` for door from xml `<UnlocalizedDescription>furn_door_desc</UnlocalizedDescription>`

longtomjr:
Thanks, will look into that in the morn, feel free to fix btw!

Assets/Scripts/Models/Furniture.cs
```diff
@@ -244,28 +291,53 @@ public List<string> ReplaceableFurniture
...
+    /// </summary>
     public string UnlocalizedDescription { get; private set; }
 
+    /// <summary>
+    /// Gets a value indicating whether this furniture is next to any furniture of the same type.
```
TomMalbran: The description is not right here. LinksToNeighbour is used to decide which sprite to use depending on which furniture are next to this one with the same object type.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -357,6 +443,10 @@ public bool IsExit()
         return false;
     }
 
+    /// <summary>
+    /// Checks if the furniture can be Entered
+    /// </summary>
+    /// <returns>Yes if furniture can be entered</returns>
```
TomMalbran: It returns ```Enterability``` which has the values of Yes, No, Soon.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -371,6 +461,11 @@ public Enterability IsEnterable()
         return (Enterability)ret.Number;
     }
 
+    /// <summary>
+    /// Check if the furniture has a unique name and return the name if it has one.
+    /// If the furniture has not unique name, it returns the object type
+    /// </summary>
+    /// <returns>Name of the furniture</returns>
```
TomMalbran: Is the name of the sprite

Assets/Scripts/Models/Furniture.cs
```diff
@@ -371,6 +461,11 @@ public Enterability IsEnterable()
         return (Enterability)ret.Number;
     }
 
+    /// <summary>
+    /// Check if the furniture has a unique name and return the name if it has one.
+    /// If the furniture has not unique name, it returns the object type
```
TomMalbran: it kind of checks if the furniture has a function to determine the sprite name and calls it to get it.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -382,22 +477,39 @@ public string GetSpriteName()
         return ret.String;
     }
 
+    /// <summary>
+    /// Check if the position of the furniture is valid or not
```
TomMalbran: we should add that is called when placing the furniture. It makes no sense to call it once is is placed.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -382,22 +477,39 @@ public string GetSpriteName()
...
+    /// </summary>
+    /// <returns>True if the furniture has power</returns>
     public bool HasPower()
     {
         IsOperating = PowerConnection == null || World.Current.PowerSystem.HasPower(PowerConnection);
         return IsOperating;
     }
 
+    /// <summary>
+    /// This does absolutely nothing
```
TomMalbran: Add that this is required to implement ```IXmlSerializable```

Assets/Scripts/Models/Furniture.cs
```diff
@@ -554,6 +670,10 @@ public void ReadXmlPrototype(XmlReader readerParent)
         }
     }
 
+    /// <summary>
+    /// Reads the specified XMLReader (pass it to <see cref="ReadXmlParams(XmlReader)"/>)
```
TomMalbran: This is used to load a furniture from a save file.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -580,11 +704,19 @@ public Parameter GetParameters()
         return furnParameters;
     }
 
+    /// <summary>
+    /// How many jobs is linked to this furniture
```
TomMalbran: is -> are

Assets/Scripts/Models/Furniture.cs
```diff
@@ -602,6 +737,10 @@ public void CancelJobs()
         }
     }
 
+    /// <summary>
+    /// Checks whether the furniture is a stockpile
+    /// </summary>
+    /// <returns>True is the furniture is a stockpile</returns>
     public bool IsStockpile()
```
TomMalbran: We should probably remove this function since is very tied to a single type of furniture.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -143,15 +146,25 @@ private Furniture(Furniture other)
         UnlocalizedDescription = other.UnlocalizedDescription;
     }
 
+    /// <summary>
+    /// This event will trigger when the furniture has been changed.
```
sboigelot: What mean "changed"? Replaced by another one?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -143,15 +146,25 @@ private Furniture(Furniture other)
...
     public event Action<Furniture> Removed;
 
+    /// <summary>
+    /// This event will trigger if <see cref="IsOperating"/> has been changed.
+    /// </summary>
     public event Action<Furniture> IsOperatingChanged;
 
     /// <summary>
-    /// Gets or sets the Furniture's pathfinding modifier which is added into the Tile's final PathfindingCost.
+    /// Gets or sets the Furniture's <see cref="PathfindingModifier"/> which is added into the Tile's final PathfindingCost.
```
sboigelot: Only one tile? what if the furniture span over multiple tiles?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -161,35 +174,51 @@ public float PathfindingModifier
...
-    // If this furniture gets worked by a person,
-    // where is the correct spot for them to stand,
-    // relative to the bottom-left tile of the sprite.
-    // NOTE: This could even be something outside of the actual
-    // furniture tile itself!  (In fact, this will probably be common).
+    /// <summary>
+    /// Gets the spot where the astronaut will stand when he is using the furniture. This is relative to the bottom
+    /// left tile of the sprite. This can be outside of the actual furniture.
+    /// </summary>
+    /// <value>The spot where the astronaut will stand when he uses the furniture.</value>
```
sboigelot: Please replace astronaut by a link to Character, astronaut isn't a thing in the codebase

Assets/Scripts/Models/Furniture.cs
```diff
@@ -161,35 +174,51 @@ public float PathfindingModifier
...
     public Vector2 JobSpotOffset { get; private set; }
 
-    // Flag for Lua to check if this is a vertical or horizontal door and display the correct sprite.
+    /// <summary>
+    /// Gets or sets a value indicating whether the door is Vertical or not.
+    /// Should be false if the furniture is not a door.
+    /// </summary>
+    /// <value>Whether the door is Vertical or not.</value>
     public bool VerticalDoor { get; set; }
```
sboigelot: Documentation should probably indicate that this field will most probably be removed/move to another class soon.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -161,35 +174,51 @@ public float PathfindingModifier
...
 
     /// <summary>
+    /// Gets the EventAction for the current furniture.
     /// These actions are called when Trigger is called. They get passed the furniture
```
sboigelot: I'm not sure I understand the following
     /// These actions are called when Trigger is called. They get passed the furniture
      /// they belong to, plus a deltaTime (which defaults to 0)

Assets/Scripts/Models/Furniture.cs
```diff
@@ -209,15 +238,29 @@ private set
...
+    /// <value>Whether the furniture is selected or not.</value>
     public bool IsSelected { get; set; }
 
-    // This represents the BASE tile of the object -- but in practice, large objects may actually occupy
-    // multile tiles.
+    /// <summary>
+    /// Gets the BASE tile of the furniture. (Large objects can span over multiple tiles).
+    /// </summary>
+    /// <value>The BASE tile of the furniture.</value>
     public Tile Tile { get; private set; }
```
sboigelot: Shoudn't we rename this to 
public Tile BaseTile { get; private set; }
?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -236,6 +279,10 @@ private set
         }
     }
 
+    /// <summary>
+    /// Gets a list of furniture this furniture can be replaced with.
+    /// </summary>
+    /// <value>A list of furniture that this furniture can be replaced with.</value>
```
sboigelot: A list of furniture **ObjectType** that this furniture can be replaced with.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -244,28 +291,53 @@ public List<string> ReplaceableFurniture
...
 
-    // This is a multipler. So a value of "2" here, means you move twice as slowly (i.e. at half speed)
-    // Tile types and other environmental effects may be combined.
-    // For example, a "rough" tile (cost of 2) with a table (cost of 3) that is on fire (cost of 3)
-    // would have a total movement cost of (2+3+3 = 8), so you'd move through this tile at 1/8th normal speed.
-    // SPECIAL: If movementCost = 0, then this tile is impassible. (e.g. a wall).
+    /// <summary>
+    /// Gets the movement cost multiplier that this furniture has. 1 is default, 2 is twice as slow.
+    /// Tile types and environmental effects will be combined with this value (additive).
+    /// If this value is '0' then the furniture is impassable.
```
sboigelot: Is this limited to 0,1,2 or can it have any value?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -279,6 +351,12 @@ private set
         }
     }
 
+    /// <summary>
+    /// Used to place furniture in a certain position
+    /// </summary>
+    /// <param name="proto">The prototype furniture to place</param>
+    /// <param name="tile">The tile to place the furniture on</param>
```
sboigelot: The **base** tile to place the furniture on, **The tile will be the bottom left corner of the furniture** (to check)

Assets/Scripts/Models/Furniture.cs
```diff
@@ -382,22 +477,39 @@ public string GetSpriteName()
         return ret.String;
     }
 
+    /// <summary>
+    /// Check if the position of the furniture is valid or not
+    /// </summary>
+    /// <param name="t">The tile to check</param>
```
sboigelot: this again, is the base tile, it will check more then the base tile

Assets/Scripts/Models/Furniture.cs
```diff
@@ -580,11 +704,19 @@ public Parameter GetParameters()
         return furnParameters;
     }
 
+    /// <summary>
+    /// How many jobs is linked to this furniture
+    /// </summary>
+    /// <returns>A number</returns>
```
sboigelot: <returns>the number of jobs linked to the furniture</returns>

Assets/Scripts/Models/Furniture.cs
```diff
@@ -580,11 +704,19 @@ public Parameter GetParameters()
...
     public int JobCount()
     {
         return jobs.Count;
     }
 
+    /// <summary>
+    /// Adds a job linked to the current furniture.
```
sboigelot: The job isn't link before

"Link a job to the current furniture"

Assets/Scripts/Models/Furniture.cs
```diff
@@ -701,43 +843,75 @@ public void Deconstruct()
...
+    /// Gets the tile that is used to do a job.
+    /// </summary>
+    /// <returns>Tile that is used for jobs</returns>
     public Tile GetJobSpotTile()
     {
         return World.Current.GetTileAt(Tile.X + (int)JobSpotOffset.x, Tile.Y + (int)JobSpotOffset.y, Tile.Z);
     }
 
+    /// <summary>
+    /// Gets the tile that is used to spawn new objects or spacemen
```
sboigelot: spacemen is not a thing, use Character

sboigelot: objects is probably incorrect too, is it Inventory?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -701,43 +843,75 @@ public void Deconstruct()
...
+    /// Returns the name of the furniture.
+    /// </summary>
+    /// <returns>name of the furniture</returns>
     public string GetName()
     {
         return LocalizationCode; // this.Name;
     }
 
+    /// <summary>
+    /// Returns the description of the furniture
```
sboigelot: /// Returns the UnlocalizedDescription of the furniture

Assets/Scripts/Models/Furniture.cs
TomMalbran: It is called when certain properties values change. Maybe all? Will be nice to add that there.

sboigelot: Should we rename it PropertyChanged then? Follow the INotifyPropertyChanged interface?

Assets/Scripts/Models/Furniture.cs
TomMalbran: Each tile has a reference that points to the same instance. So if a tile wants the ```PathfindingModifier``` it gets its value, and if it changes it it changes it for all.

I think that the description is fine since the MovementCost is per tile.

Assets/Scripts/Models/Furniture.cs
TomMalbran: it can be any thing. The "1 is default, 2 is twice as slow." is more of an example of how it works.

Assets/Scripts/Models/Furniture.cs
TomMalbran: I would go with "objects (i.e. Inventory, Character)", since in the future Mods can use it for anything and we don't what it will be used for.

Assets/Scripts/Models/Furniture.cs
sboigelot: I disagree, if a description require explanation for it to be understood, the description is not complete. We should add that explanation in.

Assets/Scripts/Models/Furniture.cs
sboigelot: It's a scoped example. It limit the understanding and seems obvious reading the code, but the goal of the documentation is to explain it easily to someone not reading the code.

We should really add that it can be any float value from 0 to any positive number. Negative are not valid, and 0 is a special case

Assets/Scripts/Models/Furniture.cs
TomMalbran: I don't think that we need to explain here how placing a furniture works here. Maybe the issue is that the Tile that is referenced in the comment is not actually the Tile of the Furniture, but the tile that called this function.

sboigelot: Agreed, however this is a property and nothing forbid something else then a Tile to use it.
It may be used later on by something else. Tying the comment to the unspecific caller is not a great idea neither. Should we describe what the property means instead of what it is currently used for ?

Assets/Scripts/Models/Furniture.cs
TomMalbran: That will not work since then you will think that it is for a particular property and not for any. I think that changed is fine as a name, since a new name would have to enclose many possible changed events. Like sometimes changed can be called because a neighbour changed, which would make the sprite change (but no property of the furniture changed). You can also trigger it from Lua, without any change. The description could tell a bit more. For now it is mostly used to update the sprite.

Assets/Scripts/Models/Furniture.cs
TomMalbran: What I mean is that is the description you add something like "(i.e. 1 is the default so 2 is twice as slow)", and what you said.

Assets/Scripts/Models/Furniture.cs
sboigelot: I'm not a fan of Jack-Of-All-Trade event, but ok. The description should probably include more info.

TomMalbran: I agree with that.

We could also make more events that replace this one to be more specific, or just leave it like that.
Maybe as a separate issue/pr

Assets/Scripts/Models/Furniture.cs
bjubes: Renaming properties is a bit outside this PRs scope

Assets/Scripts/Models/Furniture.cs
TomMalbran: Maybe both. This property is really tied up to the pathfinder code. It could be used for something else, but calling it directly from the Furniture, doesn't make much sense.

Assets/Scripts/Models/Furniture.cs
sboigelot: I disagree, property names are the main documentation of your interface.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -206,8 +206,12 @@ public float PathfindingWeight
     /// These actions are called when Trigger is called. They get passed the furniture
     /// they belong to, plus a deltaTime (which defaults to 0).
     /// </summary>
+<<<<<<< HEAD
     /// <value>The event actions that is called on update.</value>
     public EventActions EventActions { get; private set; }
+=======
```
Grenkin1988: That's not good

Assets/Scripts/Models/Furniture.cs
longtomjr: I do not know enough about the path finding to change this comment. Feel free to change it, but atm I think it is as descriptive as I will be able to get it.

Assets/Scripts/Models/Furniture.cs
longtomjr: Will be fixed in next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: Done - in next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: Attempted fix in next commit, Please Check to see if it is any better

Assets/Scripts/Models/Furniture.cs
sboigelot: It could be:

>  Gets or sets the Furniture's pathfinding modifier which is added into the final PathfindingCost computation

Just remove the Tile reference.

Assets/Scripts/Models/Furniture.cs
longtomjr: Yes, property names is really important, I will add that this needs to be renamed in the documentation, but we have not agreed completely on how we will do that. I will add a separate PR after this one that will deal with the renaming of everything in Furniture .cs

Assets/Scripts/Models/Furniture.cs
longtomjr: Added to next commmit

Assets/Scripts/Models/Furniture.cs
longtomjr: Fixed - In next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: Tried to explain better, in next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: Added a more to clear it up a bit. In next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: Fixed in upcoming commit

Assets/Scripts/Models/Furniture.cs
longtomjr: Will be fixed in next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: updated docs

Assets/Scripts/Models/Furniture.cs
longtomjr: done in next commit
Along with the return tags

Assets/Scripts/Models/Furniture.cs
longtomjr: done next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: done next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: done next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: done next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: done next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: done next commit

Assets/Scripts/Models/Furniture.cs
longtomjr: Added a comment about that in the next commit.

Assets/Scripts/Models/Furniture.cs
longtomjr: Done, will be in the next commit. Ty

Assets/Scripts/Models/Furniture.cs
longtomjr: done in next commit

Assets/Scripts/Models/Furniture.cs
```diff
@@ -280,7 +283,8 @@ private set
     }
 
     /// <summary>
-    /// Gets a list of furniture this furniture can be replaced with.
+    /// Gets a list of furniture ObjectType this furniture can be replaced with.
```
TomMalbran: This is actually a list of Type Tags and it is fine as a string since those are defined in the xml which modders can modify.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -320,12 +325,13 @@ public List<string> ReplaceableFurniture
     public string LocalizationCode { get; private set; }
 
     /// <summary>
-    /// Gets the description of the furniture if there is no localization support.
+    /// Gets the description of the furniture. This is used by localization.
     /// </summary>
     public string UnlocalizedDescription { get; private set; }
 
     /// <summary>
     /// Gets a value indicating whether this furniture is next to any furniture of the same type.
```
TomMalbran: The comment is still wrong since it make me believe that LinksToNeighbour = true means that is next to a furniture which is not true. Instead it means that if is next to a furniture of the same then, the the sprite will change so that both furniture look connected.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -682,6 +691,7 @@ public void ReadXmlPrototype(XmlReader readerParent)
 
     /// <summary>
     /// Reads the specified XMLReader (pass it to <see cref="ReadXmlParams(XmlReader)"/>)
+    /// This is used to load furniture form a save file
```
TomMalbran: form -> from

Assets/Scripts/Models/Furniture.cs
```diff
@@ -882,16 +894,16 @@ public bool HasTypeTag(string typeTag)
...
     public string GetName()
     {
         return LocalizationCode; // this.Name;
     }
 
     /// <summary>
-    /// Returns the description of the furniture
+    /// Returns the UnlocalizedDescription of the furniture
     /// </summary>
     /// <returns>description of the furniture</returns>
```
TomMalbran: This should be "UnlocalizedDescription of the furniture"

Assets/Scripts/Models/Furniture.cs
```diff
@@ -244,28 +295,55 @@ public List<string> ReplaceableFurniture
...
     public bool LinksToNeighbour { get; private set; }
 
+    /// <summary>
+    /// Gets the type of dragging that is used to build multiples of this furniture. 
+    /// e.g walls
+    /// </summary>
     public string DragType { get; private set; }
 
+    /// <summary>
+    /// Gets. or sets the parameters that is tied to the furniture.
```
TomMalbran: "Gets or sets the parameters that are tied to this furniture."

mikejbrown:
Umm... did you mean to clobber all of your comments? I think you shouldn't have merged with "Resolved using Upstream." The upstream changes were pretty tiny so it's probably best just to roll back that merge and resolve manually.

longtomjr:
Yep, I am doing that
You should not try to git when you feel like i do atm. I need food, but all the mushy food is gone

Assets/Scripts/Models/Furniture.cs
```diff
@@ -573,6 +698,11 @@ public void ReadXmlPrototype(XmlReader readerParent)
         }
     }
 
+    /// <summary>
+    /// Reads the specified XMLReader (pass it to <see cref="ReadXmlParams(XmlReader)"/>)
+    /// This is used to load furniture form a save file
```
mikejbrown: from

Grenkin1988:
Merging it as is. If there is still error in documentation. Should be done on other PR

PR #1137, opened 2016-09-04
Autosaves

mikejbrown:
Implements an `AutosaveManager` class which manages autosaves using the `Scheduler` system. There is a new field `AutosaveInterval` in Settings.xml (type: int, default value: 2, values <= 0 disable autosave) which gives the interval in minutes. Saves are stored in successive files `Autosave1.sav`, `Autosave2.sav`, `Autosave3.sav` and so on in the usual save directory. Autosaves from previous game sessions are not overwritten or deleted automatically.

Note 1: I have not implemented the UI for the settings dialog to change the autosave interval. (Nor do I really plan to: Unity UI stuff is beyond me.) Hopefully someone can do that in another PR or a PR on this branch.

Note 2: Saving seems to work fine, but loading occasionally breaks references to WorldController, MouseController etc... I'm not sure why this occurs but I don't see how it could be specific to my changes.


Assets/Scripts/Controllers/AutosaveController.cs
```diff
@@ -0,0 +1,131 @@
...
+// and you are welcome to redistribute it under certain conditions; See 
+// file LICENSE, which is part of this source code package, for details.
+// ====================================================
+#endregion
+using System;
+using System.IO;
+using System.Xml.Serialization;
+using Scheduler;
+
+public class AutosaveController
```
TomMalbran: I think this should be an AutosaveManager.

mikejbrown:
@TomMalbran good point... name changed.

TomMalbran:
At some point we need to decide where to place this managers. There are managers in the Controllers folder that are not managers, but I am not sure if they should be moved to the Models folder, since they are not like the other Models.

In on of my projects I starting using Services as middleware between controllers and models. This would be like our managers. I might make an issue to discuss that.

svmnotn:
@mikejbrown I'm not sure what happened with this PR but [GetSettingAsInt](https://github.com/mikejbrown/ProjectPorcupine/blob/feature/autosave/Assets/Scripts/Utilities/Settings.cs#L98) was replaced with [GetSetting](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/Assets/Scripts/Utilities/Settings.cs#L98)

Not quite sure how it hasn't been marked as a merge conflict... Please update and fix

mikejbrown:
@svmnotn Hmm.. #1092 got rid of that function and I guess the CI wasn't redone for this PR... that doesn't seem like the most robust CI system.

Anyway I'm fixing now.

svmnotn:
now it has conflicts.... What?.... ???

mikejbrown:
@svmnotn should be good now...

PR #1139, opened 2016-09-04
New FunctionsManager

### LINKED ISSUE #1130: Move Lua Global Registers to a more suitable place.
As the TODO says these lines aren't furniture related and so shouldn't be in the furniture class

```
public FurnitureActions()
    {
        // TODO: This should be moved to a more logical location
        LuaUtilities.RegisterGlobal(typeof(Inventory));
        LuaUtilities.RegisterGlobal(typeof(Job));
        LuaUtilities.RegisterGlobal(typeof(ModUtils));
        LuaUtilities.RegisterGlobal(typeof(World));
        LuaUtilities.RegisterGlobal(typeof(WorldController));
        LuaUtilities.RegisterGlobal(typeof(Power.Connection));

        LoadScripts();
    }
```

TomMalbran:
This replaces the static LuaUtilities with a generic Actions class. I then created an ActionsManager which creates and instance of Actions for each Lua file, so that they are all on its own domain. Now all the lua functions are in a single location which is a good step into a better modding.

This should fix #1130.


koosemose:
You may want to change the trigger method (line 126 of EventAction) to a generic, as is it can only accept Furnitures.

koosemose:
Also, ran this for a while with no crashes on save or load, building several furnitures and letting them run. Only errors are preexisting ones.

TomMalbran:
@koosemose Changing how EventAction works is not part of the scope of this PR. That is a different issue and should be fixed in its own PR.

koosemose:
I interpreted the scope as making the actions system more generalized, which is why I suggested it. But no problem doing it elsewhere.

TomMalbran:
Ok. This is about making the Lua Actions generalized. The EventActions is something totally different.

Grenkin1988:
Why class is names `Actions` ?
As I see from code what it does is:
1. Load and initialize lua script.
2. Call **functions** from lua script. `public DynValue Call(string functionName, params object[] args)`

Did I miss something related to actions?

TomMalbran:
Mainly because it replaced the FurnitureActions, NeedActions, GameEventActions and QuestActions classes. It all started with Quill using the Actions suffix in FurnitureActions and later people copying that file for other Actions. I started with a BaseActions class and having the other extend it, but at the end I didn't needed to, so I renamed BaseActions to just Actions.

But yes, is just a collection of lua functions, which could come from 1 or multiple files/scripts. I don't mind changing the name, but what would you call them?

Grenkin1988:
@TomMalbran I do not mind to name it as `Actions`, but we need consistency. If you name it `Actions` than change name of parameter of `public DynValue Call(string functionName, params object[] args)` to `public DynValue Call(string actionName, params object[] args)`

TomMalbran:
Renaming functionName with actionName would make it unclear, since you are calling a function. Maybe the class should be Functions? Although that seems too generic. LuaFunctions might work too, and have a FunctionsManager.

sboigelot:
I think we should use Function over Action as we return a value from the lua functions

TomMalbran:
I renamed the classes to FunctionsManager and LuaFunctions, since that is what they are.

Also fixed the issues and moved the lua utils tests to be the LuaFunctions tests.

TomMalbran:
I can see the failing test. Trying to figure how to fix it, since it is now testing a different function.

koosemose:
If LoadScriptFromText should fail with a System.NullReferenceException when passed null, then it's working correctly, you just need to change what exception to expect.

TomMalbran:
Before the change the function was checking for LoadScriptsFromFile which where throwing an exception because we didn't catch some of the errors. Now with LoadScriptsFromText, the exception is catch, so in the tests there is no exception to catch. I could throw back the exception (not sure if we want those), or add a boolean return value to the function.

koosemose:
An exception isn't caught somewhere. It's throwing a System.NullReferenceException when you pass it null. If that's what it's supposed to do, then you need to change:
```
[ExpectedException(typeof(System.ArgumentNullException))]
      public void Test_LoadScript_Null()
```
to:
```
[ExpectedException(typeof(System.NullReferenceException))]
      public void Test_LoadScript_Null()
```

if it's not supposed to throw an error, you need to stop it from throwing that error, and tell the test what to expect instead.

mikejbrown:
```
    [ExpectedException(typeof(System.ArgumentNullException))]
    public void Test_LoadScript_Null()
    {
        // Try loading a file from the path null
        functions.LoadScriptFromText(null, "");
    }
```
I think, contextually, `ArgumentNullException` is better. You should check for null in `LoadScriptFromText` and raise that instead of NRE. Also the comment is misleading... it's not loading a file.

Regarding the `SchedulerEditorTest`s, any unit test touching lua functions should be rewritten to use `LoadScriptFromText`. Tests should be self-contained not touch the filesystem.

koosemose:
Then you need to make it throw an argument null exception, so it does what the test expects.

But I will also note that it might be good to still test directly for if the file load functionality works, in case someone breaks something in it in the future. It shouldn't pass tests if it can't load a file, but it should clean up the file afterwards.

TomMalbran:
I fixed the tests. I added a return on LoadScript to know if it loaded correctly or not.

I didn't wanted to change the load functions since my ideas was to move them. So I just moved them. The LuaFunctions class shouldn't know about the path and files. The ModManager will deal with all that stuff.

I forgot that I wanted to fix the Scheduler test. I remember the request with the lua change, and was thinking that this PR would make that even better.

mikejbrown:
This has conflicts now.

TomMalbran:
Conflicts resolved

koosemose:
You have some Stylecop violations, Get those fixed and this should be ready for merging.

mikejbrown:

TomMalbran:
Conflicts and StyleCop fixed

koosemose:
Looks good, no unexpected bugs, no Stylecop violations. Merged.

Grenkin1988:
@koosemose Is there expected bugs? :-)

koosemose:
Yes, lots... though I'm mostly referring to bugs that are known to exist prior to a given PR, and therefore weren't caused by them.

PR #1146, opened 2016-09-05
Prototype Manager Update

TomMalbran:
I moved the load functions outside of the constructor. All the prototypes are loaded by the ModsManager, since that is how it will work in the near future. This makes it possible to test with the PrototypeManager, since you can create an instance and then load the prototypes you want manually.

I also changed some of the function names to make them shorter.


TomMalbran:
The 3 unit test failing are from the Scheduler, because the Prototypes are now empty, but I believe that those should be fixed in #1132. But at the same time this change should make fixing them in #1132 a lot easier.

mikejbrown:
Gatekeepers: If you prefer to go this route you could close #1132 (as this makes that one largely irrelevant) and we'll make the needed fixes here so we have one working PR instead of two broken ones.

TomMalbran:
@mikejbrown I can also make a pr to your code changing only the loading and keep this one with just the functions names. And you could then fix the tests. Or you can just copy the loading fix and I can remove it from this pr.

mikejbrown:
I think we should close #1132 and get the tests working on this.

mikejbrown:
https://github.com/TomMalbran/ProjectPorcupine/pull/1 addresses the unit tests

kd7uiy:
Okay, so this failed the CI tests. I'm not understanding why it failed, perhaps someone could illuminate me on that?

mikejbrown:
@kd7uiy This breaks the `SchedulerEditorTest`s because they use the `PrototypeManager`. I have a PR on this https://github.com/TomMalbran/ProjectPorcupine/pull/1 that should fix the issue. @TomMalbran just needs to merge it in.

bjubes:
@kd7uiy  when my pr #1151 is merged it will say at the bottom of the Travis log exactly which unit test failed. for now you can expand the XML file travis logs and find it manually.

mikejbrown:
This has merge conflicts now.

TomMalbran:
Fixing them right now.

@mikejbrown The test fix is way better now, the code looks much cleaner. Soon we will be able to do the same with the Lua functions.

TomMalbran:
Conflicts resolved. Waiting on Travis now.

mikejbrown:
Works for me. Happy to merge as soon as Travis passes.

TomMalbran:
All Travis checks passed.

mikejbrown:
Merged!

PR #1151, opened 2016-09-05
Log failed unit tests to Travis CI

bjubes:
when a Travis build fails due to a unit test failing, Travis will now log the exact failure at the end of the build log.

here is a sample where two unit test fail:
https://travis-ci.org/bjubes/ProjectPorcupine/builds/157662489

### DIFF
--- a/Scripts/test.sh
+++ b/Scripts/test.sh
@@ -40,7 +40,7 @@ if [ "$errorCount" != "0" ]; then
     echo $errorCount ' unit tests failed!'
      
      #show the exact unit test failure
-    echo 'The following unit tests failed:'
+    echo '\nThe following unit tests failed:'
     echo | grep 'success="False"' EditorTestResults.xml | grep 'test-case'
    
     exit 1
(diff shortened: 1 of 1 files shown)


bjubes:
for comparison purposes, this is a build that failed two unit tests on the current master 
https://travis-ci.org/TeamPorcupine/ProjectPorcupine/builds/157551821

Grenkin1988:
Could we somehow highlight ` success="False" ` ?

svmnotn:
@Grenkin1988 why? from the link it shows the name of what test failed, so there is no need to specifically highlight that they failed (if they are shown, they failed) if anything I think maybe only showing the name property instead of the whole xml line might look better.

Grenkin1988:
I am blind perhaps. Where?
image

svmnotn:
wrong link @Grenkin1988 it is this one: https://travis-ci.org/bjubes/ProjectPorcupine/builds/157662489

Grenkin1988:
O.K. I am blind :-(

PR #1158, opened 2016-09-05
Furniture Animation defined in xml

Raketfart:
Furniture animations are now easy to add the data/furniture.xml, and switching animation states in lua.

I redid the cloning pod, to test that it worked on an existing animation.
I also stole (with permission) @Tranberry 's power generator, and added a light, that will shine when the power generator is working:
furniture animation


sboigelot:
As I and many other, I think you are branching out of your local/master and not out of upstream/master, this explain why you have all the previous merge in your PR. It's not a big deal for this PR by think about it for the next one ;)

Raketfart:
@sboigelot Yes, you are probably right. I don't remember exactly how I created this branch, but I will try to do what you suggested next time.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -485,29 +485,17 @@ function CloningPod_UpdateAction(furniture, deltaTime)
         false
     )
 
+    if (furniture.JobCount() < 1) then
```
TomMalbran: At this point ```furniture.JobCount() == 0``` and we are about to create a new job, so this isn't required.
You could move the SetAnimation after the job is added.

TomMalbran:
This looks like a great first step. We can expand this to handle other animations like the door or the oxygen compressor and the accumulator.

For the oxygen compressor and the accumulator it will be easy to have a 1 fps animation for each stage, and set those on the update. For the door we could do the same or try to have an animation that does the opening which is synched with the openness. Or do you think that it might benefit of a different type frame by frame animation where each frame is set manually or is based on a param?

Eventually we could remove GetSpriteName. It was a partial solution until we got proper animations.

Tranberry:
everything can't be animated it will take focus from more important things for the player.. ;)

Raketfart:
Yeah, I was planning to go through the other existing animations, and find solutions for those.
Doors and oxygen compressors are great examples.
It's definetly possible to just have 1 frame "states" and switch them in lua, but I would like to make something a bit better, because lots of furniture will have animations depending on some value. 

Maybe normalizing the value in lua (zero to one), and the the animation system could figure out what frame to go to, depending on total frames.

Tranberry:
what about a FPS value for the XML/SKON/JSON ?

mikejbrown:
Haven't looked at the code or tested yet, but I love the look.

Raketfart:
@TomMalbran Thanks for correction. It's changed now.
There seems to be a small bug, where the job time is reset, if power goes out. Then the cloning pod gets a new 10 second job, but that's unrelated to this.

Raketfart:
@Tranberry Not sure what you mean. The FPS is controlled here:
<Animation state="running" fps="6" looping="true">

Tranberry:
please ignore, I'm stupid :cactus:

TomMalbran:
Right, that is because the jobs are canceled and recreated. Didn't thought that it wouldn't make sense for things like the cloning pad. We should add an issue for that, the job should be paused not canceled on that case.

For the animations what if in the animation you have an attribute for a param name, a min and a max, and have as many frames as you want, each with a value: 0 to 100 (0% to 100%, or 0 to 1). You could then map the attribute value to a frame. The frame could also be set in lua, by passing the param name, or a percentage.

This can be added here on in a different pr. We don't really need to wait for that implementation to merge this PR.

Raketfart:
@TomMalbran Yeah, I'm trying to keep my PR's relatively short :) 

There are many ways to do the value based animation. I will try to go with the simplest one. I think it's more important to easily understand the xml and lua, rather than having full control over small percentages. So 5 animation frames will probably just be 20% each of the value they are monitoring. I think that will work for most of our animation needs.

mikejbrown:
What about using the Scheduler for the timing and updating?

Raketfart:
I guess the Scheduler could easily be used instead of update for timed animations like these.
I'm not sure how well that would work with value-based animation like doors.

koosemose:
@Raketfart Just to be clear, are you intending on investigating the use of scheduler, or is this good to be looked at for merge now?

Raketfart:
If i am correct, that the scheduler would only work for one case, I would rather just use the current version and have both cases work similarly.

koosemose:
Alright, however you now have conflicts, adding [Conflicts], as usual remove when fixed

Raketfart:
Fixed now

Tranberry:
_powerplantlight

my current proposal for a new power generator

TomMalbran:
The new sprite looks really good. The on running effect using only the light seems a bit sutil (I didn't notices the differences in the sprite at first), but I would need to try it in the game to see if we need something more visual.

Tranberry:
_powerplantlight

@TomMalbran 
Do you want more animation?

TomMalbran:
It looks good. Will see how it looks once is on the game. We can change it later if we need too.

sboigelot:
Works great and the new power generator sprite is awesome :)

PR #1166, opened 2016-09-06
Removed redundant HasPower() call in Lua

Tritaris:
`Furniture.cs` already has this check in `Update()`it before it calls the Lua `OnUpdate`


TomMalbran:
@koosemose For furniture that wants to handle things different when it has no power, we can add an attribute in the xml, and when is set, it will not do the power check. Since most of the furniture works only when it has power, it makes sense to have it in Furniture for all, and make the furniture that wants to handle it, the special case.

koosemose:
That sounds reasonable.

koosemose:
Looks good, no unexpected bugs, only touched Lua so no stylecop. Merged.

PR #1167, opened 2016-09-06
Fixes #1161 - Jobtime reset after powerloss

### LINKED ISSUE #1161: [Bug] Furniture jobtime reset after loss of power.
There seems to be a bug, where the job time is reset, if power goes out. This applies to at least the Cloning Pod in Lua. I'm not sure if this is a general problem with all furniture, or just some of them.

Tritaris:
Fixes #1161

When power is lost the jobsTimes would reset because the jobs were canceled. This new method takes the current jobs running and stores them in a new List until the power is restored then the jobs are added back to the normal job queue with any current work done to them saved.

### DIFF
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -58,6 +58,8 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider
 
     private List<Job> jobs;
 
+    private List<Job> jobsQueued;
+
     // This is the generic type of object this is, allowing things to interact with it based on it's generic type
     private HashSet<string> typeTags;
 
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Models/Furniture.cs
```diff
@@ -58,6 +58,8 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider
 
     private List<Job> jobs;
 
+    private List<Job> jobsQueued;
```
TomMalbran: We should call these pausedJobs, or something different. Since now it seems the same as the jobs list.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -613,6 +621,26 @@ public void CancelJobs()
         }
     }
 
+    public void EnqueueJobs()
```
TomMalbran: I would call these functions PauseJobs and ResumeJobs, since that is essentially what they are for.

Raketfart:
:+1:

koosemose:
Looks good, no unexpected bugs, clean stylecop on all changed files. Merged.

PR #1169, opened 2016-09-06
Build Menu - Non-buildable furniture

TheRiven:
This adds the ability to tag furniture with a `TypeTag` that stops it from showing up in the build menu, which checks before generating a button and moves on to the next furniture if it has the tag.

Currently this is only implemented on the mining points so that the player cannot build them, but will likely be used for most furniture objects created during world gen such as asteroid walls.

For the moment the tag is `Non-buildable` but I am not sure this is ideal and I am open to better suggestions.

I have also redone a bunch of names in the `FurnitureBuildMenu` to hopefully meet standards and to pre-empt any suggestions that they be changed to something else. Please point out if its to much or not correct so I know for the future. :+1:


bjubes:
can you make it so that if dev mode is enabled then those items reappear in the build menu?

PR #1163, opened 2016-09-05
Revert "UberLogger can detect that he called from Unit Tests"

mikejbrown:
Reverts TeamPorcupine/ProjectPorcupine#1129

PR #1181, opened 2016-09-06
Added Debug.IsLogEnabled to be able to disable log.

Grenkin1988:
One more try after #1129 #1163 failed.

Now you just need to disable Log from tests.


mikejbrown:
_Looks_ clean this time. Even tests fine... **merging** _What could possibly go wrong?_ :worried:

PR #1183, opened 2016-09-06
Rework of FurnitureBuildMenu

TheRiven:
Reworked `FurnitureBuildMenu` to allow the furniture list to be rebuilt on the fly, this also allows for Dev mode to make all furniture visible/buildable regardless of Non-buildable flag.

You can call the `RebuildMenuButtons` function to rebuild the list and either tell it to display everything or just rebuild the list.

It should be fairly easy to expand on the checks to make certain buildings become visible when required or vice versa.

Any suggestions for changes welcome! :+1:


bjubes:
Awesome change! this is great now for switching in and out of dev mode and will be great when items get unlocked in the future.

PR #1184, opened 2016-09-06
[small] Fix Stylecop warning in furnitures.cs documentation

Ohda:
Fix Stylecop warning in furnitures.cs documentation

### DIFF
--- a/Assets/Scripts/Models/Furniture.cs
+++ b/Assets/Scripts/Models/Furniture.cs
@@ -251,7 +251,7 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider
 
     /// <summary>
     /// Gets the BASE tile of the furniture. (Large objects can span over multiple tiles).
-    /// This should be RENAMED (possibly to BaseTile)
+    /// This should be RENAMED (possibly to BaseTile).
     /// </summary>
     /// <value>The BASE tile of the furniture.</value>
     public Tile Tile { get; private set; }
@@ -340,7 +340,7 @@ public class Furniture : IXmlSerializable, ISelectable, IContextActionProvider
 
     /// <summary>
     /// Gets the type of dragging that is used to build multiples of this furniture. 
-    /// e.g walls
+    /// e.g walls.
     /// </summary>
     public string DragType { get; private set; }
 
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Models/WorldGenerator.cs
```diff
@@ -80,12 +80,12 @@ public static void Generate(World world, int seed)
 
         for (int z = 0; z < depth; z++)
         {
-            float zScale = Mathf.Lerp(1f, .5f, Mathf.Abs((depth / 2f) - z) / depth);
```
mikejbrown: Why are you changing this?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -700,9 +700,9 @@ public void ReadXmlPrototype(XmlReader readerParent)
 
     /// <summary>
     /// Reads the specified XMLReader (pass it to <see cref="ReadXmlParams(XmlReader)"/>)
-    /// This is used to load furniture form a save file
+    /// This is used to load furniture form a save file.
```
mikejbrown: from

Assets/Scripts/Models/WorldGenerator.cs
kd7uiy: There is a StyleCop error that shows up, because it things that z is a prefix. The more correct solution would be to add z to the list of allowed prefixes, or just disable that check entirely.

Ohda: I've got this (something bad on me setup ?)
SA1305 : CSharp.Naming : The variable name 'zScale' begins with a prefix that looks like Hungarian notation. Remove the prefix or add it to the list of allowed prefixes.

mikejbrown: Well scaleZ isn't that bad and it's only a local variable anyway.

kd7uiy:
A few things with the title. First of all, it's StyleCop, not StyleCoop. Secondly, you should add that most of the changes were simply adding a period to existing documentation, or something about that. It only matters because the title of the PR becomes the title of the commit, which can help to figure out what happened. I know, minor detail, but... Thanks!

Issue #630, opened 2016-08-22
Inappropriate code in MouseController


abackwood:
From MouseController.UpdateSelection():

`if (Input.GetKeyUp(KeyCode.Escape))
        {
            mySelection = null;
        }`

This should not technically be in here. This controller should only control mouse inputs, not keyboard inputs

FilippoLeon:
can add to keyboardController

Mizipzor:
Do it! :)

TomMalbran:
I think that the MouseController does too much stuff and it makes so that if we need something that depends on the mouse we need to go to that file.

I think that it should be more event driven and don't depend on any other module, but have other modules depend on it. Something like if you need an on click event, you can register an onClick callback on the MouseController that would give you the tile under the mouse, or just the x and y position. You could also register events like startDrag, onDrag, endDrag, onMouseMove, and other functions that would allow to show something when the mouse moves. Essentially the MouseController would be in charge of only keeping track of the mouse position, the drag rectangle, and it can show the cursor. A different module could keep track of the MouseMode which would be renamed, and we can have a third module to deal with the Selection MouseMode.

I am still not sure on how (or if) it will all work together, but it is a better approach than a centralised code that would keep growing.


A similar thing can be done with the KeyboardController. The different modules would register modes and then bind keys to those mode and a callback. This is so that the same key can be bind to multiple callbacks. For example, I could create a KeyboardMode for the settings dialog, and then bind "Esc" to the closing of the settings dialog. When the dialog opens, the mode is set and when it closes it is unset. Then I can do the same thing for the options menu, so that "Esc" works for both. But doing it in this way, the KeyboardController knows which function to call when a key is pressed.

Grenkin1988:
@abackwood  This code is still there. What was the decision?

kd7uiy:
Looked at this a bit, it doesn't belong in KeyboardManager. I think I agree that the MouseController is too large, and some of the logic should be separated out. If no one gets to this, I'll take a look at it at some point.

TomMalbran:
@kd7uiy If you want to work on it, it would be cool if the mouse controller was more like the KeyboardManager, where it had events or register functions and callback and no logic from any other class. Will be harder to do, but it won't be a monolithic class. (Like I said before...)

kd7uiy:
[...]

PR #1185, opened 2016-09-06
Drops the current build item when switching construction menus

### LINKED ISSUE #508: [Bug, Easy] Changing menus should drop the active build.
If I'm building a furniture, and click on the floor, we should get rid of the furniture building menu.

kd7uiy:
This will drop the item being build when switching menus. As a result of this change, it allows for ease of access to this clearing function, which might make the change in #630 easier.

Fixes #508


kd7uiy:
gif

kd7uiy:
gif

bjubes:
Merged!

Issue #1126, opened 2016-09-04
[Discussion]Require in code documentation with new Pull Requests


longtomjr:
Pretty much what the title say.  
At the moment there is way to much systems, code etc. that have no documentation about them. I think to start solving this we should enforce a rule for documenting everything you do.

**How to document:**  
I recommend the StyleCop accepted way. It is standard and after a lot of work we might be able to add the doc rules back into our rule set.

### Current proposal: 
**This is what is required to have a PR be merged:**  
* Summary on top of all public properties and methods.
* Description of Parameters of Methods.
* Comments inside of methods is discouraged - Code should be readable and speak for itself!
* Summaries for private methods and fields is **not required** but may be added.
*  Summaries for constructors is *not* needed but it is a good idea to add for complex constructors.
**_Remember:_** Comments is not a tutorial. It should describe its function and not how it works!

vogonistic:
Having everything documented would be fantastic. Summary should be required and params optional unless it's complicated. 

We should start by just reviewing for it until we get close to being complete.

Tranberry:
Sounds very good for all aspiring developers. :+1:

mattpott:
This is a great idea but it is a fine line between good comments and over commenting ie
`int world = new world(); // This creates a new world` 
To me that would be redundant, thoughts?

bjubes:
I think that documenting on a separate file, like the wiki or some other system, may be more helpful in the long run. A separate place for documentation allows the documentation to be longer and higher level, so instead of just explaining a function, you can explain how a function is used in the grand flow of the program.

bjubes:
also long summaries clutter up code. And it also can get in the way of devs who know what their doing. for example, everyone who uses unity sees this file:
```
public class test : MonoBehaviour {

	// Use this for initialization
	void Start () {
	
	}
	
	// Update is called once per frame
	void Update () {
	
	}
}
```
The first thing I do is delete the comments, as they are redundant/unneeded for anyone who knows what they are doing. Imagine working on a file in this project and having to scroll through long summaries to find the actual code. Much better to offload the explanation to a seperate place where beginners can find it IMO.

sboigelot:
@mattpott & @bjubes concerns are understandable. Over documentation can make code unreadable. It's always better - for me - to have a code that is easy to understand without documentation then a complex code with documentation.

We should trive to have self-explanatory class and method names

[...]

PR #1187, opened 2016-09-06
Allow disabling saves of ScheduledEvent. Fixes #1180.

### LINKED ISSUE #1180: Scheduler- Add an option to not save event on load
Some events, such as the HeadlineGenerator, really shouldn't be saved, as they interact with UI elements. There should be an option to not save those types of events. Related to #1175 .

mikejbrown:
Fixes #1180 
Adds a `bool ScheduledEvent.IsSaveable` flag which is checked by `Scheduler` before serializing the event.

Also adds documentation comments per #1126. :grinning:

### DIFF
--- a/Assets/Editor/SchedulerEditorTest.cs
+++ b/Assets/Editor/SchedulerEditorTest.cs
@@ -263,6 +263,27 @@ public class SchedulerEditorTest
     }
 
     [Test]
+    public void SchedulerWriteXmlDoesNotWriteNoSaveFlaggedEventsTest()
+    {
+        ScheduledEvent evt = new ScheduledEvent(
+            "test",
+            callback,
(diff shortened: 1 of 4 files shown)


bjubes:
can you also make the Headline Generator use the IsSaveable flag. This will close the issue as opposed to just making a system for someone else to close it (ie. will never get fixed)

mikejbrown:
@bjubes I prefer to make PRs touch as few systems as possible, but it's done.

bjubes:
I understand wanting to make a PR small, but when you make a fix, you should implement it too. otherwise we end up with tons of systems that no ones knows exist because their not in use anywhere in the project

bjubes:
Thanks for the fix!

PR #1194, opened 2016-09-06
Fixes #844

### LINKED ISSUE #844: [Bug] Temperature system doesn't update correctly
The temperature system as it is since #835 updates every 5 seconds. However, the deltaTime passed to UpdateTemperature() is only the current frame's deltaTime. That should be the total elapsed time since last update, or the effect will be much too small.

sboigelot:
Fixes #844 

Use the actual time passed between update for temperature.cs instead of the Time.DeltaTime

### DIFF
--- a/Assets/Scripts/Models/Temperature.cs
+++ b/Assets/Scripts/Models/Temperature.cs
@@ -96,8 +96,8 @@ public class Temperature
 
         if (elapsed >= updateInterval)
         {
-            ProgressTemperature(Time.deltaTime);
-            elapsed = 0;
+            ProgressTemperature(updateInterval);
+            elapsed = elapsed - updateInterval;
         }
     }
 


mikejbrown:
Why not using the scheduler?

sboigelot:
We shouldn't confuse the scheduler with everything.

For me the scheduler is not for defined task that are core to the system. It's more for events. You wouldn't use the scheduler for the animation of the furniture, or the movement of character.

sboigelot:
What I mean is: scheduler is for events, not for tick update, the delay on the tick update here is only to reduce the load on the computing

mikejbrown:
I would think that's a perfect use case for the scheduler. You have an event ("update the temperature") which is supposed to happen at regular intervals based on the game clock. It's already being used for autosaves, for instance. It's also open to lua so modders can eventually go wild with it. So far no one is relying on the scheduler queue containing only a limited kind of gameplay related events, nor should they, I think.

sboigelot:
@mikejbrown I think we need to decide this as a group, I see your point but then everything should be moved to the scheduler.

I feel it is out of the scope of this bugfix PR. Let's fix the bug, have the feature work again (this is just 2 line of code). Then we can discuss about changing the whole update tick process.

mikejbrown:
@sboigelot Yeah, fair enough. Let's open another issue for that discussion.

sboigelot:
I'm doing it ;)

bjubes:
so the whole scheduler debate may take a while, can we merge this as is to fix the bug and then add scheduler support when there is consensus?

Grenkin1988:
I think we should not use scheduler for game mechanics such as power, temperature, job...

koosemose:
We shouldn't let even small bugs stand while waiting on a decision to be made if we already have a fix, as the decision may take a long time, and it can be changed in a future PR if it's needed. Merged.

PR #1199, opened 2016-09-07
Fix skybox loading in build - Fixes #1044

### LINKED ISSUE #1044: [Bug] SkyBox not working after build
I was wondering why the skybox is not rendered when the game is builded.

koosemose:
Fixes #1044
LoadSkybox was previously searching for available skyboxes as files, which don't exist in that form on a build. Changed to make use of LoadAll


Tritaris:
I can confirm the background is now working again in the build version.

sboigelot:
Works great for me too.

PR #1201, opened 2016-09-07
Moved the jobs list and functions from Furniture to a new class

TomMalbran:
In an attempt to simplify a bit the Furniture code I moved the stuff related to jobs in Furniture to a new class (component). Furniture is still huge, but is now a tiny bit smaller. I also did some minor refactoring on the code in Furniture and in Furniture.lua.


sboigelot:
@TomMalbran why did you add a space after each if in lua, is it a new standard I don't know about?

Assets/Scripts/Models/FurnitureJobs.cs
```diff
@@ -0,0 +1,214 @@
...
+        SpawnSpotOffset = new Vector2(
+            int.Parse(reader.GetAttribute("X")),
+            int.Parse(reader.GetAttribute("Y")));
+    }
+
+    /// <summary>
+    /// How many active jobs are linked to this furniture.
+    /// </summary>
+    /// <returns>The number of jobs linked to this furniture.</returns>
+    public int Count()
```
mikejbrown: Maybe `ActiveJobsCount` or the like for a more descriptive name?

Assets/Scripts/Models/FurnitureJobs.cs
```diff
@@ -0,0 +1,214 @@
...
+    public Vector2 WorkSpotOffset { get; private set; }
+
+    /// <summary>
+    /// Gets the spot offset where inventory is spawn when a Job is done with this machine.
+    /// </summary>
+    /// <value>The spawn spot offset.</value>
+    public Vector2 SpawnSpotOffset { get; private set; }
+
+    /// <summary>
+    /// Gets the <see cref="Job"/> with the specified index.
```
mikejbrown: active Jobs only

Assets/Scripts/Models/FurnitureJobs.cs
```diff
@@ -0,0 +1,214 @@
...
+        job.OnJobStopped += OnJobStopped;
+        World.Current.jobQueue.Enqueue(job);
+    }
+
+    /// <summary>
+    /// Cancel all the jobs linked to the current furniture.
+    /// </summary>
+    public void CancelAll()
+    {
+        Job[] jobsArray = activeJobs.ToArray();
```
mikejbrown: Should also cancel pausedJobs?

TomMalbran: All the paused jobs are already canceled. For now pause cancels a job to remove them from the queue and the character, and stores them in a list, to save the progress. Later pause adds all the jobs to the jobs queue, like if it was a new job, but it has the same progress as before.

Assets/Scripts/Models/FurnitureJobs.cs
TomMalbran: I was kind of going with the idea that you didn't need to know about the implementation of how active and paused jobs worked, and was going for similar names as before. Since all paused jobs are canceled jobs, and we are storing them in a way similar to holding a prototype, Count returns the amount of jobs linked to this furniture. In a way you don't Add a paused or active job, you add a job and we deal with it. I can change it for ActiveCount (since you always call it as Jobs.ActiveCount).

I just change the lists names internally to make it clear.

Assets/Scripts/Models/FurnitureJobs.cs
TomMalbran: Good point

TomMalbran:
@sboigelot The Lua file is very inconsistent in that aspect. Many ifs had a space after (mostly the new ones), many didn't (many added by Quill). I went with space for all because:
1. It makes it look similar to C# where we have a space after.
2. Lua doesn't really require parenthesis in the if statement if you have a space after if and before then.
3. If there is always a space before then, i think there should always be one after if for symmetry.

We should really have a standard for Lua. Right now there are for loops without parenthesis, which make the code look inconsistent. We should either remove all the outer parenthesis or add them to all.

sboigelot:
While testing I'm getting an error while a character is using the metal smelter:

```
chunk_8:(393,4-59): attempt to index a nil value
UnityEngine.Debug:LogError(Object)
UberLogger.Logger:ForwardToUnity(Object, LogSeverity, Object, Object[]) (at Assets/UberLogger/UberLogger.cs:506)
Debug:ULogErrorChannel(String, String, Object[]) (at Assets/UberLogger/Debug.cs:322)
LuaUtilities:CallFunction(String, Object[]) (at Assets/Scripts/Utilities/LuaUtilities.cs:44)
Job:DoWork(Single) (at Assets/Scripts/Models/Job.cs:230)
Character:CheckForJobMaterials() (at Assets/Scripts/Models/Character.cs:926)
Character:Update_DoJob(Single) (at Assets/Scripts/Models/Character.cs:732)
Character:Update(Single) (at Assets/Scripts/Models/Character.cs:345)
World:TickEveryFrame(Single) (at Assets/Scripts/Models/World.cs:169)
WorldController:Update() (at Assets/Scripts/Controllers/WorldController.cs:162)

```

I'm not getting this error in master, loading the same file.

I attached the file for testing purpose
[chunk error.zip](https://github.com/TeamPorcupine/ProjectPorcupine/files/460321/chunk.error.zip)
(unzip it in C:\Users\_**{user}**_\AppData\LocalLow\DefaultCompany\ProjectPorcupine\Saves to make it available in your game

You can also reproduce it by building a metal smelter after having a mining drone and power

TomMalbran:
It was probably not the best idea to both some functions called a lot in lua and the lua styling in the same PR. I see that file and I want to fix it, is way too messy.

I replaces funiture.lua with the original file and only changed the functions that are now in FurnitureJobs.

There are no errors now. I both tested the same and test case.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -216,48 +226,48 @@ function Stockpile_UpdateAction(furniture, deltaTime)
...
-        nil,
-        nil,
-        0,
-        itemsDesired,
-        Job.JobPriority.Low,
-        false
-    )
-    job.JobDescription = "job_stockpile_moving_desc"
-    job.acceptsAny = true
+	local j = Job.__new(
```
Grenkin1988: NO!  `j` is no!

TomMalbran: That is an older version, where I tried to fix all those things and made this PR a lot harder to review and work with. I am not going to refactor Lua in this PR, changing Lua is harder since you don't get errors, so it requires more testing.

Make an issue to settle on a style for Lua, since Lua is a mess. And then we can fix all the issues in 1 or multiple PR. I will be happy to fix them later.

TomMalbran: Notice that I didn't change job to j. j was the original name, changed all of them to job, and lots of other things. Added an error somewhere in the file. And since we don't get Lua errors as we do with C#, it makes it harder to find and test. So I backed up the changes. I already changed several calls in Lua.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -577,7 +577,7 @@ public void ReadXmlPrototype(XmlReader readerParent)
                     DragType = reader.ReadContentAsString();
                     break;
                 case "BuildingJob":
-                    ReadXMLBuildingJob(reader);
+                    ReadXmlBuildingJob(reader);
```
Grenkin1988: Why? XML is appropriate name. This is common abbreviation.

TomMalbran: It is. But for some reason it is Xml everywhere else, even the library uses Xml. So for consistency I changed it.

Assets/Scripts/Models/FurnitureJobs.cs
```diff
@@ -0,0 +1,214 @@
...
+    /// </summary>
+    /// <param name="i">The index.</param>
+    public Job this[int i]
+    {
+        get
+        {
+            return activeJobs[i];
+        }
+    }
+
```
dusho: is there a reason why only indexer for Jobs? exposing jobs collection would allow various linq on jobs.. otherwise we all need to put new methods with various queries on jobs...

sboigelot:

Assets/Scripts/Models/FurnitureJobs.cs
TomMalbran: Yes we don't want to expose the collection since we need for people to use the Add and Remove methods. Adding or removing jobs directly from the collection will cause issues.

TomMalbran:
Conflicts fixed and some comments.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -413,12 +422,18 @@ function MetalSmelter_JobWorked(j)
 end
 
 function PowerCellPress_UpdateAction(furniture, deltaTime)
-    local spawnSpot = furniture.GetSpawnSpotTile()
+    local spawnSpot = furniture.Jobs.GetSpawnSpotTile()
 
     if(spawnSpot.Inventory == nil) then
+<<<<<<< HEAD
```
Grenkin1988: Not supposed to be here

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -319,12 +323,17 @@ function MiningDroneStation_UpdateAction( furniture, deltaTime )
 
 	j.RegisterJobCompletedCallback("MiningDroneStation_JobComplete")
 	j.JobDescription = "job_mining_drone_station_mining_desc"
-	furniture.AddJob( j )
+	furniture.Jobs.Add( j )
 end
 
 function MiningDroneStation_JobComplete(j)
+<<<<<<< HEAD
```
Grenkin1988: Not supposed to be here

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -301,12 +301,16 @@ function MiningDroneStation_UpdateAction( furniture, deltaTime )
 		return
 	end
 
+<<<<<<< HEAD
```
Grenkin1988: Not supposed to be here

TomMalbran:
Opps. That wasn't a good job in the merge fix. Fixed it now.

mikejbrown:

TomMalbran:
Conflicts fixed.

mikejbrown:

TomMalbran:
Fixed the merge issues.

PR #1206, opened 2016-09-07
Fix loading character inventory from save

koosemose:


Assets/Scripts/Models/World.cs
```diff
@@ -834,31 +834,53 @@ private void ReadXml_Characters(XmlReader reader)
...
+                            }
+                            break;
                     }
                 }
             }
             while (reader.ReadToNextSibling("Character"));
         }
     }
 
+    private void ReadCharacterInventoryFromSave(XmlReader reader)
```
sboigelot: This method is never used (and empty)
You should delete it.

Grenkin1988:
Looks like working

Issue #1197, opened 2016-09-07
Repatriate localisation file or not


sboigelot:
Some time ago the localisation file have been moved to a different repository:

https://github.com/QuiZr/ProjectPorcupineLocalization/tree/Someone_will_come_up_with_a_proper_naming_scheme_later

This has been done at a time where localisation was new, a lot of PR were coming in and the gatekeeper were overflowed by them.

But currently it force everyone that add a single line of text to do it in a different PR, on a different repo.

Is this still something we want?

bjubes:
I think that we should eventually put them at TeamPorcupine/ProjectPorcupineLocalization so it is logical where the localization is while separating code from translation for gatekeepers and those following the repo.

mikejbrown:
What about making it a git submodule?

bjubes:
only downside to submodule is it requires the extra command to sync and update it, and there are a lot of git beginners in this project. It didnt work to well with uberlogger #641

sboigelot:
We should maybe just keep the en_US default in this repo at least
So we can make complete PR

koosemose:
Yeah, having at least the default en_US in this repo would be good, much easier to set what the default text should be from which other text should be based.

vogonistic:
I agree that the en_US translation should never be missing. I've emailed @quill18 and asked him to add the repo under `TeamPorcupine` as well.

Sommerbo:
I agree with @sboigelot, en_US should be in the main repo.

dusho:
anything new about this? I was wondering why my localization for en_US got lost.. guess it's still not part of main repo?

bjubes:
I assume were waiting on quill to make the localization repo. We don't want to add the file yet in case the version in this repo differs from Quizrs version.

QuiZr:
I've marked @quill18 sevral times when making my issue and then pr about this but got no answer so I left this as a "[temporary](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/Assets/Scripts/Localization/LocalizationDownloader.cs#L48)" thing  I've also noticed some bugs in that code so I will fix them in 2 or 3 days.

About that en_US - I think that all localization files should be attached when making a build - in travis or something.

kd7uiy:
I vote for moving it all back to the main repository. I think an admin might be able to pull in the commits that have been made to give credit where credit is due, but there's only 26 commits and 2 contributors to the other repository, not enough to keep them separate if you ask me.

[...]

PR #1207, opened 2016-09-07
Contributor name script

kd7uiy:
This python script generates the standard list of Character Names, which should be used for all future updates to the Character Names list


mikejbrown:
Why not [str.capitalize()](https://docs.python.org/3/library/stdtypes.html#str.capitalize) (also in 2.7)?

kd7uiy:
Because I didn't know that existed... Sigh.

sboigelot:
Would it be possible to add in the script an array that override default names? I would prefer having my surname in the game instead of the compress version of my fullname as username. That would be a great addition!

mikejbrown:
@kd7uiy  No worries. It's a big standard library. If you find yourself wanting to do something that seems like it should be pretty common, do a quick search of the library reference. It's probably in there! :grinning:

kd7uiy:
I did, which lead me to a SO question, with that code... Sigh.

As far as the exceptions, it can be done, I don't think we should do that for the base version. I'll think about it...

mikejbrown:
Another future TODO might be to call this on game load to always fetch the latest list...

kd7uiy:
The last could in theory be done, but I'd rather not have a web dependency to start the game. Plus, running the script takes a few seconds. Lastly, I don't want to pound the Github API. Making it as part of the build process might be a good idea, but that's as far as I would go.

sboigelot:
I'm with @kd7uiy, let's not add download at game start, we should already get rid of the one that downalod localisation, but that's another topic :)

mikejbrown:
I was about to say doesn't the localisation already do that.. :grinning: 

Could always do some caching.

kd7uiy:
I was trying to figure out why I didn't see the localization files in my branch! 

Honestly though, for now it doesn't matter much, and when it does matter, the list will be pretty constant, so...

sboigelot:
@kd7uiy check #1197 for more info

PR #1208, opened 2016-09-07
Fixes #1198 - Furniture GetSpriteName also checks animation for a sprite

### LINKED ISSUE #1198: Furnitures with Animations reset to base sprite when furniture is placed adjacent.
This is actually two (presumably related) bugs.

First, the furniture preview is based off of the base sprite of a furniture (So the sprite with the same name as the furniture), this is most obvious when placing a power generator, as it will show the No Texture pink square, since none of its sprites use the same name as the furniture ("Power Generator" vs. "Power-Generator-01" through "Power-Generator-04".

Second a furniture will reset to its base sprite (So for the power generator it will be no texture pink), when a furniture is placed adjacent to it, and will stay that way until something changes it (changes state, or changing during an animation).

The first is an easy fix, just by changing the name of the idle state sprite to match the name of the furniture, though perhaps it might be preferred to use the idle state sprite as the base sprite if it exists. The second I'm not sure how it would need fixed, though it is less of a problem if the base sprite is the idle state, at least with our currently animated sprites (it's either idle, in which case changing to idle is unnoticeable, or it's animating in which case it will only flash idle for a moment, and go back to animating), but the flicker seems undesirable, and if we ever have multiple states on an item that aren't animated (for example, a trash bin that shows as either empty or full), it will stay as whatever sprite matches the furniture name until a state change happens.

Raketfart:
See excellent bug description in #1198.

### DIFF
--- a/Assets/Scripts/Controllers/FurnitureSpriteController.cs
+++ b/Assets/Scripts/Controllers/FurnitureSpriteController.cs
@@ -53,7 +53,8 @@ public class FurnitureSpriteController : BaseSpriteController<Furniture>
     public Sprite GetSpriteForFurniture(string objectType)
     {
         Furniture proto = PrototypeManager.Furniture.Get(objectType);
-        Sprite s = SpriteManager.current.GetSprite("Furniture", objectType + (proto.LinksToNeighbour ? "_" : string.Empty));
+        string spriteName = proto.GetSpriteName();
+        Sprite s = SpriteManager.current.GetSprite("Furniture", spriteName + (proto.LinksToNeighbour ? "_" : string.Empty));
 
         return s;
     }
--- a/Assets/Scripts/Models/Animation/FurnitureAnimation.cs
+++ b/Assets/Scripts/Models/Animation/FurnitureAnimation.cs
@@ -12,6 +12,9 @@ using UnityEngine;
 
 namespace Animation
 {
+    /// <summary>
+    /// Animations for furniture. Can have several "states" that can be switched using SetState.
+    /// </summary>
     public class FurnitureAnimation
     {
         // current shown frame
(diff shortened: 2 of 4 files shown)


Raketfart:
I fixed it by making the furniture look for default sprites in the animation also.
So we don't need to worry about always using default names for idle states. 

The only thing is, that the first animation added, will determine what ghost image and starting image will be used. So idle animation should always be added first.

Tranberry:
> So idle animation should always be added first.

is that in the XML or on the sprite image?

Raketfart:
@Tranberry Yes, just in the furniture data xml.

TomMalbran:
@Raketfart If we want to fix the Ghost image using the idle one without having it first, we can just add a ```GetGhostSpriteName``` function in furniture and ```GetFirstIdleSprite``` In furniture animation.

sboigelot:
The fist animation in the list is a good fix for me, simple and we+moders don't have to think about it

Raketfart:
@TomMalbran I prefer that animation doesn't really need a specific idle animation, if that's irrelevant for some furniture. Maybe the modder would prefer to call it on/off state instead, and that's fine.

TomMalbran:
Right that is true. We can add a property for the ghost image. I am not sure if always using the first sprite is the best solution.

sboigelot:
@Raketfart currently the furniture doesn't need an idle animation, if there is no animation the basic sprite is still use a default. But I may be wrong in my understanding.

Raketfart:
No, you are correct. 
We could add a ghost image tag in the xml, if we feel it's needed

sboigelot:
Let's see if the artist team ask for it. Otherwise I feel it would be extra xml tag cluttering the file for not much.

PR #1209, opened 2016-09-07
[ART] Replace the RTG placeholder art

Tranberry:
screengrab

With the support of @Tritaris I draw a replacement sprite for the RTG that he made. This is using the 64 PPU base, and are closer to the perspective of most new art.


kd7uiy:
Looks nice, and pretty accurate! The fins are usually darker, as can be see from https://en.wikipedia.org/wiki/Radioisotope_thermoelectric_generator. In fact, the whole thing is usually darker. Not that it has to be pure black, but it might make it look a bit more realistic.

Tranberry:
@kd7uiy most reference pictures where actually white, like in the most recent Mars rover. I would have liked to make it darker.

kd7uiy:
There's a lot more color to them than I would have thought. Hmmm.... I see some black, some white, and some silver. https://www.google.com/search?q=radioisotope+thermoelectric+generator&biw=1920&bih=943&tbm=isch&tbo=u&source=univ&sa=X&sqi=2&ved=0ahUKEwirupHBwP3OAhXKFx4KHazLA00QsAQILw White cools better, black traps more heat, so I wonder if it depends on what the destination is. Hmmm...

sboigelot:
Dam it, this is the future, make it pink with purple dots for what we know :D (how as your art preference feels like)

Tranberry:
Maybe we can make it different by what material its made of (might be stupid but that's a little bit dorfy). Not that I would support having manually recolored sprites, much better to make the code do that.

kd7uiy:
I was interested enough about it I asked a question at Space Stack Exchange, will be interesting to see what if anything they come up with. I do have a black, white, and gray version, the white being the newer of the 3. http://space.stackexchange.com/q/18100/25

Raketfart:
Looks cool! Does it rotate like a fan? 
(Yes, I know nothing about RTG's)

Tranberry:
@Raketfart what a pain to animate, but no it's is currently static as I should be I think. maybe it should have a some light on it indicating if it is on or off ??

Tranberry:
but with access shouldn't we be able to reuse it with new fuel? (I know nothing)

kd7uiy:
Refueling it should not be possible. I guess a recycling thing could happen maybe? But that's really beyond this PR, but would make a very interesting issue to discuss.

Tritaris:
Yeah in theory we could have RTGs that are refillable. It would still be on the order of years before you would have to though. They are mainly designed to extremely low maintenance.

kd7uiy:
So I have it on semi-good authority that RTGs are darker if not intended for use inside an atmosphere, and lighter if intended for use inside of an atmosphere. It really could go either way here. I'll leave it up to you, but I thought you should know.

sboigelot:
You can provide both sprite, dark and light, we will toggle between them according to the placement on the map, inside or outside.

sboigelot:
@Tranberry tell us what you prefer to do with this, for me, we merge it whenever you feel ready

Tranberry:
@sboigelot I believe a solution like the one with the space suite would be the best solution. That is if someone like @Raketfart want to do that?

Raketfart:
Yeah, at some point i will add furniture color replacement through c# and lua. This could then be solved using that.

Tranberry:
So, please merge?

bjubes:
if you say so ;)

PR #1211, opened 2016-09-07
Deletes the unneeded Power.meta file.

kd7uiy:
The power.cs.meta file was added in #1158 without having an associated power.cs file. Deleting this meta file that currently isn't needed.


Raketfart:
Yeah, probably me - sorry. My git-fu is messed up, and I haven't even done anything with power.

kd7uiy:
No worries, stuff happens. The lack of the file was just annoying me, so I went back and traced through where it came from.

My git-fu is getting to be better, but I still make some pretty big mistakes from time to time... Luckily I am getting better overall.

PR #1214, opened 2016-09-07
Reduce nesting of if statements

Grenkin1988:


sboigelot:
tested, no stylecop error, ready to merge as soon as travis catch up

PR #1215, opened 2016-09-07
Fix default ppu

### LINKED ISSUE #1061: Default PPU in loader should be 64
As it stands, sprites are put at 32 PPU when no XML is provided. SInce we decided that 64 is the standard, the loader should reflect this too and assume 64 if no more info is given.

jaselito:
This should fix #1061


sboigelot:
tested, stylecop-ed, ready to merge when travis catch up

PR #1217, opened 2016-09-07
Update/SpriteToXML

Tritaris:
-  Changed default Pixel per Unit from 32 to 64 #1215, #1061
-  Added button to open /Resources/Editor/SpriteToXML/ in Unity based mode
-  Added check for /Resources/Editor/SpriteToXML/ creates it if not found
- Added entries to .gitignore


bjubes:
Merged!

PR #1219, opened 2016-09-08
[Updated] Character state machine refactoring

vogonistic:
**Warning:** I rebase often, which has a tendency to drops comments made on code. I'll try not to do it before I've dealt with any comments, but give no guarantees until I remove `[WIP]`.

There is a lot of changes in here and I'm interested in all kinds of feedback. The most important would be if there are more efficient and clever way to implement the FSM.

Changes in this PR:
 - All movement, job and need code has been moved into an FSM (finite state machine)
 - Pathfinding has a new interface that allows for more flexibility, like searching for an adjacent tile
 - waitingJobQueue has been integrated into JobQueue who also manage all the waiting jobs

Known issues:
- [x] Characters sometimes get "stuck" and stops performing jobs
- [x] Either the JobState or Stockpile has a bug. We get a spam of `Need to fetch 0 ice` in the log and the game gets really slow. Found by @dusho. Can probably be solved with `return FetchInventory(neededInventory.ObjectType, neededInventory.MaxStackSize - neededInventory.StackSize)` in `JobState.cs:187`
- [x] JobQueue.GetJob() needs to separate unreachable jobs from the available jobs. Figure out at least a temporary solution
- [x] Characters might not abandon jobs when someone steals the resources they want. Found by @Raketfart 
- [x] Jobs are not abandoned if there are no more resources of that type, but there was when they took the job. Found by @dusho 
- [x] Building floor tiles results in error: `Input string was not a correct format` Found by @Raketfart 
- [ ] Furniture can be built diagonally even though it's not a walkable diagonal. Found by @Raketfart #1248 
- [ ] Trapped characters still try to do jobs #1321 

Notes:
 - The JobState is more complicated than it will be when we integrate the new job system
 - There is a fair amount of debug logs still enabled in the code. Mostly through State.FSMLog.
 - Not all bugs have been hunted down yet, but the code is good enough for people to start taking a look.
 - Saving and loading is not implemented.

FSM Notes:
 - When Character.Update is run and there is no active state, it picks up any queued state first, then tries to get a job and ultimately falls back to IdleState for 0.2-2.0 seconds.
 - Each state can create new states (e.g. JobState can create a MoveState) if there is a dependency that needs to be filled. The new state has a reference NextState back to the creator, so that it can be started up again when the dependency has been fulfilled. Side note: This means the originating state is deepest in the chain, but it's slightly more efficient and less confusion since we only call Update() on the state that is currently working.
 - Needs (and in the future threat detection) is implemented through a list of global states that are always run, before the active state. They can create new JobStates to solve any problem they have and can either put it in the stateQueue for execution next time we look for something new to do, or can interrupt the current states (who are completely stopped and will not be picked up again).

Example - Stockpile_UpdateAction:
 1. `JobState` The character picks up a Job of type `Stockpile_UpdateAction` (i.e. carry something to the stockpile)
 1. `JobState` The character isn't carrying anything, so we find the first type that can the stockpile wants and that exists somewhere on the map.
 1. `JobState` Find the path to an inventory of the right type and create a `MoveState` where NextType is pointing back to the job.
 1. `MoveState` Walk to the inventory
 1. `MoveState` When arrived, pick up inventory and exit this state (goes back to `JobState`)
 1. `JobState` If stackSize is less than MaxStackSize and there is more of this inventory, then go to 3.
 1. `JobState` Drop off inventory, run `job.DoWork(deltaTime)`


sboigelot:
I just red a bit during my lunch time, I will read more later tonigth.

My initial impressions are good. I wouldn't call the current job JobState as the composition of this job is very specific to working a machine. The name is too generic.

It would also be good to have a composite creation of jobs. For instance I could see a furniture definition defining it's own job in xml/json/whatever with a queue of subtasks (this may be for a different PR of course, but we can keep it in mind to allow it in the fundation of the FSM)

```
<Job name="xxx">
     <SubStates>
        <HaulInventoryToJobTile ObjectType="cup of tea" Quantity="3"/>
        <WaitAtJobTile TimeInSeconds="15"/>
        <DestroyJobTileInventory/>
        <CreateInventotyAtJobTile ObjectType="empty sad cup" Quantity="3"/>
        <UpdateNeed Need="Thirst" Value="-15"/>
        <UpdateNeed Need="Pee" Value="+15"/>
     </SubStates>
</Job>
```

Raketfart:
I've just skimmed the code, and it looks really good.

I just have one question. If the need for food gets > 50 you queue a job for getting food. But if the character is really busy, he might reach a need of 100, and then you "setstate" a get food job. Won't he then eat twice in that case? Maybe we need to check if same need is in the queue when we "setstate".

vogonistic:
@sboigelot The way it works right now, everything but moving is implemented as a job. Mostly because it gives you a target position and then a callback to actually performing the job. One thing we could do with the upcoming job system is to rename Job to Task, then as the resolver breaks down the job, not entirely unlike what you describe, each action in there could be a state in it's own right if we want states to be more broken down.

We already have the two first tasks completely taken care of. HaulInventoryToJob is a list of required inventories and WaitAtJobTile is jobTime. What actually happens when you build the job could definitely be factory actions for things like building and destroying furniture and floor, and creating gear like better helmets, drills and those things that are more inventory than furniture.

Thirst and pee shouldn't be attributes on a job, it should be an effect of drinking tea.

vogonistic:
@Raketfart I'll document it better, but Interrupt() will interrupt and remove the current state, then clear out the queue as well, so it's safe queue and then interrupt if it gets bad. Before interrupting, we should probably check if the current state is related to solving this need. As it is, we might interrupt eating to eat.

The job that is created for a soft need is different from the hard need job, so we can't just prioritize it out of the queue.

If someone who actually understands the needs system wants to improve it and make it simpler, that would be awesome.

Grenkin1988:
Some observation while playing:
- Got strange freeze several times
- First char took not full stack of ice and went to another one but that stack was also targeted by other character. After first one "filled" his stack and went to stockpile, second character took not full stack an stuck. He was standing till first characted went to another stack.

vogonistic:
@Grenkin1988 Yeah, there are some bugs with handling of jobs that I haven't gotten figured out yet. Odd that the character didn't pick up the full pile the first time. I'll look into that too.

Assets/Scripts/Models/State/JobState.cs
```diff
@@ -0,0 +1,328 @@
...
+            // TODO: GetFirstFulfillableInventoryRequirement doesn't do the stockpile and lock check.
+            Inventory neededInventory = Job.GetFirstFulfillableInventoryRequirement();
+            if (neededInventory == null)
+            {
+                AbandonJob();
+                Finished();
+                return true;
+            }
+
+            return FetchInventory(neededInventory.ObjectType, neededInventory.StackSize);
```
dusho: so this should be probably:
`return FetchInventory(neededInventory.ObjectType, neededInventory.MaxStackSize - neededInventory.StackSize);`
code when character has something in Inv. is handled above that

dusho:
ye.. now dropping off inventory is broken in some cases
image

mikejbrown:
Just had a read through the main classes. This is looking great! :heart:

vogonistic:
After spending several evenings trying to get the bugs around performing jobs figured out, I've decided to break it out into a lot more states. It'll delay things a bit and make the PR a little bit bigger, but I think it's needed.

Anyone who wants to participate in getting it worked out or have opinions should ping me on discord and we can discuss it.

mikejbrown:
More simpler states is better, as long as it doesn't lead to an absurd number of transitions :+1:

dusho:
How I imagined things would work:
- dude picks up job from queue
- JobResolver determines what the initial state will be and switches to it
- after first state is done, other will follow from chain (Job instance will be used to store all the state data -> job instance will be passing from state to state)
- when all is done, job is gone
- when something gets wrong during some state, job may be en-queued again or even dropped (could be property of StateChangePlan), so e.g. SeekMaterialState fails to find material, it will just put job back to queue (flag the last execution time, to prevent firing often again and will be checked later on depending on frequency)

States would follow some StateChangePlan depending on what was initial impulse for work (outside action - mine, request from machine for input materials, need)
say definition of StateChangePlan for hauling input to machine:
- SeekMaterialState, MoveState, PickInventoryState, MoveState, DropInventoryState

vogonistic:
A lot of the JobState functionality has been split out into a HaulState and DumpState (which actually finds a tile to dump the inventory. yay). It seems to work more reliably, but needs a lot of testing.

vogonistic:
Fixed a couple of new issues. It'd be great if someone can test and see what is still broken.

vogonistic:
@dusho Other than splitting the states further, which we can do in a later PR, what do you think should be different about how the state framework?

For anyone reading this:
I also want to point out that this isn't intended to fix the job allocation problem or how that works, the only real change that I was looking for was to get a lot of the redundant and complicated job logic split out into something that makes more sense.

If we decide that this isn't a good way forward, the work logic could still be refactored into something similar to this but without a state machine. The pathfinding, job queue and inventory manager changes could all be useful changes on their own.

dusho:
tried your branch.. getting:

> ArgumentNullException: Argument cannot be null.
> Parameter name: key
> System.Collections.Generic.Dictionary`2[System.String,System.String].TryGetValue (System.String key, System.String& value) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System.Collections.Generic/Dictionary.cs:588)
> ProjectPorcupine.Localization.LocalizationTable.GetLocalization (System.String key, FallbackMode fallbackMode, System.String language, System.String[] additionalValues) (at Assets/Scripts/Localization/LocalizationTable.cs:132)
> ProjectPorcupine.Localization.LocalizationTable.GetLocalization (System.String key, System.String[] additionalValues) (at Assets/Scripts/Localization/LocalizationTable.cs:64)
> Job.GetName () (at Assets/Scripts/Models/Job.cs:466)
> ProjectPorcupine.State.JobState..ctor (.Character character, .Job job, ProjectPorcupine.State.State nextState) (at Assets/Scripts/Models/State/JobState.cs:31)
> Character.Update (Single deltaTime) (at Assets/Scripts/Models/Character.cs:298)
> World.TickEveryFrame (Single deltaTime) (at Assets/Scripts/Models/World.cs:182)
> TimeManager.InvokeEvent (System.Action`1 eventAction, Single time) (at Assets/Scripts/Controllers/TimeManager.cs:198)
> TimeManager.Update (Single time) (at Assets/Scripts/Controllers/TimeManager.cs:131)
> WorldController.Update () (at Assets/Scripts/Controllers/WorldController.cs:142)
is this related to en_US not being in main repo still?
other than that, implementaion always looks better if you have code spread out to correct places (like states now)
FSM code looks good.. I think it will be easier to debug and fix the job system problems.. 
You want to fix those here in this PR?

vogonistic:
@dusho I added names to all the Jobs so I can log them, but if there is no name this is what the localization spits out. Do you know what machine you had just built that created this job?

I figure that if we go this route, we'll merge this since as a PR it's big enough as is and then we go straight to working on the job system itself.

All the FSM logs are available in the UberLogger window (In the window menu in Unity). If we decide to go this route, I'm going to have all the FSM logs turned off by default.

vogonistic:
Last night I created a new class `RequestedItem` that contains the requirements a Job has to solve 2 problems:
  1. Inventory instances are a bit heavy for the requirements. Especially for Stockpile jobs that allow any kind of inventory right now
  1. We needed a MinAmountRequired so that we know how to handle the case where a stockpile asks for 0 to 50 ice but there is only 15 available, as opposed to a wall needing 5 steel plates, but there is only 1 available.

There was also some logging changes to avoid errors. The diagonal build bug still exists and `RequestedItem` isn't fully used yet.

dusho:
@vogonistic is this still worked on and to be used?
When tested, getting:

> chunk_1:(215,4-43): attempt to index a nil value
> UnityEngine.Debug:LogError(Object)
> UberLogger.Logger:ForwardToUnity(Object, LogSeverity, Object, Object[]) (at Assets/UberLogger/UberLogger.cs:506)
> UberLogger.Logger:Log(String, Object, LogSeverity, Object, Object[]) (at Assets/UberLogger/UberLogger.cs:452)
> Debug:ULogErrorChannel(String, String, Object[]) (at Assets/UberLogger/Debug.cs:322)
> LuaFunctions:Call(String, Object[]) (at Assets/Scripts/Models/Functions/LuaFunctions.cs:90)
> Job:DoWork(Single) (at Assets/Scripts/Models/Job.cs:280)
> ProjectPorcupine.State.HaulState:Update(Single) (at Assets/Scripts/State/HaulState.cs:88)
> Character:Update(Single) (at Assets/Scripts/Models/Character.cs:308)
> World:TickEveryFrame(Single) (at Assets/Scripts/Models/World.cs:187)
> TimeManager:InvokeEvent(Action`1, Single) (at Assets/Scripts/Controllers/TimeManager.cs:198)
> TimeManager:Update(Single) (at Assets/Scripts/Controllers/TimeManager.cs:131)
> WorldController:Update() (at Assets/Scripts/Controllers/WorldController.cs:146)

Other issues:
- framerate drops when hauling
- all inventories brought to stockpile are gone (not placed, just removed)

vogonistic:
It takes too long to merge in the changes that happens on master so I'm extracting the non-FSM changes into separate PRs in hope of reducing the time I spend merging instead of fixing the implementation. 

It'd be helpful if you go look at the pathfinder interface refactoring #1350 and the requested item #1376 PRs.

koosemose:
Numerous changes have been made, in addition to resolving conflicts. I think I've gotten most (hopefully all) of the bugs squashed. Would appreciate further testing to see if any others can be found. 

Note: There is some degree of frame dropping, I'm pretty sure this is due to all of the debugging statements which I am leaving in for now (to make debugging easier) but am looking to implement a solution vog used elsewhere to allow them to be turned off en masse when not needed.

vogonistic:
Awesome job @koosemose! This is what I was talking about: https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/Assets/Scripts/Pathfinding/Pathfinder.cs#L205

PR #1220, opened 2016-09-08
Spawn inventory dev. dialog now allows to pick quantity

jaselito:
Now we can pick an amount of inventory spawned when using the developer spawn inventory dialog

spawn inventory dialog

### DIFF
--- a/Assets/Scripts/Controllers/SpawnInventoryController.cs
+++ b/Assets/Scripts/Controllers/SpawnInventoryController.cs
@@ -6,6 +6,7 @@
 // file LICENSE, which is part of this source code package, for details.
 // ====================================================
 #endregion
+using System;
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.UI;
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Controllers/SpawnInventoryController.cs
```diff
@@ -86,60 +89,90 @@ private void CreateSpawnUI()
...
-            colorBlock.normalColor = Color.white;
-            colorBlock.pressedColor = Color.blue;
-            button.colors = colorBlock;
             string localName = invName;
-            button.onClick.AddListener(
-                () => OnButtonClick(localName));
 
-            CreateTextComponent(inventoryButton_go, invName);
+            CreateTextComponent(inventorySlot_go, invName, TextAnchor.MiddleLeft);
+            CreateButtonComponents(inventorySlot_go, localName, new int[] { 1, 20, 50 });
```
Grenkin1988: It would be good instead of magic 1, 20, 50 to connect it to maxStack of `Inventory`

jaselito: I thought of that, but this being a dev tool decided to leave the chance to go up to 50 if you want.  As it is, if you want to get a heap with maxStack, you just need to click a tile twice with the 50 option (the game will automatically resize it to the maxStack value in that case)

bjubes:
Merged!

PR #1014, opened 2016-08-29
[small] Adds ProjectSettings/GraphicsSettings.asset to .gitignore

mikejbrown:
Stops spamming the working tree with unimportant changes

PR #1225, opened 2016-09-08
[small] Fix gitignored ProjectSettings/GraphicsSettings.asset

mikejbrown:
So PR #1014 added ProjectSettings/GraphicsSettings.asset to the project .gitignore but it is still apparently being tracked by the repo.

Just needed to follow the steps [here](https://stackoverflow.com/questions/11451535/gitignore-not-working/11451731#11451731) to fix the issue.

This deletes the file, but the next time you run Unity it will be recreated and properly ignored this time. You can see that it reappears if you do `git status --ignored`.


koosemose:
Merged

PR #1227, opened 2016-09-08
Stat class cleanup

Grenkin1988:


dusho:
.. I don't like using only Type as I often think of System.Type when seeing it in intellisense.. (but just personal preference)

Grenkin1988:
@dusho I cannot even imagine how I could think of this as class `Type`
When you see something like:
```
Stat stat;
stat.Type == "Something"
```
how it could be something like `typepf(Stat)` ?

But I have one more solution: `public string ID { get; private set; }`

vogonistic:
I'm 👍  on `Type`. Class name as prefix inside that class gets redundant so fast.

PR #1228, opened 2016-09-08
Inventory refactoring and cleanup

Grenkin1988:


Assets/Scripts/Models/InventoryManager.cs
```diff
@@ -6,100 +6,89 @@
...
+    private static readonly string InventoryManagerLogChanel = "InventoryManager";
 
     public InventoryManager()
     {
-        inventories = new Dictionary<string, List<Inventory>>();
+        Inventories = new Dictionary<string, List<Inventory>>();
     }
 
-    public bool PlaceInventory(Tile tile, Inventory inv)
+    public Dictionary<string, List<Inventory>> Inventories { get; private set; }
```
TomMalbran: I think that Inventories should be completely private. I can't switch the dictionary for another one, but I can still grab the instance and add/remove items without using the proper functions. For anything that is using the dictionary directly, we can add a method.

Assets/Scripts/Models/InventoryManager.cs
```diff
@@ -121,89 +110,74 @@ public bool PlaceInventory(Character character, Inventory sourceInventory, int a
...
-        if (QuickCheck(objectType))
+        foreach (Inventory inventory in Inventories[objectType].ToList())
         {
-            foreach (Inventory inventory in inventories[objectType].ToList())
+            if (onlyFromStockpiles)
             {
-                if (onlyFromStockpiles)
+                if (inventory.Tile == null || 
+                    inventory.Tile.Furniture == null ||
+                    inventory.Tile.Furniture.ObjectType != "Stockpile")
```
TomMalbran: We should probably use here ```inventory.Tile.Furniture.HasTagType("Stockpile") == false```  to support future types of Stockpiles.

Assets/Scripts/Controllers/TradeController.cs
```diff
@@ -109,7 +109,7 @@ private void TrasfertTradedItems(Trade trade, Vector3 tradingCoordinates)
             }
             else if (tradeItem.TradeAmount < 0)
             {
-                WorldController.Instance.World.inventoryManager.QuickRemove(tradeItem.ObjectType, -tradeItem.TradeAmount, true);
+                WorldController.Instance.World.inventoryManager.RemoveInventoryOfType(tradeItem.ObjectType, -tradeItem.TradeAmount, true);
```
TomMalbran: We could use World.Current here.

Assets/Scripts/Models/InventoryManager.cs
```diff
@@ -6,100 +6,89 @@
...
+            if (Inventories.ContainsKey(tile.Inventory.ObjectType) == false)
             {
-                inventories[tile.Inventory.ObjectType] = new List<Inventory>();
+                Inventories[tile.Inventory.ObjectType] = new List<Inventory>();
             }
 
-            inventories[tile.Inventory.ObjectType].Add(tile.Inventory);
+            Inventories[tile.Inventory.ObjectType].Add(tile.Inventory);
 
             World.Current.OnInventoryCreatedCallback(tile.Inventory);
```
TomMalbran: Since there is an ```InventoryManager``` that handles all the Inventories in the World, the ```inventoryCreated``` event should be part of the ```InventoryManager```. Only the ```InventoryManager``` calls this method in world, and world only executes it when that method is call.

Assets/Scripts/Models/InventoryManager.cs
Grenkin1988: Agree but this is too much for one PR

TomMalbran: OK. We can make an issue and fix it in another PR. Or add a TODO. Or nothing if you want to do it soon. But something if not, so that we don't forget.

Assets/Scripts/Models/Inventory.cs
```diff
@@ -191,4 +159,13 @@ private void ImportPrototypeSettings(int defaulMaxStackSize, float defaultBasePr
             Category = defaultCategory;
         }
     }
+
+    private void OnStackSizeChanged(Inventory inventory)
```
TomMalbran: Would calling this something like TriggerStackSizeChanged make more sense, since in a way you want to send a StackSizeChanged event, and not do something when the StackSizeChanged. I might be wrong though.

Assets/Scripts/Models/InventoryManager.cs
```diff
@@ -213,26 +189,31 @@ public Path_AStar GetPathToClosestInventoryOfType(string objectType, Tile t, int
...
-                inv.Character = null;
-            }
+        if (inventory.Tile != null)
+        {
+            inventory.Tile.Inventory = null;
+            inventory.Tile = null;
+        }
+    }
+
+    protected virtual void OnInventoryCreated(Inventory inventory)
```
TomMalbran: Does this need to be protected and virtual? Only private would work fine.

Assets/Scripts/Models/Inventory.cs
Grenkin1988: This is default name that VS gives to this method.

TomMalbran: Ok. It feels a bit weird to me, since you are not using it to attach it to an event and having it work when something changes. One thing should be to trigger/send an event and other to receive it. But if we go for the standard, that is fine.

Grenkin1988: I would recond to "read" this method as - This is the place where all On.... things would happen.

But if someone more think that this is not good - I will change it to `InvokeStackSizeChanged`

Assets/Scripts/Models/InventoryManager.cs
Grenkin1988: No it does not. This is just default template.

Assets/Scripts/Models/Inventory.cs
TomMalbran: I usually go for the Trigger prefix when sending an event, and On when receiving it. I think it makes it more clear.

Grenkin1988: `handler(inventory);` -> `handler.Invoke(inventory);`
That's why `Invoke...`

TomMalbran: Invoke is fine, my issue was with On.

Grenkin1988:
Looks like merge with master created a bug that is not in master :-(

dusho:
can 'Locked' be persisted, so it will keep the state after save/load? or should it be in different issue

Grenkin1988:
@dusho Do not know what it suppose to do. But still this would be done later.

dusho:
Inventory.Locked is used when hauling items as input to machines (workshops), after it's placed at machine input, it is locked, to prevent other dudes to steal it from there
this flag is lost after save.. so dudes start stealing stuff from machine inputs

Grenkin1988:
This hit lots of lua files probably. Because now right at start of the game character go to stockpile and places there a lot of iron.

TomMalbran:
Notice that #1219 will make it so that characters won't be able to take inventory from a furniture, without using Locked. But we could also later on make Locked what determines if a character can grab an inventory instead of checking for multiple conditions.

Grenkin1988:
Bug that I made during development - fixed

vogonistic:
The locked check is just ported over, but we've always had it. When the stack is the input hopper for a furniture, we lock it so that it's not available for other machines (or in worst case, for itself)

Grenkin1988:
@dusho @TomMalbran @vogonistic O.K. so what should I do with this `Locked`? I would prefere to leave it as it was before. Or I broke something?

vogonistic:
It wasn't broken by this PR, so I'm good with either:
 1. Open a separate bug that it's not doing it right and leave this PR as is
 2. Fix it in this PR

Grenkin1988:
@dusho @TomMalbran @vogonistic Could someone make bug issue for this, cause I do not really undestand the problem, or I just want to sleep :-(

vogonistic:
I've filed a bug on it.

Grenkin1988:
@vogonistic Does this solves this? ^

vogonistic:
As far as I'm concerned. @dusho @TomMalbran Are you guys ok too?

dusho:
yes sure.. locked bug can be solved elsewhere 
just noticed that you switched order of stackSize and maxStackSize in Inventory constructor.. you tested whether machines still get resources and stuff ? Aren't there some LUA scripts that need updating because  of that ?

Grenkin1988:
Hopefully no bugs left. But it needs to be tested.

dusho:
from what I tried, inventory still works, stacking, hauling
there seems to some unrelated problem with power of machines.. everything is just offline.. or am I doing something wrong..

sboigelot:
Do we still need to keep the
```
      public int MaxStackSize { get; set; }
  
      public float BasePrice { get; set; }
  
     public string Category { get; set; }
```

As part of the Inventory? It's duplicate information from the InventoryCommon. We could probably get around by not copying it at all.

Or do you think that they may change according to external factor per Iventory?

For instance, damage reduce basePrice
Time add rot to the category for food? etc...

Grenkin1988:
@sboigelot For next step I am going to get rid of `InventoryCommon` or do some inheritence. This in part 1 of refactoring I would say. I just did not what to touch all the things. Despite that I already did that :-(

TomMalbran:
@Grenkin1988 The InventoryCommon works as a prototype kind of thing. At one point we didn't copy all the stuff from the InventoryCommon, but now that we do we should totally remove it. I can do it, if you are not.

Grenkin1988:
@TomMalbran As I said I am going to continue refactoring of `Inventory` with `InventoryCommon` removal

PR #1229, opened 2016-09-08
Need class cleanup

Grenkin1988:


Assets/Scripts/Models/Character.cs
```diff
@@ -376,7 +376,7 @@ public void WriteXml(XmlWriter writer)
         foreach (Need n in needs)
         {
             int storeAmount = (int)(n.Amount * 10);
-            needString = needString + n.needType + ";" + storeAmount.ToString() + ":";
+            needString = needString + n.Type + ";" + storeAmount.ToString() + ":";
```
TomMalbran: We are using ObjectType for Furniture and Inventory, StatType for Stats and Type for Needs, and maybe others for other clases. I think we need some sort of standard. Is it Type or ObjectType for all, or ClassNameType?

Grenkin1988: I was thinking: "It is a little dully to have `needType` in `Need` class."
I mean, it is quite obvious that `Type` that is property of `Need`, is - `NeedType`

TomMalbran: Type would work good for all the clases, and even ObjectType too. But the inconsistency in the names is not great, since you need to remember which one is used for each class, or just check the class.

Grenkin1988: I would like to change all of them in other classes, but not in one PR

TomMalbran: Sure. That sounds good.

vogonistic:
This looks good to me, but it has conflicts.

Grenkin1988:
@vogonistic Resolved

koosemose:
No new problems introduced, Stylecop clean, looks good. Merged.

Issue #664, opened 2016-08-23
Xml Elements & Attributes Naming _conVenTions


sboigelot:
We currrently have a lot of inconsistency in the structure and naming convention of elements and attributes in XML.

This is a non-issue for now, but as the code base grows, a lot of error may arise from it. (and this will also confuse moders a lot)

![image](https://cloud.githubusercontent.com/assets/6545998/17885307/fec2c9de-691c-11e6-8fd6-32d6bb8c9465.png)

Do we have a naming (and structure?) convention from XML file? If not, should we have one?

Tranberry:
I who thought that pascal case was camel case and vice versa.

sboigelot:
Yep reverse it, dyslexia kills my brain sometimes, but the concept stay the same :)

koosemose:
Yeah you've got your camelCase and PascalCase switched. But yeah consistency is important, particularly in the names  and casing. I tend to prefer Pascal for Elements, and camel for Attributes (as it is in Furnitures).

I'm not sure what you mean with the SubElement part, but I've solely had my nose stuck in Furnitures for a while, and am quite a few commits back where I'm working. Could you provide a bit more of an example of what you mean on the SubElements?

sboigelot:
It's hard to explain, let's say you have a room with furniture in it

You can represent it by

(my naming: flat structure)
```
[room]
    [furniture name="n1"/]
    [furniture name="n1"/]
    [furniture name="n1"/]
[/room]
```

or

(my naming: elementParent_subElement)
```
[room]
    [room_furniture name="n1"/]
    [room_furniture name="n1"/]
    [room_furniture name="n1"/]
[/room]
```

or

(my naming and my personal :heart:  favorite: SubElement)
```
[room]
    [furnitures]
        [furniture name="n1"/]
        [furniture name="n1"/]
        [furniture name="n1"/]
    [/furnitures]
[/room]
```

koosemose:
Ahh I see, I think I am also fond of the SubElement version. Furnitures seems to match that style (though I may have overlooked something), it just has a large number of singular unrelated elements in it.

sboigelot:
@koosemose furniture is a mix bag

For instance the onUpdate lua Function are "flat structure"

koosemose:
[...]

PR #1230, opened 2016-09-08
Fixed XML files layout

Grenkin1988:


TomMalbran:
So the standard in XML is 2 spaces instead of 4 that we use in every other type of file?

Grenkin1988:
@TomMalbran You are right. Fixed to be 4 spaces.

dusho:
so it should be 4 spaces afterall? just now changed it to 2 from 4.. to fit furniture.xml.. so putting it back, I guess

TomMalbran:
I think 4 is the best. Makes it easier to read and is the same for all the files. Also in editor that support xml, you don't need to change the settings just for those files.

@Grenkin1988 Should we also change all the attributes to camelCase? There are a few that are in PascalCase. And all the Tag names to PascalCase, Trader.xml has some as camelCase.

Grenkin1988:
@TomMalbran Yes. But not in this go. Create issue and I or somebody will do this

TomMalbran:
Here is the issue #664, which was closed but never fixed.

Grenkin1988:
@TomMalbran Reopened

Grenkin1988:
Should go after #1229

vogonistic:
👍

koosemose:
All good. Merged.

PR #1232, opened 2016-09-08
Added black outlines and shading to walls sprite

### LINKED ISSUE #1142: Black outlines for wall sprites
This is something that has been bugging me... in order to be consistent with the rest of the artwork, I think  the wall sprites should also have a black outline... I did a little editing in GIMP to show you a before and after of what I mean (tried to do it in-game by modifying the actual sprite but couldn't get the outlines in the right places... maybe someone more familiar with the configuration of the wall sprite?)

Before: 
wall sprites no outline

After:
wall sprites outline

The actual outline should be thicker than what I did here and maybe even have some antialiasing, but you get the idea... (this should also apply to the doors btw)

Any thoughts?

jaselito:
Most rendering artifacts should disappear once #1231 is merged.

Closes #1142


Grenkin1988:
@jaselito Please add scenes from game. So people could see how it looks in game

jaselito:
Sure!

screenshot2

screenshot1

jaselito:
I'd like to point out that #1231 was merged, so the rendering artifacts that can be seen in the pictures should be gone now.

Tranberry:
Looks fine too me.

PR #1233, opened 2016-09-08
[small] Moved the interfaces to a new folder

TomMalbran:
For a better file organization and a place to hold more interfaces in the future.


Grenkin1988:
I've never had separate folder/namespace for Interfaces. That's a little bit strange

TomMalbran:
I haven't worked mucho with interfaces. I just figure it would make sense. But if it doesn't is ok. We can just close this PR.

Grenkin1988:
@TomMalbran I wouldn't do that. But other GateKeeper could have another opinion

sboigelot:
I wouldn't do that either.

dusho:
it usually makes sense to move interfaces to different namespace (even to separate assembly) if you're providing API to some outside world.. but in this project it really doesn't make sense to have separate namespace

sboigelot:
Closing this issue then as we seem to agree.

PR #1234, opened 2016-09-08
[small] Makes the Scheduler class visible to Lua scripts

mikejbrown:
Previously there was no way for Lua scripts to access the Scheduler
directly. They could create or modify ScheduledEvent objects but not
actually (de)schedule them! This was a simple oversight on my part. The
intent for Lua access was there from the beginning.

Fixing this was simply a matter of adding a [MoonSharpUserData]
declaration and registering the globals. Now any Lua function run through
LuaFunctions.Call() or CallWithInstance() automatically gets the Scheduler
class and (most importantly) Scheduler.Current in its globals.

On commit all tests and Stylecop were passing.

### DIFF
--- a/Assets/Scripts/Models/Functions/LuaFunctions.cs
+++ b/Assets/Scripts/Models/Functions/LuaFunctions.cs
@@ -32,6 +32,8 @@ public class LuaFunctions
         RegisterGlobal(typeof(World));
         RegisterGlobal(typeof(WorldController));
         RegisterGlobal(typeof(Connection));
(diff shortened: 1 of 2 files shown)


TomMalbran:
I think that you also need to add ```[MoonSharpUserData]``` in ```ScheduledEvent``` to be able to use it.

mikejbrown:
@TomMalbran It's already there https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/Assets/Scripts/Scheduler/ScheduledEvent.cs#L38

PR #1235, opened 2016-09-09
Fixed some warnings and StyleCop errors

TomMalbran:
1. Fixed some warnings when launching the game.
2. Fixed an issue where you couldn't place a solar panel over ladders.
3. Fixed some StyleCop issues.


TomMalbran:
Conflicts fixed.

PS: Github doesn't send email notifications on title changes.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -1105,7 +1104,7 @@ private bool DefaultIsValidPosition(Tile tile)
                 }
 
                 // Make sure tile is FLOOR
-                if (t2.Type != TileType.Floor && tileTypeBuildPermissions.Contains(t2.Type) == false)
+                if (t2.Type != TileType.Floor && tileTypeBuildPermissions.Contains(t2.Type.Type) == false)
```
Grenkin1988: It should work without this.
`TileType` already knows how to compere it with `TileType`:
```
public bool Equals(TileType other)
{
    return string.Equals(Type, other.Type);
}

public override bool Equals(object obj)
{
    return Equals((TileType)obj);
}
```

TomMalbran: Notice that ```tileTypeBuildPermissions``` is now a HashSet of strings because when the prototypes are loaded we shouldn't assume that other prototypes are loaded and is worst with TileTypes since is not even in the PrototypeManager (i need to fix that :P), and not TileTypes as before. I don't think that comparing a string with a TileType will work.

koosemose:
Conflicted

TomMalbran:
Conflicts fixed

Grenkin1988:
@TomMalbran Please fix this also:
`153:        tileTypeBuildPermissions = other.tileTypeBuildPermissions;`
to:
`tileTypeBuildPermissions = new HashSet<string>(other.tileTypeBuildPermissions);`

TomMalbran:
@Grenkin1988 Sure. Done.

Grenkin1988:
Merged

PR #1243, opened 2016-09-09
Stylecop Room.cs

koosemose:


Grenkin1988:
@koosemose You could merge such PR yourself :-)

PR #1245, opened 2016-09-09
[small] Add a default zoom value on camera..

Ohda:
This allow to change the main camera initial zoom and fix a bug with empty value on start.
I have set the initial value higher than actually.


Tranberry:
will I be able to set it so that 1px actually is 1px on screen?

Ohda:
Hum, i don't know. Mabye someone else will.
The camera orthographics size have a range set here. So it may acually go from 3 to 25.
`Camera.main.orthographicSize = Mathf.Clamp(target, 3f, 25f);`

koosemose:
No unexpected problems, looks good, Stylecop clean. Merged

PR #1246, opened 2016-09-09
Character animation spring cleaning and bug fixing

Raketfart:
This fixes a couple of bugs.

The character facing wasn't being updated, when encountering a door, and waiting for it to open.

The character inventory sortorder was sometimes flipping in front and behind the character.

Then there is some cleanup of the animations. 
The characterAnimation now uses the same "SpriteNameAnimation" as FurnitureAnimation. So the sprites are back in the hands of the SpriteManager, instead of being stored in the animation.


koosemose:
I'm not entirely sure if this is caused by your code, or just made evident due to your updates. But at certain times (most often when leaving the top airlock door, and heading directly to the left), the character will end up facing odd ways, either when waiting on the door (several times facing to the right even though they came from the right and will be heading left), or when going through the door (appear to walk through the door facing to one side or another), and occasionally a tile or two after leaving the door (continue facing up, when moving directly to the left).

Raketfart:
I thought i fixed that. I'll take a look again later. Thanks

Raketfart:
Found the bug, and I think it means that this glitch could also have happened with furniture. 
Never again.

Assets/Scripts/Controllers/CharacterSpriteController.cs
```diff
@@ -103,10 +102,13 @@ protected override void OnChanged(Character c)
...
         // Make sure the furniture's graphics are correct.
         SpriteRenderer inv_sr = objectGameObjectMap[c].transform.GetChild(0).gameObject.GetComponent<SpriteRenderer>();
+
+        // Important to set the characters SortOrder first.
+        int charSortOrder = c.animation.SetAndGetSortOrder();
         if (c.inventory != null)
         {
             inv_sr.sprite = SpriteManager.current.GetSprite("Inventory", c.inventory.GetName());
-            inv_sr.sortingOrder = c.animation.CurrentSortingOrder + 1;
+            inv_sr.sortingOrder = charSortOrder + 1;
```
TomMalbran: Don't add trailing spaces.

Raketfart:
Extra spaces removed now. :/

Grenkin1988:
Conflicts

Raketfart:
Should be fixed now

Raketfart:
I don't know why the heater file is renamed. I've never touched those files. How do I remove that change from my PR?

sboigelot:
@Raketfart it's because it has been renamed from heater->Heater on master and your auto merge reverted it.

Just rename it manually to Heater, it's the simpliest way.

Raketfart:
They were Capitalized in my files. I think I got rid of all the weirdness now

PR #1256, opened 2016-09-10
Fixes heater quest bug

kd7uiy:
The heater quest cannot be finished without this small fix.


bjubes:
why is heater lowercase while every other furniture is capitalized?
Is heater just named inconsistently?

kd7uiy:
Because... Reasons? I agree with you, I'll switch it around.

sboigelot:
This branch is breaking the Heater sprite
image

It's still working on master

Grenkin1988:
@kd7uiy You should rename this also
image

kd7uiy:
That's what I get for not properly testing it... Thanks guys!

PR #1259, opened 2016-09-11
New CharacterNameManager

TomMalbran:
I moved the loading of the character names into a new CharacterNameManager, from where you can get new names.

When loading names they are randomized and added to an unused names queue. Each time you ask for a new name, it retrieves the first from the list. This makes it so that you don't get repeated names, until all the names are used, and we start using all the used names again.

This also makes it possible for a mod to add its own names.

### DIFF
--- a/Assets/Scripts/Controllers/ModsManager.cs
+++ b/Assets/Scripts/Controllers/ModsManager.cs
@@ -82,7 +82,8 @@ public class ModsManager
     }
 
     /// <summary>
-    /// Loads the given file from the given folder in the base and inside the mods and calls the Delegate.
+    /// Loads the given file from the given folder in the base and inside the mods and
+    /// calls the Action with the file path.
     /// </summary>
     /// <param name="folderName">Folder name.</param>
     /// <param name="fileName">File name.</param>
--- a/Assets/Scripts/Controllers/ModsManager.cs
+++ b/Assets/Scripts/Controllers/ModsManager.cs
@@ -25,12 +25,12 @@ public class ModsManager
 
     public void LoadPrototypes()
     {
-        LoadFunctionScripts("Furniture", "Furniture.lua");
-        LoadFunctionScripts("Need", "Need.lua");
-        LoadFunctionScripts("GameEvent", "GameEvent.lua");
-        LoadFunctionScripts("TileType", "Tiles.lua");
-        LoadFunctionScripts("Quest", "Quest.lua");
-        LoadFunctionScripts("ScheduledEvent", "ScheduledEvent.lua");
+        LoadFunctions("Furniture", "Furniture.lua");
+        LoadFunctions("Need", "Need.lua");
+        LoadFunctions("GameEvent", "GameEvent.lua");
+        LoadFunctions("TileType", "Tiles.lua");
+        LoadFunctions("Quest", "Quest.lua");
+        LoadFunctions("ScheduledEvent", "ScheduledEvent.lua");
 
         PrototypeManager.Furniture.LoadPrototypes(mods);
         PrototypeManager.Inventory.LoadPrototypes(mods);
@@ -39,6 +39,8 @@ public class ModsManager
         PrototypeManager.SchedulerEvent.LoadPrototypes(mods);
         PrototypeManager.Stat.LoadPrototypes(mods);
         PrototypeManager.Quest.LoadPrototypes(mods);
+
+        LoadCharacterNames("CharacterNames.txt");
     }
 
     public DirectoryInfo[] GetMods()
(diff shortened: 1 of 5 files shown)


Assets/Scripts/Models/CharacterNameManager.cs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /// <summary>
+    /// Load new names to use for Characters and add them in a random way.
+    /// </summary>
+    /// <param name="nameStrings">An array of name strings.</param>
+    public static void LoadNames(string[] nameStrings)
+    {
+        // Randomize the given strings
+        for (int i = 0; i < nameStrings.Length; i++)
```
sboigelot: This is very verbose and low level coding.

What about nameStrings.OrderBy(Unity.Random.Range(0f, 1f)) to replace this whole block of code?

Assets/Scripts/Models/CharacterNameManager.cs
```diff
@@ -0,0 +1,69 @@
...
+        foreach (string name in nameStrings)
+        {
+            unusedNames.Enqueue(name);
+        }
+    }
+
+    /// <summary>
+    /// Gets a name which wasn't used before.
+    /// </summary>
+    /// <returns>Psossibly an unused name.</returns>
```
mikejbrown: Possibly -- also this conflicts a bit with the summary. Could even drop the returns tag... I would suggest something like
```c#
/// <summary>
/// Returns a randomly chosen name, prioritizing names which have not been used yet.
/// </summary>
```

Assets/Scripts/Models/CharacterNameManager.cs
```diff
@@ -0,0 +1,69 @@
...
+// This program comes with ABSOLUTELY NO WARRANTY; This is free software,
+// and you are welcome to redistribute it under certain conditions; See
+// file LICENSE, which is part of this source code package, for details.
+// ====================================================
+#endregion
+using System.Collections.Generic;
+using UnityEngine;
+
+/// <summary>
+/// Character name manager that holds all the possible names and tryes to give a random and unused name everytime is requested.
```
mikejbrown: tries

Assets/Scripts/Models/CharacterNameManager.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// Initializes a new instance of the <see cref="CharacterNameManager"/> class.
+    /// </summary>
+    public CharacterNameManager()
+    {
+        unusedNames = new Queue<string>();
+        usedNames = new Queue<string>();
+    }
+
+    /// <summary>
+    /// Load new names to use for Characters and add them in a random way.
```
mikejbrown: Don't need the second half if you change the doc for `GetNewName` as suggested. Just "Load new names to use for Characters." The fact that you do the randomization here and not in `GetNewName` is an invisible implementation detail which may change without changing the behavior in any important way.

Assets/Scripts/Controllers/ModsManager.cs
```diff
@@ -47,27 +49,59 @@ public DirectoryInfo[] GetMods()
...
+            });
+    }
+
+    /// <summary>
+    /// Loads the given file from the given folder in the base and inside the mods and calls the Delegate.
+    /// </summary>
+    /// <param name="folderName">Folder name.</param>
+    /// <param name="fileName">File name.</param>
+    /// <param name="readText">Called to handle the text reading and actual loading.</param>
+    private void LoadTextFile(string folderName, string fileName, Action<string> readText)
```
mikejbrown: Nice refactoring

Assets/Scripts/Controllers/ModsManager.cs
```diff
@@ -47,27 +49,59 @@ public DirectoryInfo[] GetMods()
...
+            fileName,
+            (filePath) =>
+            {
+                string[] lines = File.ReadAllLines(filePath);
+                CharacterNameManager.LoadNames(lines);
+            });
+    }
+
+    /// <summary>
+    /// Loads the given file from the given folder in the base and inside the mods and calls the Delegate.
```
mikejbrown: Would say something like "...and calls the delegate with the text of each file."

Assets/Scripts/Models/CharacterNameManager.cs
TomMalbran: Sure. Had no idea I could do that.

TomMalbran:
Fixed the comments and the randomizer. Thanks for the comments.

Assets/Scripts/Models/CharacterNameManager.cs
TomMalbran: @sboigelot I tried your code and it didn't work. I searched for a way to do it, and the threads suggested to implement Shuffle, which is basically what I did before. So I reverted the change. If you know a way to make it work, let me know.

sboigelot: Check https://github.com/sboigelot/TwentyCardsToTheApocalypse/blob/master/Assets/Scripts/Models/Deck.cs

Lines

```
        public void Shuffle()
        {
            Cards = Cards.OrderBy(c => Random.Range(0f, 1f)).ToList();
        }
```

But it's on a List, should work on an array too

TomMalbran: Great. That worked.

TomMalbran:
Now the Randomizer works.

PR #1260, opened 2016-09-11
UI REDESIGN: reworked construction menu and repainted options menu

alexfriesen:
the merge conflict mess was too much so I started again #999
pp_menu


Tritaris:
I like where this is going. The 2 columns are definitely an improvement on the current `Construction` menu. 
- Would it be possible to display an image of the furniture on its build button.? 
- Can we rename `Options` to `Menu` ?  `Options` and `Settings` seems confusing. A new player might think they would be able to change some aspect of the game there. 
- `Settings ` should be placed in the `Menu` panel below `Resume`, It doesn't really make sense to have a separate button on the bottom bar for it. 
- Also can we move the renamed `Menu` to the last position where `Quests` is? As that is a button that will be used frequently we should place it we it is easy to find.

bjubes:
I wouldnt worry about renaming menus and such since that is outside the scope of this change (although if you want to you can). In your version there are some problems with the game however.
* settings dont presist between games, ie dev mode
* clicking furniture buttons doesn't allow you to build furniture
* the dev mode dialog is on top of the construction menu

I'm assuming this is due to you restarting the pr, but accidentally keeping some old code. Keep trying to comb those out, because this new ui is going to be awesome!

TomMalbran:
I am liking the new UI. I think that the white in the scrollbar should be a blue color.

We can also leave the Furniture images for a different PR. Would we need a different image for the menu, or we can use the sprite one?

It is not an issue right now but we should think how we can add categories in the construction menu. An easy way to do it would be to show first the categories covering the entire space in 2 columns and when one is selected, you move the categories and show the objects. With a back arrow to go back. Or maybe we can add a scrollabe bar of categories to the left of the menu with the categories.

Raketfart:
Looks so good, and a great ux improvement.  I see you even have the subtle glow on the edges. Yum!

I agree with all of tritaris' suggestions, but they could easily be in next PR. Probably other wishes will appear after using this in game.

alexfriesen:
I fixed the problem with building. 
I didn't want to make the same mistake again and over complicate this PR.

Next thing will be moving settings menu to the options menu.

alexfriesen:
I commented out the option to regenerate the list in devmode. 
I think this is not needed anymore and looks more like a hack. The category system mentioned by Tom will be a better solution.

TomMalbran:
A couple of Ui issues.

1. When you are in DevMode, the resources spawner window stay always on top of the construction menu, making it hard to use it. The construction menu should on top.

2. When there are Quests those are on top of the build menu. Since you can hide them it doesn't matter, but maybe the quests should be behind?

sboigelot:
There will always be issue with the UI for a while and one PR can't solve them all. This PR is a very good start for us all to start making mini PR fixing small issues.

PR #1265, opened 2016-09-11
Renamed ObjectType to Type in Furniture

TomMalbran:
We recently started changing to Type in other classes: Inventory, Stats, Needs... If we standardize on Type for all the classes that use the PrototypeManager I can simplify it by using an interface.

### DIFF
--- a/Assets/Scripts/Controllers/BuildModeController.cs
+++ b/Assets/Scripts/Controllers/BuildModeController.cs
@@ -20,7 +20,7 @@ public enum BuildMode
 public class BuildModeController
 {
     public BuildMode buildMode = BuildMode.FLOOR;
-    public string buildModeObjectType;
+    public string buildModeType;
 
     private MouseController mouseController;
     private TileType buildModeTile = TileType.Floor;
@@ -39,7 +39,7 @@ public class BuildModeController
             return true;
         }
 
-        Furniture proto = PrototypeManager.Furniture.Get(buildModeObjectType);
+        Furniture proto = PrototypeManager.Furniture.Get(buildModeType);
 
         return proto.Width == 1 && proto.Height == 1;
     }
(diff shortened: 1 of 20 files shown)


dusho:
If we would go with serialization/deserialization maybe also abstract class (that would hold at least that Type for now).. everything else would inherit from it

TomMalbran:
Sure it could work for that.

For PrototypeManager I need an Interface where the Classes require to have a Type and a ReadXml method. Then I could remove all the individual classes Prototypes classes, which are all almost exactly the same, but I had to make them since most classes used a different way to store the Type.

dusho:
ok cool.. so prototype will be gone. All variables in input class (XML now) are not changeable and dynamic things will be stored inside Parameters (cloning, load game, save game). So load game/save game will just fetch those dynamic parameters..

TomMalbran:
No. What I mean is that FurniturePrototypes, InventoryPrototypes, NeedPrototypes, TraderPrototypes, QuestPrototypes, StatPrototypes, SchedulerEventPrototypes (they are all the same except for the read part) will be replaced by a single class Prototypes where the class that is the prototype will have to implement an Interface which will require a Type and a ReadXml.

Grenkin1988:
I am glad somebody made this after `Inventory`, `Stat` and `Need`. :+1: 

Need to be tested well, could be bugs

TomMalbran:
It will need some testing. But MonoDevelop does a good job on a rename and I did an extensive search for all the ObjectType. On this PR only ones that use ObjectType are in Jobs and in the Trader.

Grenkin1988:
@TomMalbran you should do search for `objectType` also.

TomMalbran:
I did, on both a Case sensitive search and a not case sensitive one.

Grenkin1988:
Tested: build, power, production.
Spotted issue: deconstruction of `Furniture` immediate and do not need `Character`.

TomMalbran:
> Spotted issue: deconstruction of Furniture immediate and do not need Character.
That is how it works in master. I believe that #1002 fixes that.

PR #1268, opened 2016-09-11
Hide OverlayUI Tooltip

Ohda:
Few changes on the overlayUI : 
- Fix a bug with oxygen map (LUA). Only 1 color is displayed if there is some O² in the room. Color change with O² still need to be added.
- Hide the overlayUI Tooltip if no overlay is displayed.

There is a small change in the scene on the dropdown menu : position and avoide streetching it.

### DIFF
--- a/Assets/Scripts/UI/Overlay/OverlayMap.cs
+++ b/Assets/Scripts/UI/Overlay/OverlayMap.cs
@@ -256,6 +256,7 @@ public class OverlayMap : MonoBehaviour
         {
             meshRenderer.enabled = false;
             currentOverlay = name;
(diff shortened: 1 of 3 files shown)


Assets/Scripts/UI/Overlay/OverlayMap.cs
```diff
@@ -283,6 +284,7 @@ public void SetOverlay(string name)
 
             ColorMapSG = descr.colorMap;
             Bake();
+            ShowGUIToolitip();
```
sboigelot: Tool**i**tip

PR #1267, opened 2016-09-11
[WIP] Inventory refactoring part 2 and BasePrototypes indexer

Grenkin1988:
- 	Get rid of InventoryCommon
- 	BasePrototypes indexer

Next (part 3):
- I am going to add RequiredStackSize for `Furniture` in order to fix 
`local itemsDesired = { Inventory.__new("Raw Iron", 0, desiredStackSize) }`
that is not working now
- Will try to add tests

PR #1279, opened 2016-09-11
Prototypes Manager Refactoring

TomMalbran:
I did a refactor on the Prototypes Manager.

1. There is no a single `PrototypeMap` class that holds the prototypes, which is a combination of `BasePrototypes` and `XmlPrototypes`. All the other classes that extended `XmlPrototypes` where removed.
2. There is a new `IPrototypable` interface. Every class that uses prototypes needs to implement it.
3. I made a refactor on the file loading in `ModManager` which is basically the same as the change in #1259.
4. Added an indexer to `PrototypeMap` since it was added in #1267.


sboigelot:

TomMalbran:
Conflicts resolved.

koosemose:
Looks good, Merging.

PR #1282, opened 2016-09-11
Images in construction menu

sboigelot:
Use the default floor & furniture sprite in construction menu
Also add a bulldozer image (created by hand so no licence issue)

image


TomMalbran:
Cool. Now we need to get rid of the ugly buttons and have a cool background.

A couple of things, for now or later:
1. When the name of the item is too long and occupies 2 lines it is placed partially on top of the image. We need o increase the size.
2. Images that are not square look bad. Should we have a different image for the menu?
3. Should we just show the images and have the names as tooltips?

sboigelot:
@TomMalbran I mostly agree with you

> 1. When the name of the item is too long and occupies 2 lines it is placed partially on top of the image. We need o increase the size.

I do not wish to change too much the construction menu while it's still having some pending PR

> 2. Images that are not square look bad. Should we have a different image for the menu?

We should change the stretch mode of the image not to fill the full square, but I have no idea how.

> 3. Should we just show the images and have the names as tooltips?

I would disagree here, the images are not representative enough. For instance, take the steal wall :/

vogonistic:
This is a great idea! In a future PR I think we need to add support for a custom image for things like doors, floors and walls.

bjubes:
Getting some errors with the new master:
```
Assets/Scripts/UI/InGameUI/MenuLeft/ConstructionMenu.cs(188,38): error CS0117: `SpriteManager' does not contain a definition for `current'
Assets/Scripts/UI/InGameUI/MenuLeft/ConstructionMenu.cs(154,42): error CS0117: `SpriteManager' does not contain a definition for `current'
```

sboigelot:
@bjubes I'll check that now

sboigelot:
@bjubes I updated the code to the new SpriteManager API

Issue #1196, opened 2016-09-07
Should we use the Scheduler for every game related tick update?


sboigelot:
@mikejbrown and I just had a discussion in #1194 about "when should we use the scheduler or not"

At this point, we have some tick update still using the Time.DeltaTime or a delayed accumulated Time.DeltaTime to trigger update method.

Simple example are the Temperature or the QuestController.

By using the scheduler we would open the doors to moders'modifications.

What are your opinion about it?

bjubes:
I was under the impression the scheduler was for "long term tasks" that take at least a few seconds. if something needs to run every frame it shouldn't use the scheduler. The main point of the scheduler is to prevent this type of thing:
```
float timer = 0;
public float UpdateTime = 3f;

Update() {
    timer += time.DeltaTime;
    if (timer >= UpdateTime){
        //do a task
        timer = 0;
    }
}
```

Ticks are different. passing deltaTime around to use it _constantly_ seems fine to me.

mikejbrown:
I pretty much said my piece on the PR, reproduced here:

> I would think that's a perfect use case for the scheduler. You have an event ("update the temperature") which is supposed to happen at regular intervals based on the game clock. It's already being used for autosaves, for instance. It's also open to lua so modders can eventually go wild with it. So far no one is relying on the scheduler queue containing only a limited kind of gameplay related events, nor should they, I think.

As principal author of the scheduler system so far though I'm going to step back a bit here and let people tell me how they want to use it. :smile: 

PS: Keep in mind that in the future it is possible to implement multiple concurrent scheduler queues or a filterable queue so that events could be tagged with a type and you could easily look up all events having to do with, e.g., traders or temperature updates or whatever.

PPS: Note that the scheduler `Update` _already_ runs every tick. At the moment this incurs an O(n) cost (n being the number of events in the queue) per tick, though this could probably be optimized fairly aggressively if the queues get too long.

sboigelot:
@bjubes This type of things is done in the temperature and quest, But only to reduce the number of call to the actual update and not overload the system. This is fine for me, but @mikejbrown makes some good points (about modding)

mikejbrown:
And to clarify I'm not talking about using it for "every gameplay related tick update" really. I'm talking about things that are meant to happen at regular intervals by game clock (or once off with a delay).

bjubes:
[...]

PR #1284, opened 2016-09-12
Time manager events

TomMalbran:
This is an idea proposed in #1196. Instead of having to go to WorldController and add a line to call the Update method, the TimeManager now has several events that any module can listen to, without adding extra code in the WorldManager.

Eventually we can even add Registers methods for Lua functions. But for now is just for C# code.

Since I moved some Time related code to the TimeManager, I changed a few things there.


TomMalbran:
There is an error in a test. Will fix it soon.

Assets/Scripts/Models/World.cs
```diff
@@ -159,6 +162,12 @@ public void DeleteRoom(Room r)
         r.ReturnTilesToOutsideRoom();
     }
 
+    public void AddEventListeners()
+    {
+        TimeManager.Instance.EveryFrameUnpaused += TickEveryFrame;
+        TimeManager.Instance.FixedFrequencyUnpaused += TickFixedFrequency;
```
sboigelot: You should probably un-register this when the world get GC, we plan to have N worlds in a game session.

TomMalbran: Good point. Will do that.

TomMalbran: And done

Issue #1276, opened 2016-09-11
left/right arrow keys are flipped


Sommerbo:
As title says left and right input is flipped

rafaeu749:
I did a small fix!
Swapping two lines: 5s...
Committing changes to Github: 3h...

PR #1285, opened 2016-09-12
Fixed swapped arrow keys

rafaeu749:
Quick fix for #1276
(Trying for the third time, ty git!)


vogonistic:
A better title would be "Fixed swapped arrow keys" and then in the message body, you put in "Fixes #1276"

When the people who get emails about new PR's see this, there is nothing in the email that explains what it is about.

PR #1289, opened 2016-09-12
Redo needs

koosemose:
* Changes Needs to no longer have hardcoded decay specific to Oxygen.
* Needs fires EventActions for updates (for specific decay/regular gain implementation for a need), and for various severity levels of the need. If no OnUpdate EventAction is defined it will use a default Update to decay needs. OnEmptyNeed will also have default, but it currently does nothing as there is not yet good supporting framework, particularly for Oxygen, to have negative actions on an empty need.
* Needs still run from 0 (good) to 100 (bad - empty), as code outside of Need relies on that ordering, and I wanted to minimize files touched outside of Need, until the game could better support Need directly handling what a character does when a need needs fulfilled.

This includes changes to EventActions:
EventActions no longer specifically calls Furniture LUA functions, allows any target Type (that has Lua Functions).

### DIFF
--- a/Assets/Scripts/Models/EventAction.cs
+++ b/Assets/Scripts/Models/EventAction.cs
@@ -123,7 +123,7 @@ public class EventActions : IXmlSerializable
     /// <param name="actionName">Name of the action being triggered.</param>
     /// <param name="target">Object, passed to LUA function as 1-argument (TODO: make it an object).</param>
     /// <param name="deltaTime">Time since last Trigger of this event.</param>
-    public void Trigger(string actionName, Furniture target, float deltaTime = 0f)
+    public void Trigger<T>(string actionName, T target, float deltaTime = 0f)
     {
         if (!actionsList.ContainsKey(actionName) || actionsList[actionName] == null)
         {
@@ -131,7 +131,7 @@ public class EventActions : IXmlSerializable
         }
         else
         {
-            FunctionsManager.Furniture.CallWithInstance(actionsList[actionName].ToArray(), target, deltaTime);
+            FunctionsManager.Get(target.GetType().ToString()).CallWithInstance(actionsList[actionName].ToArray(), target, deltaTime);
         }
     }
 }
--- a/Assets/Scripts/Models/EventAction.cs
+++ b/Assets/Scripts/Models/EventAction.cs
@@ -134,4 +134,14 @@ public class EventActions : IXmlSerializable
             FunctionsManager.Get(target.GetType().ToString()).CallWithInstance(actionsList[actionName].ToArray(), target, deltaTime);
         }
     }
+
+    /// <summary>
+    /// Determines whether this instance has any events named actionName.
+    /// </summary>
+    /// <returns><c>true</c> if this instance has any events named actionName; otherwise, <c>false</c>.</returns>
+    /// <param name="actionName">Action name.</param>
+    public bool HasEvent(string actionName)
+    {
+        return actionsList.ContainsKey(actionName);
+    }
 }
--- a/Assets/Scripts/Models/EventAction.cs
+++ b/Assets/Scripts/Models/EventAction.cs
@@ -121,7 +121,7 @@ public class EventActions : IXmlSerializable
     /// Fire the event named actionName, resulting in all lua functions being called.
     /// </summary>
     /// <param name="actionName">Name of the action being triggered.</param>
-    /// <param name="target">Object, passed to LUA function as 1-argument (TODO: make it an object).</param>
+    /// <param name="target">Object, passed to LUA function as 1-argument.</param>
     /// <param name="deltaTime">Time since last Trigger of this event.</param>
     public void Trigger<T>(string actionName, T target, float deltaTime = 0f)
     {
--- a/Assets/Scripts/Models/Need.cs
+++ b/Assets/Scripts/Models/Need.cs
@@ -122,21 +122,21 @@ public class Need : IPrototypable
         } 
         else if (Amount > 90f)
         {
-            if (EventActions != null && EventActions.HasEvent("OnSevereNeed"))
+            if (EventActions != null)
             {
                 EventActions.Trigger("OnSevereNeed", this, deltaTime);
             }
         }
         else if (Amount > 75f)
         {
-            if (EventActions != null && EventActions.HasEvent("OnCriticalNeed"))
+            if (EventActions != null)
             {
                 EventActions.Trigger("OnCriticalNeed", this, deltaTime);
             }
         }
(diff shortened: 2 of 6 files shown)


Assets/Scripts/Models/Need.cs
```diff
@@ -83,25 +89,57 @@ public string DisplayAmount
...
+        else 
         {
-            Amount += (addedInVacuum - (addedInVacuum * (character.CurrTile.GetGasPressure("O2") * 5))) * deltaTime;
+            DefaultNeedDecay();
         }
 
         if (Amount.AreEqual(100))
         {
-            // FIXME: Insert need fail damage code here.
+            if (EventActions != null && EventActions.HasEvent("OnEmptyNeed"))
```
sboigelot: Why are these method static string name? Shoudn't we have a property storing the function name in the needs for each to fit the same pattern as the rest of the game?

sboigelot: Forget this, I didn't see the new event system :) my bad.

sboigelot:
The need for sleep is dropping like a stone in this pr, in less then 5 sec from 100 to 0. Is this intended?
This is not the case in master :)

koosemose:
Hmm it shouldn't be behaving differently... but I'll check what it was doing before exactly

koosemose:
Left deltaTime out of default need decay

Assets/Scripts/Models/Need.cs
```diff
@@ -33,18 +30,21 @@ private Need(Need other)
...
-        restoreNeedAmount = other.restoreNeedAmount;
-        CompleteOnFail = other.CompleteOnFail;
-        addedInVacuum = other.addedInVacuum;
-        dps = other.dps;
-        luaUpdate = other.luaUpdate;
-        luaOnly = other.luaOnly;
+        RestoreNeedAmount = other.RestoreNeedAmount;
+        Damage = other.Damage;
+
+        if (other.EventActions != null)
```
TomMalbran: Can we assume that we always have an instance of EventActions?

In Furniture sometimes we assume is there and sometimes we dont... is weird like that.

Assets/Scripts/Models/Need.cs
```diff
@@ -83,25 +89,57 @@ public string DisplayAmount
...
+                EventActions.Trigger("OnEmptyNeed", this, deltaTime);
+            }
+            else
+            {
+                DefaultEmptyNeed();
+            }
+        } 
+        else if (Amount > 90f)
+        {
+            if (EventActions != null && EventActions.HasEvent("OnSevereNeed"))
```
TomMalbran: Trigger does nothing if it doesn't have that action. Maybe we don't even need to check if it exists. And if we assume that EventActions is always present, we could remove the extra ifs.

Assets/StreamingAssets/LUA/Need.lua
```diff
@@ -5,6 +5,15 @@
...
 ---------------------------- Need Actions --------------------------------
 
+function OnUpdate_Oxygen( need, deltaTime )
+    if (need.Character != nil and need.Character.CurrTile.GetGasPressure("O2") < 0.15) then
+        need.Amount = need.Amount + ((0.3 - (0.3 * (need.Character.CurrTile.GetGasPressure("O2") * 5))) * deltaTime)
+    else
+        need.Amount = need.Amount - (need.GrowthRate * deltaTime)
+    end
+    
+    return
```
TomMalbran: we dont need this return

Assets/Scripts/Models/Need.cs
koosemose: This being a project involving lots of people, I'd prefer to go the defensive coding route, to ensure things don't crash unnecessarily if someone later does something unexpected that leads to it being called before EventActions exists, or some other thing I'm not thinking about ahead of time.

Assets/Scripts/Models/Need.cs
koosemose: You're probably right, for now as far as checking if the event exists beforehand, though in the future I'd like to come up with default things, but I don't have a clear idea what all of the needs are going to want to do, and some of the what to do on needs is still tied up in Character, but I don't know if there's a good way with the current job system to have jobs to do for a specific character but that don't have to be done immediately ( such as you might want with moderate need ).

But for now I could probably remove the HasEvent Checks... though I wonder what people's opinions are on having Trigger() return a bool depending on if it performed any actions or not (so rather than checking first if one exists and then triggering if it does, and doing something else if it doesn't, we could just trigger and if it doesn't do anything we could then do something else.)

Assets/Scripts/Models/EventAction.cs
```diff
@@ -123,15 +123,25 @@ public void Deregister(string actionName, string luaFunc)
     /// <param name="actionName">Name of the action being triggered.</param>
     /// <param name="target">Object, passed to LUA function as 1-argument (TODO: make it an object).</param>
     /// <param name="deltaTime">Time since last Trigger of this event.</param>
-    public void Trigger(string actionName, Furniture target, float deltaTime = 0f)
+    public void Trigger<T>(string actionName, T target, float deltaTime = 0f)
```
TomMalbran: An object might work too?

Assets/Scripts/Models/EventAction.cs
```diff
@@ -123,15 +123,25 @@ public void Deregister(string actionName, string luaFunc)
     /// <param name="actionName">Name of the action being triggered.</param>
     /// <param name="target">Object, passed to LUA function as 1-argument (TODO: make it an object).</param>
```
TomMalbran: TODO is done now :)

Assets/Scripts/Models/EventAction.cs
koosemose: It might, but this works at least as well. When the T is one of the arguments passed, you don't even need to include the brackets (so it would still be called as `Trigger("OnUpdate",need)`).

Assets/Scripts/Models/Need.cs
TomMalbran: Sure. We don't have many needs yet to make a proper decision about Defaults. Maybe we won't even need them.

Since we will keep checking for the nullability of EventActions, I am ok with leaving the the other part.

I am not sure about a return value for trigger, since then it might not be clear if the return is because it run, or was the result of the action. It might need a new name if it does that.

Assets/Scripts/Models/EventAction.cs
koosemose: Also, that's a terrible place to have put a TODO, I didn't even see it... it took me a bit to even spot it after you mentioned it directly.

PR #1294, opened 2016-09-12
Fix OnJobComplete LUA

### LINKED ISSUE #1292: [Bug] OnJobComplete LUA
There is a bug actually on master with LUA.

When thoses functions are called : CloningPod_JobComplete, PowerGenerator_JobComplete, MiningDroneStation_JobComplete, an error occured because  job.furniture is null at this time.

TomMalbran:
Fixes #1292.

In PR #1084 this [line](https://github.com/TeamPorcupine/ProjectPorcupine/pull/1084/files#diff-dcf105114d4aaa546cf063e0bc311e20R817) was added. It make sense to add it, but it broke the the Lua callbacks.

Adding that line means that when a job was completed, the ```OnJobCompleted``` event was triggered which called ```OnJobStopped``` in furniture, which removed the job, and when removing the job it removes the furniture association.

By calling the Lua functions first we avoid that issue.


sboigelot:
I still get an exception with this fix (same as master)

image

```
chunk_1:(326,1-157): attempt to index a nil value
UnityEngine.Debug:LogError(Object)
UberLogger.Logger:ForwardToUnity(Object, LogSeverity, Object, Object[]) (at Assets/UberLogger/UberLogger.cs:506)
Debug:ULogErrorChannel(String, String, Object[]) (at Assets/UberLogger/Debug.cs:322)
LuaFunctions:Call(String, Object[]) (at Assets/Scripts/Models/Functions/LuaFunctions.cs:89)
Job:DoWork(Single) (at Assets/Scripts/Models/Job.cs:259)
Character:Update_DoJob(Single) (at Assets/Scripts/Models/Character.cs:732)
Character:Update(Single) (at Assets/Scripts/Models/Character.cs:346)
World:TickEveryFrame(Single) (at Assets/Scripts/Models/World.cs:167)
WorldController:Update() (at Assets/Scripts/Controllers/WorldController.cs:165)

```

TomMalbran:
This issue got fixed in master from another of my PR where I might have failed on a merge. Anyway it was what I was going to do to fix it here, remove the line which I mention in the first comment.

I think that the change made here might still make sense, but it dos nothing now.
I went ahead and added a meta file I forgot.

sboigelot:
Works great

sboigelot:
All test runs, travis is lagging behind but will succeed -> merge

PR #1280, opened 2016-09-11
StyleCop + add missing meta

sboigelot:

PR #1295, opened 2016-09-12
Remove unwanted metas

TomMalbran:
PR #1280 added unwanted meta files. This PR removes them.

### DIFF
--- a/Assets/Scripts/Models/Inventory.meta
+++ b/Assets/Scripts/Models/Inventory.meta
@@ -1,9 +1,0 @@
-fileFormatVersion: 2
-guid: bcfa1a8fa66fb3e4a91b95020d7d5822
-folderAsset: yes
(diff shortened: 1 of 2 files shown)


dusho:
what is the general deal with .meta files? I saw them being committed, but I tend to remove them when I can..
Should all .meta files be added into .gitignore or are some actually needed?

TomMalbran:
Unity creates a meta file for each folder and each file, which contain a GUID, and that is then used to associate a script with a prefab or a scene. When the file is a standalone class, I think is not important, but when it is associated with a prefab or with the scene, then meta file is essential, since everyone needs to have the same GUID for Unity to work fine.

So we can't really ignore all of them, and when some are missing and/or there are extras, it makes it harder to commit and ignore them.

I was just reading about this since I was curious. So I might wrong in some details.

PR #1299, opened 2016-09-13
Implement utilities (Phase 1)

koosemose:
Utilities are a separate layer from furniture, multiple utilities (of different types) can exist in a single tile.

Presently they show at 25% opacity. This is just the implementation of the utilities system itself, there is not presently any actual functionality, that will be implemented later (possibly individually by utility).

There are currently two utilities, for demonstration purposes (water was added simply to demonstrate multiple utilities in a single tile).

Phase 1: Basic Implementation, allowing creation of utilities on a layer separate from furniture, allow multiple utilities per tile.
Phase 2: Control placement position depending on number of other utilities already in the tile, ensure neighboring utilities match up same utilities in the same relative positions.
Phase 3: Fully functional Utilities, that determine what grids a connected furniture is on.

Phase 2 and 3 may be reversed if my guess as to the time it will take to implement each is incorrect, and further refinements may be made as the system gets used more and other use cases are found.

Separate from direct involvement in the utility implementation phases, I will be investigating refactoring IBuildable into a Buildable class from which Furniture and Utility will inherit, and refactoring as appropriate.


sboigelot:

koosemose:
Conflicts Resolved

Assets/Scripts/Models/Prototypes/UtilityPrototypes.cs
```diff
@@ -0,0 +1,38 @@
+﻿//#region License
```
TomMalbran: You can delete this file. It shouldn't be needed.

Assets/Scripts/Models/Utility.cs
```diff
@@ -0,0 +1,753 @@
...
+        int x = Tile.X;
+        int y = Tile.Y;
+        if (Tile.Utilities != null)
+        {
+            utility.CancelJobs();
+        }
+
+        // We call lua to decostruct
+        // TODO: EventActions doesn't currently allow anything but Utility Fix this when implemented
+//        EventActions.Trigger("OnUninstall", this);
```
TomMalbran: This should work now :)

Assets/StreamingAssets/Data/Utility.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
```
TomMalbran: There are 2 xml files?

koosemose: Must not have committed the deletion of one when I changed the name...

Assets/Scripts/UI/UtilityBuildMenu.cs
```diff
@@ -0,0 +1,101 @@
+#region License
```
TomMalbran: Are we still using this Build files? I believe that the ConstructionMenu replaced them. We still have the Furniture one, which might need to be removed too.

sboigelot:
Can you move the water sprite 1 pixel up and 1 pixel to the left? to avoid supperposition
That would be much more clear and easy to do
image

koosemose:
Water Pipes were only ever there to make it clear that multiple things can fit into a single tile. And while not difficult, it's an excessively tedious process to move everything over, which isn't worth it for a placeholder image.

I've removed the water pipes for now so that an art issue isn't what's holding up the PR.

sboigelot:
@koosemose sorry it look complicate for you, I think we should keep the water pipes as it proof we can have multiple utilities on a tile.

My request for moving around the water pipe was not only cosmetic but also to provide guidelines to the art team.

I made the shift myself, if you wish to add it back, here is the file : 

water_pipes

koosemose:
Apparently then I need to change what I'm using (or not be terrible with what I'm using), as I got perhaps a quarter of the way through the sheet in about an hour and a half. 

The other reason I was reluctant to do the shift is that I intend to have it shift on the fly, through code, so that pipes don't have to have their "slot" designated solely by the art. However, I also feel that the coding needed for this is going to be complex enough to warrant it's own PR. Going this route means that we could implement, for example, all pipes carrying a fluid as one Utility, and you could still run multiple through a single tile. I'd rather not have our artists working on offset sprites when they may not need to be, and may in fact need to be shifted back to center to work properly. 

I can certainly go with the offset sprites, it's much simpler to implement (code side anyways), but limits the available expansion space of utilities, as every different type of utility will have to have it's own predesignated slot, and we would have to either have separate utilities for every type of fluid (for example) that could go through a pipe, or be limited to only a single fluid pipe in a tile.

sboigelot:
@koosemose yes, just 

- edit image with any paint (even the windows one, or gimp if you are on !windows)
- ctrl-a
- ctrl-x
- ctrl-v
- press arrow up a bunch
- press arrow left a bunch
- ctrl s
- magic

About the pipe position, I agree with you, it would be better later, on a different PR, to have them automatically assign an offset per tile. But that can wait.

Moving sprite around in a picture being simple enough though, I think we should just have the hard-drawn offset in the sprite for now.

koosemose:
I had a lot more issues dealing with the overhangs from shifting things... but as I mentioned, I'm pretty terrible with gimp or anything else really...

sboigelot:
no problem :)

programmer art

Small stylecop issues and we are good to merge:

`Warning	27	SA1507 : CSharp.Layout : The code must not contain multiple blank lines in a row.	D:\git\ProjectPorcupine\Assets\Scripts\Models\Utility.cs	667	1	ProjectPorcupine.CSharp
Warning	28	SA1506 : CSharp.Layout : A documentation header line must not be followed by a blank line.	D:\git\ProjectPorcupine\Assets\Scripts\Models\Utility.cs	668	1	ProjectPorcupine.CSharp
`

koosemose:
Stylecopped.

sboigelot:
@koosemose I think you forgot to commit or push your stylecopping (or my github lag)

koosemose:
It's definitely been there the whoooole time ***innocent whistle***

koosemose:
Though it's still not showing the commit in the conversation like it usually does... but it's definitely showing on the commit tab now

Issue #1004, opened 2016-08-28
Floor tile idea- Keep track of how often it is stepped on


kd7uiy:
I've stated I don't like the floor tile. One thought I just had was to keep track of how often a tile is walked on, and change accordingly. If a tile is heavily used, it will start to look more worn after a while. It seems easy enough to do, and could be a cool effect. Thoughts?

Sommerbo:
I like it. I am a fan of dirty/used space look.

Fatalexceptionz:
Space janitors!

kd7uiy:
That could be an interesting idea... Clean the dirty floor periodically, otherwise, not sure what happens, bad stuff?

bjubes:
Maybe very dirty floors slow movement? Or do a rimworld beauty mechanic? Either way keeping track is a good idea

vogonistic:
Not sure about slowing movement, but it might increase the rate that they get sick, failure rate on things that require concentration (sneezes and stuff) and happiness.

kd7uiy:
Increased chance of tripping?

koosemose:
Rather than tracking how much a given piece of floor is stepped on, how about every time a floor is walked on there is a small percentage chance that it gains a "dirtiness level", less data to track, it would be less deterministic, and it would be relatively easy to have other activities have a higher likelihood of making it dirty (so walking is 1% chance for example, running is 5% chance. Also these are completely arbitrary numbers and examples for illustrative purposes). It would also be possible to have different characters just be generally messier (character with a slob trait for example may double the chances of dirtying the floor).

kd7uiy:
I like the random tracking idea, that makes sense as well.

kd7uiy:
Okay, so let's do this. I believe this would be a good beginner project to do. For the time being, let's have the character call a function each time a Tile is stepped on. That function for now will add 1 to a float value that keeps track of how much a tile has been walked on. We can then enhance this further, as desired, to include random wear and tear, character based wear and tear, etc.

Tranberry:
How do I vote against this? only blood should tarnish the floors.. *I'm trying to make a joke please don't be offended*

vogonistic:
If we are talking about dust, walking on it would probably clean it, or move it to a neighboring tile.

kd7uiy:
That's only for fairly heavy dust. But I'm talking more of the standard wear and tear that tends to happen.

vogonistic:
[...]

PR #1300, opened 2016-09-13
Keeps track of how often a step has been made.

kd7uiy:
There are also changes for frequently walked on tiles. The tiles right now are "Programmer Art". The naming schema, values, etc can be tweaked as required, but this is at least a starting point for future changes.
capture


kd7uiy:
Fixes #1004

kd7uiy:
capture

Updated the tiles slightly based on suggestion.

vogonistic:
Which tile denotes what state? Is this an implementation in the direction of tiles needed repairs?

Assets/Scripts/Controllers/SoundController.cs
```diff
@@ -56,8 +56,11 @@ private void OnTileChanged(Tile tileData)
             return;
         }
 
-        AudioClip ac = Resources.Load<AudioClip>("Sounds/Floor_OnCreated");
-        AudioSource.PlayClipAtPoint(ac, Camera.main.transform.position);
-        soundCooldown = 0.1f;
+        if (tileData.walkCount == 0 && tileData.damageEventCount == 0)
```
vogonistic: Why is the sound only played on new tiles?

Assets/Scripts/Controllers/TileSpriteController.cs
```diff
@@ -56,11 +56,16 @@ protected override void OnCreated(Tile tile)
...
+    }
+
+    protected override void OnChanged(Tile obj)
+    {
+        OnChanged(obj, false);
     }
 
     // This function should be called automatically whenever a tile's data gets changed.
-    protected override void OnChanged(Tile tile)
+    protected void OnChanged(Tile tile, bool forceTile)
```
vogonistic: As far as I can tell, forceTile is never used and I can't figure out what it means.

Assets/Scripts/Controllers/TileSpriteController.cs
```diff
@@ -75,8 +80,20 @@ protected override void OnChanged(Tile tile)
             Debug.ULogErrorChannel("TileSpriteController", "tileGameObjectMap's returned GameObject is null -- did you forget to add the tile to the dictionary? Or maybe forget to unregister a callback?");
             return;
         }
-        
-        tile_go.GetComponent<SpriteRenderer>().sprite = SpriteManager.current.GetSprite("Tile", tile.Type.Name);
+
+        //TODO Evaluate this criteria and naming schema!
```
vogonistic: Can we evaluate it before we merge the PR? Doesn't sound like a discussion that would be controversial.

Assets/Scripts/Models/Tile.cs
```diff
@@ -327,61 +367,51 @@ public Enterability IsEnterable()
...
+        walkCount++;
+        ReportTileChanged();
     }
 
-    public Tile South()
+    public void OnTileClean()
     {
-        return World.Current.GetTileAt(X, Y - 1, Z);
+        walkCount = 0;
+        damageEventCount++; //TODO Probably not the best place for this, but...
```
vogonistic: Cleaning and damage seems like two different features.

vogonistic:
You need to describe this PR more. You talked about tracking steps only but there is also some uncalled code for cleaning the tiles and accumulating damage.

If this is the first PR of several, then you should create an issue that shows the progression after this one.

Assets/Scripts/Controllers/SoundController.cs
kd7uiy: We could add an alternative sound for walking on tiles, but it didn't seem like we should be playing the floor tile created sound every time a tile is walked on.

Assets/Scripts/Controllers/TileSpriteController.cs
kd7uiy: The force tile is used for when first creating a tile to ensure it is in the proper state. I will agree that it doesn't do anything at the moment. The tile sprite is actually changed every time it is walked on at the moment, I had in mind a "force" state to only change if the conditions changed, or if the tile was created. I can continue with those improvements prior to implementation if you think it best.

kd7uiy:
Okay, I removed the Damaged tile counter, which was being badly abused. I do have a "TypeChanged" flag which will be set to false after any update. The first time before the update is sent the flag is set to true. 

The cleaning code is currently used in 1 area, namely to make sure the tile is "clean" when it is first created. If an empty space was walked on, then those walks would actually count, making this potentially occur more often than required.

Assets/Scripts/Controllers/TileSpriteController.cs
kd7uiy: The truth is, I'm not really sure how this should look until it's been in the hands of a few users. Is 100 times entering a tile so rare it will never happen, or happen all the time? I really don't know. We could discuss this further if desired, I'm just not sure if that would help significantly.

dusho:
don't get those 2 tiles.. one is shiny new and other is more used or something?
that shaded one looks out of place comparing to other existing art.. especially furniture looks like 'hidden' in there

kd7uiy:
The shiny is new, the duller one is used. I'm certainly willing to change the artwork, I did call it "Programmer Art", but I would suggest that be in a separate PR.

Ohda:
For the temporary titles i suggest : keep original for clean floor and to a darker (just a bit) version for dirty.

kd7uiy:
I will admit, I've never liked the current version of the tile and a small part of me wanted to change those while I was at it... Still, I'll go ahead and make that change.
capture

Assets/Scripts/Controllers/TileSpriteController.cs
vogonistic: Changing it if we don't like it is the default if you don't put in the comment. You might even get better suggestions on it if you put it out to people rather than a TODO comment.

By putting it in the code you are requiring someone to make a PR to either remove the comment if we agree on it or to remove it and change the names.

Assets/Scripts/Models/Tile.cs
```diff
@@ -132,6 +130,7 @@ public float MovementCost
     }
 
     public bool IsSelected { get; set; }
+    public bool TypeChanged { get; protected set; }
```
vogonistic: What is `TypeChanged`?

Assets/Scripts/Controllers/TileSpriteController.cs
```diff
@@ -84,11 +80,17 @@ protected void OnChanged(Tile tile, bool forceTile)
         //TODO Evaluate this criteria and naming schema!
         if (DoesTileSpriteExist(tile.Type.Name + "_Heavy") &&(tile.walkCount>=100))
         {
-            ChangeTileSprite(tile_go, tile.Type.Name + "_Heavy");
+            if (tile.TypeChanged || tile.walkCount == 100)
+            {
+                ChangeTileSprite(tile_go, tile.Type.Name + "_Heavy");
```
vogonistic: The sprite is called `_High`

vogonistic:
You need to either:
1. Document and/or rename `TypeChanged` and move the wrap all the if-statements that checks if the sprite should change inside a `tile.TypeChanged` check. Should probably check the count first and images after.
2. Track the current sprite with an enum or other method so that you don't have to keep a variable that is used to see if you have to set the sprite on the next `OnChanged` call.

vogonistic:
I'm torn on this PR. I like walking potentially leading to damage and repairs, but I don't understand where the dirt is coming from.

There is also a potential pathfinding optimization that can use this, where we make a previously walked trail slightly cheaper than than less walked tiles.

Assets/Scripts/Controllers/TileSpriteController.cs
```diff
@@ -90,4 +109,15 @@ protected override void OnChanged(Tile tile)
...
     {
     }
+
+    private void ChangeTileSprite(GameObject tile_go, string name)
+    {
+        // TODO How to manage if not all of the names are present?
+        tile_go.GetComponent<SpriteRenderer>().sprite = SpriteManager.current.GetSprite("Tile", name);
+    }
+
+    private bool DoesTileSpriteExist(string name)
```
vogonistic: Is the idea that in the future some floor types might not have all 3 sprites?

kd7uiy: Yes, and in fact, the only tile type that does is the Floor, so... Also, this allows for loading it to the correct sprite, without changing the sprite every time.

sboigelot:

kd7uiy:
How did that happen? Actually, I see how... Will have this fixed ASAP!

Assets/Scripts/Controllers/TileSpriteController.cs
```diff
@@ -75,8 +75,27 @@ protected override void OnChanged(Tile tile)
...
+
+        // TODO Evaluate this criteria and naming schema!
+        if (DoesTileSpriteExist(tile.Type.Name + "_Heavy") && (tile.walkCount >= 30))
+        {
+            if (tile.ForceTileUpdate || tile.walkCount == 30)
+            {
+                ChangeTileSprite(tile_go, tile.Type.Name + "_Heavy");
+            }
+        }
+        else if (DoesTileSpriteExist(tile.Type.Name + "_Low") && (tile.walkCount >= 10))
```
sboigelot: I'm not a fan of the 2 state only tile visualisation
What about using something like the fancy new: #1296 ?
Or something more flexible, defined in the tile type? for instance an array of threshold/sprite mapping

kd7uiy: It's actually a 3 state, but I'll look in to that particular PR. I will admit that I was mostly concentrating on the Tile updates, and less on the actual visual impacts.

sboigelot:
@vogonistic 

> I'm torn on this PR. I like walking potentially leading to damage and repairs, but I don't understand where the dirt is coming from.

What about space dust from asteroid mining? Would that make sense?

vogonistic:
@sboigelot Dust from asteroid mining would be small particles that are dry and gets pushed to the corners and into sensitive machines that then start malfunctioning. It doesn't behave like dirt and mud that gets tracked in by people. Moving over a tile with space dust should disturb it and some of it should move to other neighboring tiles, so moving would result in the tile having less dust than before.

sboigelot:
@vogonistic But that's assuming the asteroids do not contain any liquids. And we know that's false as we already plan to have ice mining.

vogonistic:
@sboigelot I might have missed something, so correct me if I'm wrong, but asteroids contains so little water that it could be nothing for all practical purposes while comets are mostly ice.

If we decide that our asteroids are mixed material I guess you could say that you will drag in a mixture of ice and dirt when returning from outside and not taking the suite off, but I still feel weird about tracking in dirt this way and would prefer it just being wear and tear on the tile.

When it comes to suspending reality, I'm good with veins of metal in asteroids even though there isn't anything pointing to metal existing in that way in asteroids, so maybe mixing in veins of ice makes sense too.

sboigelot:
@vogonistic I'm really not a space expert so I may be completly wrong

> C-type asteroids have a high abundance of water which is not currently of use for mining but could be used in an exploration effort beyond the asteroid. Mission costs could be reduced by using the available water from the asteroid. C-type asteroids also have a lot of organic carbon, phosphorus, and other key ingredients for fertilizer which could be used to grow food.[19]
S-type asteroids carry little water but look more attractive because they contain numerous metals including: nickel, cobalt and more valuable metals such as gold, platinum and rhodium. A small 10-meter S-type asteroid contains about 650,000 kg (1,433,000 lb) of metal with 50 kg (110 lb) in the form of rare metals like platinum and gold.[19]
M-type asteroids are rare but contain up to 10 times more metal than S-types[19][19]

from https://en.wikipedia.org/wiki/Asteroid_mining

vogonistic:
@sboigelot I'm not going to argue with wikipedia

kd7uiy:
Dust certainly would come from the planetary environment. I strongly suspect there'd be a lot of dust that would get in from an asteroid environment.  The orbital environment is a bit trickier, but even the ISS has some dirt, it comes from having humans living there. 

The moon contains virtually no water, but there was considerable dust dragged in by the astronauts.

It's kind of hard to say what would really happen actually, because we are building something that doesn't really exist, a base that presumably has gravity (I don't think we've actually discussed this) that mines asteroids. Still, I would be willing to be there will be dust tracked in from the outside, if nothing else what will happen is what happens in homes, the high traffic areas will be somewhat darker.

Assets/Scripts/Controllers/TileSpriteController.cs
Raketfart: That's just for furniture. I don't know if some of it can be reused for tiles, but probably not.

vogonistic:
I'm not arguing that there isn't dust in the environment, just that I think it would behave like the dry red dust we have here where I live in the Sierra Nevada. It tends to be moved out of the high traffic areas by people walking, except for right by the door. If it's the rainy season, it makes the floors dark everywhere someone has walked.

I'm going to stop arguing now. I'm not a fan, but I don't have to be if enough other people like it.

koosemose:
I think I'm in the same camp as vog on this one, fine with wear and tear on tiles, but less so the whole dirt thing. 

Though I think with either there's the question of if they actually add any gameplay value, every other version of it I've seen it's pretty much just a time sink.

Raketfart:
I think it would take a lot of walking to scratch a metal base floor. 
I love the idea, but maybe this is more suited for outside tiles. Like tiles on planets/asteroids would eventually get pathways where characters walked a lot from and to mining areas. I think that would add a lot of flavor.
I vote for using this, and then adjusting the specifics, when we have more than one type of tile.

kd7uiy:
I actually think there will be a lot of dirt in a station, how that will affect things is difficult. I do agree this isn't a perfect solution, but I'm not sure how much we can really figure out how well that perfect solution will be until we have some kind of a system in place and are able to play with it. As it stands now, I personally think it adds a bit of flavor, and sets in motion at least research, and more likely a framework, for future more interesting changes that could be done.

sboigelot:
Yep same concern as @Raketfart on my side. I don't understand how a floor tile can get damage out of normal usage during the length of a game. It would take years of daily walking to damage a floor.

vogonistic:
@Raketfart has a good point, let's move forward with this and evolve it as needed.

jaselito:
I might be totally off base here, but wouldn't a system like this also allow for optimizing pathfinding for certain cases in the long run?  I mean, if there's a path most people are taking it's for a reason...  For instance, if you need to cross a room from door to door, you start by looking at the most trodden tiles. Or something like that.

On the whole dust vs. worn debate, I'd be on the worn side :P

kd7uiy:
@jaselito There are more optimal ways than using this. One such way would be to pre-compute the path from each doorway to each other doorway in the same room, for instance, and use those as a shortcut for pathfinding calculations.

jaselito:
Yeah, you're right.  Doors was a bad example.  I'm thinking more in a general sense; if in general after some time there arise some clearly defined pathways, surely the pathfinding algorithm should take that into consideration?  I'm talking as a layman here so feel free to smack me down :P

koosemose:
You got an assortment of Stylecop issues, looks like all but the 2 in DialogBoxSettings are yours.

koosemose:
Looks good, Stylecop clean. Merged

PR #1301, opened 2016-09-13
Delete No Longer Required meta files.

kd7uiy:
These files should be deleted.


svmnotn:
this pr and #1295 are exactly the same

PR #1302, opened 2016-09-13
UI Bottom menu images

Ohda:
I have just replaced the images of the buttom menu with @alexfriesen's ones and switching the font color to white.


TomMalbran:
This looks nice for now. But I think that later it will require a new design. @alexfriesen had one in the ui design post.

sboigelot:
@TomMalbran agree, but this is ready. I'll always prioritize a ready PR to a possible future PR. There is too much PR being dropped by the original authors.

TomMalbran:
I don't disagree with the merge. The UI looks much better. Just pointing out that this will not be the final design.

sboigelot:
Sure, nothing is final :) never ^^

Tranberry:
" I'll always prioritize a ready PR to a possible future PR. " This I agree with :+1:

PR #1303, opened 2016-09-13
UI Right Menu background

Ohda:
Another suggestion for temporary better.
I've used the menu construction background image.

### DIFF
--- a/Assets/Resources/UI/MenuRight.prefab
+++ b/Assets/Resources/UI/MenuRight.prefab
@@ -202,7 +202,7 @@ MonoBehaviour:
   m_Padding:
     m_Left: 4
     m_Right: 4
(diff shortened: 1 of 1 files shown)


sboigelot:
You lost text space:
image

Also you should probably flip the texture vertically.

Tranberry:
I would reflect or flip the UI elements

TomMalbran:
That looks better.

For the future I was thinking that the bottom menu which shows information on mouse move could not have any background or just a semitransparent black, but in a way which it doesn't get too much attention. And the selection menu can pop on top of that menu (hiding it before), and it can have an x to unselect the stuff. RimWorld does something similar with the menus on the left.

koosemose:
Looks good. Merged.

PR #1304, opened 2016-09-13
Animations support for workshops

dusho:
Furniture that is workshop now supports animations. (animation for metal furnace for now)
image


sboigelot:
I can't test this because of an unrelated bug in master: #1305

sboigelot:
This seems to work beside the bug, but I'll personalty hold up before merging anything else until the bug is fixed or proven non-reproducible. sorry :/

Raketfart:
Only read the code, but this looks great.

PR #1308, opened 2016-09-14
Destroy the TimeManager instance

TomMalbran:
Fixes the issue on starting a new world and loading from a save.


Assets/Scripts/Controllers/TimeManager.cs
```diff
@@ -190,4 +190,12 @@ private void InvokeEvent(Action<float> eventAction, float time)
             eventAction(time);
         }
     }
+
+    /// <summary>
+    /// Destroy this instance.
+    /// </summary>
+    public void Destroy()
```
dusho: maybe call it Reset() ? usage looks weird in methods that should initialize the game (NewWorld, LoadWorld)

TomMalbran: In the class the name seems right. And in NewWord/LoadWorld it actually needs to destroy the previous state to load the new one.

TomMalbran: Reset might imply to go to a previous state/instance/starting point. Where this just destroys the static reference.

koosemose:
This fixes the loading issue, and the other related issues that cropped up afterwards. Of course, as discussed on Discord we should probably avoid new static instances in the future wherever possible. But for now this takes care of the issues. Merged.

PR #1309, opened 2016-09-14
The Mining Drone can now properly produce Raw Copper

TomMalbran:
Fixes the issue when you change the Mining Drone to mine Raw Copper you get a non existent inventory.


koosemose:
All looks good and clean, merged.

PR #1314, opened 2016-09-14
Fix Copper Bug for Mining Drones

kd7uiy:
Copper is an option for mining drones, but isn't working. Fixed this bug.


kd7uiy:
capture

kd7uiy:
I see this is a duplicate, closing.

PR #1316, opened 2016-09-14
[ART] Added new Ice sprite

TheRiven:
Due to popular request I have replaced the Ice.. sheet? with an ice cube! :laughing: 

iceset


Tranberry:
this will not melt as fast I hope!

dusho:
maybe you can also post how it looks in game? next to other inventory items..

PR #1318, opened 2016-09-14
Adds a mode to explicitly not allow dragging certain furniture

kd7uiy:
Added doors, heaters, airlocks, etc to that list.


vogonistic:
I think you used tabs in the xml.

Would it make sense to have `single` being default if nothing is defined? It will probably be more common than `border` and `path` for furniture.

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -49,6 +49,7 @@
         <LinksToNeighbours>false</LinksToNeighbours>
         <EnclosesRooms>true</EnclosesRooms>
         <CanReplaceFurniture typeTag="Wall" />
+		<DragType>single</DragType>
```
TomMalbran: Use spaces

TomMalbran:
I think we should go in the opposite direction. Right now the default is to drag if the entity is 1x1, which is just an arbitrary decision. We should make the default be no drag, and add a rect drag for furniture that allows it.

kd7uiy:
Okay, I've set the default drag mode to be "single", per suggestions.

TomMalbran:
I meant change ```BuildModeController.IsObjectDraggable()``` to check if the furniture is draggable instead of assuming yes for 1x1.

kd7uiy:
Good call, I like that better. I'm still calling the default mode "single", but furniture is not draggable if the mode is "single".

Assets/Scripts/Controllers/MouseController.cs
```diff
@@ -451,6 +451,8 @@ private bool IsPartOfDrag(Tile tile, DragParameters dragParams, string dragType)
                 bool withinXBounds = dragParams.StartX <= tile.X && tile.X <= dragParams.EndX;
                 bool onPath = tile.Y == dragParams.RawStartY || tile.X == dragParams.RawEndX;
                 return withinXBounds && onPath;
+            case "single":
```
TomMalbran: This probably isn't needed anymore.

kd7uiy:
I like the idea of setting the default for now to single, because we might decide some day that most objects should be draggable using some new approach. I really just don't know right now, so for now I'm going to suggest we keep it non-empty.

Assets/Scripts/Controllers/BuildModeController.cs
```diff
@@ -41,7 +41,7 @@ public bool IsObjectDraggable()
 
         Furniture proto = PrototypeManager.Furniture.Get(buildModeType);
 
-        return proto.Width == 1 && proto.Height == 1;
+        return proto.DragType!="single";
```
TomMalbran: You need spaces around !=

TomMalbran:
This looks good now.

koosemose:
Looks good, Stylecop clean. Merged

Issue #1312, opened 2016-09-14
Furniture loss of power event


Raketfart:
I don't know much about our power system, so this might already exist in some form.

For our furniture, I think it would be usefull to have a Lua function we can use when furniture is losing power.  Both for controlling what happends when losing power, and for setting a "power off" animation.
Then we could control that the oxygen compresser maybe leaks oxygen, when not powered.

Right now, it seems OnUpdate isn't called when power is out. Could we add two new actions that are called once?
<Action event="OnPowerOn" functionName="TurnOn" />
<Action event="OnPowerOff" functionName="TurnOff" />

Or maybe just add this as a separate poweroff update?
<Action event="OnUpdatePowerOff" functionName="OnUpdate_NoPower" />

Or is there a smarter way?

dusho:
you want separate animation for power off state? isn't there overlay when power is off? shouldn't all furniture fallback to idle state.. or you just want to define somehow what is the default animation for furniture (that will also be used for powered off furniture)

Raketfart:
There is an overlay for power off, but Lua can't switch to an idle state, because OnUpdate is not called. And I would imagine that furniture would behave differently when losing power. Maybe the clone dies in the cloning machine.

dusho:
ah ok.. cause mostly on power off things can go into idle state (if defined) as default
additionally event can be triggered (if lua wants to override it)

Raketfart:
I could easily make the animation return to default (idle) state, but I imagine that having control over this i Lua would be nice. I can make a PR with the suggested change.

TomMalbran:
Sure we can add more events triggered when the furniture looses its power and when the power is really low, and any other event we need. Those are easy to add.

Issue #765, opened 2016-08-24
[Discussion] LUA functions vs item names


racerand:
Me and @abackwood were just discussing how the naming system of `type_material` in snake_casing looks bad as a function name in LUA like `generator_oxygen_OnUpdate`. What do you guys think of this?

bjubes:
I would keep Lua functions as they were with `OxygenGenerator_OnUpdate` As it makes more sense to have a capitalized function

TomMalbran:
We can go for something like ```GeneratorOxygen_OnUpdate``` so that it keeps the furniture name order.

racerand:
I like what @TomMalbran suggests

TomMalbran:
@Grenkin1988 Should we change the names, or keep them as they are?

Grenkin1988:
@TomMalbran They are not changed yet? This is 14 days old discussion without any updates. I guessed that this is done already.

TomMalbran:
@Grenkin1988 No. We haven't even finished changing the furniture names in the xml yet.

Grenkin1988:
OK. Than this needs to be done. And not just lost in time.

TomMalbran:
I agree, but it will be easies once the Xml has the new standard names.

We should also add this standard somewhere in the none Lua standards.

PR #1319, opened 2016-09-14
Furniture loss of power event

Raketfart:
I added a trigger for furniture, when they lose power, as per issue #1312.

I used the new animation system for:

Oxygen generator - now animates, and when power is off it returns to idle state.

Oxygen compressor - This doesn't return to an idle state when turned off, but it leaks half of it's compressed gas. 
I'm not sure if that's the effect it should have, but it's a nice testcase for this event trigger, as we don't always want furniture to return to "idle" when unpowered.

Feedback is appreciated!

### DIFF
--- a/Assets/StreamingAssets/LUA/Furniture.lua
+++ b/Assets/StreamingAssets/LUA/Furniture.lua
@@ -17,7 +17,7 @@ ENTERABILITY_SOON = 2
 -- ModUtils.ULogError("Testing ModUtils.ULogErrorChannel") -- Note: pauses the game
 
 -------------------------------- Furniture Actions --------------------------------
-function OnUpdate_GasGenerator( furniture, deltaTime )
+function OxygenGenerator_OnUpdate( furniture, deltaTime )
     if ( furniture.Tile.Room == nil ) then
 		return "Furniture's room was null."
 	end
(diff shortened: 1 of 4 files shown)


Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -467,16 +473,18 @@ function OxygenCompressor_OnUpdate(furniture, deltaTime)
             furniture.UpdateOnChanged(furniture)
         end
     end
+	furniture.SetAnimationState("running")
```
TomMalbran: We should use spaces.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -31,8 +31,14 @@ function OnUpdate_GasGenerator( furniture, deltaTime )
         end
     end
 	return
+	furniture.SetAnimationState("running")
 end
 
+function OnPowerOff_GasGenerator( furniture, deltaTime )
```
TomMalbran: Is the convention to have first the furniture and then the action or the other way around. I like first Furniture then action.

Raketfart: I'm not sure... we have half one and half the other. I think I agree with you - I was just trying to be consistent

TomMalbran: We have an issue #765, where it looks like we agree on `funitureName_functionName`. But the furniture names might change in the future.

Raketfart:
I renamed the actions to furniture_action instead of the other way around, and renamed gasgenerator to oxygengenerator

koosemose:
Looks good, Stylecop clean. Merged.

PR #1324, opened 2016-09-14
Add filter to construction menu

sboigelot:
ezgif com-crop

> Just because I'm blind with these kind of list

### DIFF
--- a/Assets/Scripts/Controllers/KeyboardManager.cs
+++ b/Assets/Scripts/Controllers/KeyboardManager.cs
@@ -11,6 +11,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using UnityEngine;
+using UnityEngine.UI;
 
 public class KeyboardManager
 {
@@ -22,11 +23,14 @@ public class KeyboardManager
     {
         instance = this;
         mapping = new Dictionary<string, KeyboadMappedInput>();
+        ModalInputFields = new List<InputField>();
 
         TimeManager.Instance.EveryFrameNotModal += (time) => Update();
 
         ReadXmlOrJsonAfterWeDecide();
     }
+
+    public List<InputField> ModalInputFields { get; set; }
 
     public static KeyboardManager Instance
     {
@@ -43,6 +47,14 @@ public class KeyboardManager
         set
         {
             instance = value;
+        }
+    }
+
+    public void RegisterModalInputField(InputField filterField)
+    {
+        if (!ModalInputFields.Contains(filterField))
+        {
+            ModalInputFields.Add(filterField);
         }
     }
 
@@ -68,6 +80,11 @@ public class KeyboardManager
 
     public void Update()
     {
+        if (ModalInputFields.Any(f => f.isFocused))
+        {
+            return;
+        }
+
         foreach (KeyboadMappedInput input in mapping.Values)
         {
             input.TrigerActionIfInputValid();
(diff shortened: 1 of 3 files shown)


vogonistic:
Does it have focus automatically or does it require a click? I'm asking because I'm hoping someone will implement a DF or Warcraft-like set of short keys, so `c d` would be short for construct door. If the filter doesn't have automatic focus then both methods can coexist.

sboigelot:
@vogonistic no automatic focus for now but it's one extra line of code if we decide to go for it

Raketfart:
Good idea! 
Is there any chance of increasing the scroll speed of the construction menu, while we're at it?

sboigelot:
@Raketfart sure, no prob, done :)

bjubes:
The menu seems to be anchored incorrectly. This is what i'm getting.
capture

jaselito:
Having the same problem as bjubes...

jaselito:
Somehow, the Pos Y in the rect transform of the construction menu prefab got set to -384... getting it back to 0 should solve the problem

sboigelot:
@bjubes @jaselito the anchor position of the construction menu should be fix now.

Assets/Scripts/Controllers/KeyboardManager.cs
```diff
@@ -22,12 +23,15 @@ public KeyboardManager()
...
         instance = this;
         mapping = new Dictionary<string, KeyboadMappedInput>();
+        ModalInputFields = new List<InputField>();
 
         TimeManager.Instance.EveryFrameNotModal += (time) => Update();
 
         ReadXmlOrJsonAfterWeDecide();
     }
 
+    public List<InputField> ModalInputFields { get; set; }
```
mikejbrown: Stylecop complains: `SA1204 : CSharp.Ordering : All static public properties must be placed before all non-static public properties. (Assembly-CSharp)`

Assets/Scripts/UI/InGameUI/MenuLeft/ConstructionMenu.cs
```diff
@@ -50,6 +50,24 @@ public void RebuildMenuButtons(bool showAllFurniture = false)
...
+
+        List<Transform> childs = contentTransform.Cast<Transform>().ToList();
+
+        foreach (Transform child in childs)
+        {
+            Text buttonText = child.gameObject.transform.GetComponentInChildren<Text>();
+
+            string buildableName = buttonText.text;
+
+            bool nameMatchFilter = string.IsNullOrEmpty(filterText) || buildableName.ToLower().Contains(filterText.ToLower());
```
mikejbrown: Matches are not anchored to the start of the name, e.g. if I type "dd" this matches "Ladder". Is this desired behaviour?

sboigelot: Yep this is the desired behavior. I want to see "Cloning pod" if I search pod for instance.

Assets/Scripts/Controllers/KeyboardManager.cs
sboigelot: Fixed, thanks!

Assets/Scripts/UI/InGameUI/MenuLeft/ConstructionMenu.cs
mikejbrown: Good point :+1:

PR #1326, opened 2016-09-15
Adding z lvl to thermal

Geoffrotism:
Add 3D simulation for thermal changes. 
Updated LUAs to reflect this. 

- [x] fix diffusivity so it is not applied vertically to all z lvls.
- [x ] have code only iterate through enclosed rooms when changing temperature (not feasible with current code)
- [x] adding/deleting rooms needs to work properly.

follows #1326


NogginBops:
Why should temperature only be tracked in enclosed rooms?

Geoffrotism:
Less calculations. Is there a reason to track temperature in space?

Geoffrotism:
Update: PR still under work. Can't continue until issue #1406 is resolved.

koosemose:
Looks good, there's of course room for further improvements (such as not transmitting heat where there's no atmosphere or furniture or anything else perhaps), but it certainly seems an improvement, and I think this is a good point for this PR. Merged.

PR #1327, opened 2016-09-15
Pass Z values to overlays

koosemose:


sboigelot:
It takes a while for the overlay to update when you change the layer with the overlay open. Is this intended? Shouldn't we trigger a refresh when we change the visible layer?

koosemose:
Probably should trigger a refresh, but I'm not entirely sure what the proper way to trigger it to refresh is.

koosemose:
Looks like it's completely self contained... I think the best I can do is keep track of the current layer in OverlayMap, and when the number is different then call bake and update the layer tracking variable...

koosemose:
It's not optimal and it feels hackish, but it's the only way I can see to do it. If you see a better way, please let me know.

As a side note, I've just realized how slow Baking a new overlay is... I wonder if there's a way to do it in another thread or something...

mikejbrown:
I don't see how to significantly optimize it apart from somehow speeding up `GenerateTexture` itself. You could try pre-baking all the layers (or some number of them) and caching the textures somewhere. But realistically there would probably be updates before you switch anyway. `GenerateTexture` is slow but unfortunately necessary as far as I can tell.

Assets/Scripts/UI/Overlay/OverlayMap.cs
```diff
@@ -351,11 +353,17 @@ private void Update()
             elapsed = 0f;
         }
 
+        if(currentOverlay != "None" && currentLayer != WorldController.Instance.cameraController.CurrentLayer)
```
mikejbrown: Stylecop `SA1000 : CSharp.Spacing : The spacing around the keyword 'if' is invalid. (Assembly-CSharp)`

koosemose:
Yeah, I don't see much of a way either... just figured I'd toss it out just in case someone comes by and has a clever idea of how to do it :D

Assets/Scripts/UI/Overlay/OverlayMap.cs
```diff
@@ -351,11 +353,17 @@ private void Update()
...
+        {
+            Bake();
+            currentLayer = WorldController.Instance.cameraController.CurrentLayer;
+            elapsed = 0f;
+        }
         // TODO: Prettify.
```
mikejbrown: Stylecop `SA1515 : CSharp.Layout : A single-line comment must be preceded by a blank line or another single-line comment, or must be the first item in its scope. To ignore this error when commenting out a line of code, begin the comment with '////' rather than '//'. (Assembly-CSharp)`

sboigelot:
About the speed, I'm not sure what does the Graphics.CopyTexture underneeth but I had good improvement by not generating part of the texture at a time but instead copying info to a Color array and then only creating/updating the texture once.

But this is out of this PR, I can git it a shot later this week end maybe

Assets/Scripts/UI/Overlay/OverlayMap.cs
koosemose: If this is otherwise mergeable, could a Gatekeeper fix these? As I won't be able to access it until tomorrow. I have *Allow edits from maintainers.* turned on, so I'm assuming it's possible (never had the opportunity to try it myself).

sboigelot: I'll try now

sboigelot: Well I pushed the change... somewhere. :p Github is a bit laggy laterly, let's see if it will update

sboigelot: Hem.. it pushed it in my own fork. Sorry no idea for now how to help
https://github.com/sboigelot/ProjectPorcupine/commit/480432fbda9c93cbf6825c1c0a7bb3d0c18b5403

mikejbrown: I'll try then :D

mikejbrown: Hey hey that did it! Thank you github https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/

sboigelot:
I made #1354 about the speed, if you can have a look and tell me what you think it would be great

PR #1331, opened 2016-09-15
[HotFix] Fix lua to use job.buildable

koosemose:
Mining drone and many other furniture's don't work because I didn't properly update them to use job.buildable in lua. This fixes that, along with changing all incidences of `j` to `job`.


koosemose:
This being a hotfix for something that is currently broken in master, and no other Gatekeepers presently being available, I'm going to self-merge this.

PR #1333, opened 2016-09-15
Fix context actions from lua

kd7uiy:
There is an unusual bug that this fixes that is caused by the "Action" being placed in the same piece of memory for each object yielded. The bottom line is, only the second action was possible. I'm not sure why this is the case, and I tried many means to force the two to occupy different pieces of memory, but I failed. This workaround, however, works well, and it can make it easier to debug such issues in the future.

### DIFF
--- a/Assets/Scripts/Models/ContextMenu/ContextMenuAction.cs
+++ b/Assets/Scripts/Models/ContextMenu/ContextMenuAction.cs
@@ -16,6 +16,8 @@ public class ContextMenuAction
 
     public string Text { get; set; }
 
+    public string Parameter;
+
     public void OnClick(MouseController mouseController)
     {
         if (Action != null)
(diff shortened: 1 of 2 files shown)


koosemose:
Looks good, fixes the context menu problem, no new bugs seem to have cropped up, and stylecop clean. Merged.

TomMalbran:
BTW the workshop context menu actions had a similar issue when they were being worked with and the solution was to move the creation to a new function:
```
if (IsWorkshop && workshop.WorkshopMenuActions != null)
         {
             foreach (WorkshopContextMenu factoryContextMenuAction in workshop.WorkshopMenuActions)
             {
                 yield return CreateWorkshopContextMenuItem(factoryContextMenuAction);
             }
         }
```
The same can be done with the normal Furniture Actions.

kd7uiy:
I'm all for a better solution, if you want to do it. I'm burned out on the issue for today...

TomMalbran:
Sure I can try if that solves it. It was a weird bug. Is a Scope issue and wasn't solved by saving variables in the scope. @dusho had an issue with it, and we tested a few variants until I went with the solution that I knew that will work and tried with a function.

dusho:
Yes, I think it is old Mono bug. According to Unity blog this was fixed for Unity 5.4 (now in beta) as a first step to bring newer .Net to Unity.

kd7uiy:
I'm running 5.4 and found the issue. I assume you mean 5.5?

dusho:
Yep. 5.5 , sorry.
Details: http://forum.unity3d.com/threads/upgraded-c-compiler-on-5-3-5p8.417363/

PR #1334, opened 2016-09-15
Remove meta file that is no longer required.

kd7uiy:


TomMalbran:
Since you are removing unused metas, maybe you can also remove ProjectPorcupine/Assets/Scripts/Models/State.meta

jaselito:
@TomMalbran My current PR already removes it (in fact, it also removes the meta in this PR... should I revert that change and leave both removals to this PR?)

TomMalbran:
I think that I also removed them in one of my PRs. You probably don't need to do anything. If your PR gets merged first, then we close this one. If this one gets merged first, then is fine since it will not cause any conflicts.

jaselito:
:thumbsup:

kd7uiy:
I just got yelled at by Quill for one of my first PRs because I let meta files sneak in without meaning to do so, so it's my habit to do these as separate PRs. Still, we'll see what happens.

TomMalbran:
Adding meta files is bad. But removing unused ones in a PR really makes no harm.

Is better to do it a separate PR, just because these are easy to merge and improve the project.

PR #1338, opened 2016-09-15
Implement Vents

koosemose:
Vents equalize pressure between the rooms on either side.

Super Sexy Programmer Art Included. :)


Assets/StreamingAssets/Images/Furniture/Vent.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Sprites>
+	<Sprite name="vent"    x="0" y="0" w="1" h="1" pixelPerUnit="64" />
```
vogonistic: There is a tab here ☝️

koosemose: Bloody Notepad++ no matter how much I tell it not to, it keeps sneaking tabs in when I'm not looking.

koosemose:
Changed

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -547,6 +547,43 @@
...
+
+        <Action event="OnUpdate" functionName="Vent_OnUpdate" />
+        <Action event="OnChange" functionName="Vent_SetOrientationState" />
+        <Action event="OnInstall" functionName="Vent_SetOrientationState" />
+
+        <ContextMenuAction FunctionName="Vent_Open" Text="Open Vent" RequireCharacterSelected="false" />
+        <ContextMenuAction FunctionName="Vent_Close" Text="Close Vent" RequireCharacterSelected="false" />
+
+        <Animations>
+          <Animation state="horizontal" valuebased="true">
```
TomMalbran: 4 spaces for indentation on all this animation lines, please.

Assets/StreamingAssets/Images/Furniture/Vent.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Sprites>
+    <Sprite name="vent"    x="0" y="0" w="1" h="1" pixelPerUnit="64" />
```
TomMalbran: There should either be 1 space between name and x, or have x be aligned with the other x.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -560,6 +560,37 @@ function AirPump_GetSpriteName(furniture)
     return furniture.Type .. suffix
 end
 
+function Vent_OnUpdate(furniture, deltaTime)
+    furniture.SetAnimationProgressValue(furniture.Parameters["openness"].ToFloat(), 1)
+    furniture.Tile.EqualiseGas(deltaTime * furniture.Parameters["gas_throughput"].ToFloat() * furniture.Parameters["openness"].ToInt())
+end
+
+function Vent_SetOrientationState(furniture)
+    local t = furniture.Tile
```
TomMalbran: Call this tile.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -560,6 +560,37 @@ function AirPump_GetSpriteName(furniture)
...
+    if (furniture.Tile == nil) then
+        return furniture.Type
```
TomMalbran: Do we need to return a value?

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -560,6 +560,37 @@ function AirPump_GetSpriteName(furniture)
...
+    end
+    
+    local north = World.Current.GetTileAt(t.X, t.Y + 1, t.Z)
```
TomMalbran: Use tile.North() ... tile.South()  ... tile.West() ... tile.East() ... You might not even need the variables.

koosemose:
Changed

TomMalbran:
Did you forgot to commit?

koosemose:
I've been having issues with my commits syncing properly all day... I've had to sync most of them 3 or 4 times.

TomMalbran:
Cool. The code looks good now. Haven't tested it, but it works.

I like how you implemented the sprite rotation in Lua code. Not sure if we should add better support for it in C# or keep it to Lua, since a similar logic can be applied to the doors, which is kinda hardcoded at the moment. (For a future PR)

koosemose:
I mostly stole the rotation code from the air pump, and then changed it over to use animation states, and actually the door now (partially) uses animation states to control it's rotation, it could probably do it in an OnInstall action as well.

Another thing for a future PR is a good place to send an OnChanged event to lua, but I couldn't find a good place to call it from. There's no good reason to have to set the openness of the vent every update, other than lack of an OnChanged event sent to Lua.

TomMalbran:
Yes. Since it is used by a few furnitures, I am not sure if we should add support in C# (maybe some utility functions) or just try and do it in Lua for all the furnitures. We can use that to remove the verticalDoor param in Furniture, or make it in a way so that it can work for any furniture that requires it.

Do you mean calling UpdateOnChanged from Lua which triggers a Changed event? The issue with the Changed event is that is usually triggered from the OnUpdate Lua functions.

koosemose:
No I mean when the Changed event is fired from elsewhere, having it send `EventActions.Trigger("OnChanged", furn)`. It seems strange that we can react to Installation, Removal, and Update, but not to a change.

TomMalbran:
That should be easy to do. We can add the trigger in UpdateOnChanged (make that public method and change the name to InvokeChanged), and then make it so that everywhere were we call the Changed event directly to make it call UpdateOnChanged (InvokeChanged).

koosemose:
It should be noted that the things being discussion in the latter half aren't things that are to be implemented in this PR, just ways that the surrounding system could be improved in ways useful to the vent. This PR is (barring any mistakes not yet noticed or the like) complete.

sboigelot:
Works great, no style issue, merging.

PR #1339, opened 2016-09-15
Put all unit tests in the unit test folder

kd7uiy:


Workshop.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
TomMalbran: You probably shouldn't commit this file. The Workshop test creates this files in the root and never deletes it, which is wrong.

dusho:
Ye. Sorry about that. Was actually using that file as a template for furniture settings in furnitures.xml
For future reference. Is there temp folder used for saving some files? That are maybe in .gitignore?

TomMalbran:
You probably should delete the file after the test. Not sure about a folder, maybe we could have one.

kd7uiy:
There seems to be a /Temp folder, that might be a good place for it?

mikejbrown:
Why is the test reading/writing files at all? It should not. It should be using a string instead.

mikejbrown:
All of the tests are still being found and passing. No unexpected breakages. Merging. The FurnitureWorkshopTest needs to be fixed in its own PR.

PR #1342, opened 2016-09-15
Disabled debugging code for the FastPriorityQueue,

kd7uiy:
The debug for Unity is set by default for in editor games. For this queue, it will dramatically slow down things to ensure the implementation is correct, but isn't needed now.


kd7uiy:
Here's profiling before and after:
capture

capture

mikejbrown:
Just sets a flag, doesn't actually remove debug calls so this could always be turned back on if need be. :+1:

PR #1345, opened 2016-09-16
Remove old Cloning Pod, correcting center point 

kd7uiy:
capture


Geoffrotism:
Why is the cloning pod 2x3? Is clearly is two tiles wide but it only goes back one tile according to the size of the base. It's just really tall.

TomMalbran:
The graphic is technically 2x3. Anything smaller and it will get clipped. I think that it should be change to either be 2x2 or a different thing.

Geoffrotism:
The graphic might be 2x3 in pixels but the visual representation makes it look like a 2x1. Basically it looks like you should be able to place some on the tile in the middle of that room.

TomMalbran:
But we don't have support for layers, so placing anything behind will not work, and we might not even need to support them with the right perspective. The real issue is that the perspective is kind of wrong in that graphic, is too facing forward for our game.

So this change is fine for how the game works and how the graphic was done.

dusho:
We don't need layers to solve overlaps of overreaching tiles.. Do we? Just need to set render order of furniture according to Y position - same as we have for characters.

Raketfart:
I agree with dusho. We should probably set sortorder on furniture anyway, so characters can walk behind furniture

kd7uiy:
capture

PR #1346, opened 2016-09-16
[Tiny ]Small refactoring

jaselito:
Making  some names clearer


vogonistic:
For future PRs, the title should be more descriptive so that people can determine if you are changing something where they are also making changes.

jaselito:
Roger that!

PR #1350, opened 2016-09-16
Refactored Pathfinder and InventoryManager interface

vogonistic:
This is the first PR of several with independent changes that has been extracted from the character state machine PR #1219 

Pathfinding has been refactored to return a list of tiles instead of an object. There is also a wrapper interface with convenience methods over Path_AStar so that we can replace underlying logic if we want. Path_AStar now takes a goal matching function and cost heuristic function to be provided so we can search for pretty much anything.

InventoryManager has also been refactored to provide better checks to quickly find if inventory exists and ensure that we always check for locked inventory.


gunthergun:
Tested it and it works 👍🏼

Assets/Scripts/Pathfinding/Pathfinder.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            return tile => 0f;
+        }
+
+        private static void DebugLog(string message, params object[] par)
+        {
+            Debug.ULogChannel("Pathfinding", message, par);
+        }
+
+        private static void DebugLogIf(bool condition, string message, params object[] par)
```
mikejbrown: :heart: This could probably be a utility function somewhere public and static (with channel and log level variations exposed as well). Maybe for another PR though...

Assets/Scripts/Models/Character.cs
```diff
@@ -1017,9 +999,9 @@ private bool CheckForJobMaterials()
 
     private bool WalkingToUsableInventory()
     {
-        bool destHasInventory = pathAStar != null && pathAStar.EndTile() != null && pathAStar.EndTile().Inventory != null;
+        bool destHasInventory = movementPath != null && movementPath.Last() != null && movementPath.Last().Inventory != null;
```
mikejbrown: I just did a test run and it looked like two characters (they were on top of each other) walked to pickup the same inventory item. When they reached the spot, one grabbed the inventory and walked away and the other stayed put (which is the only reason I could tell there were two there), and the log was spammed with
```
InvalidOperationException: Operation is not valid due to the current state of the object
System.Linq.Enumerable.Last[Tile] (IEnumerable`1 source)
Character.WalkingToUsableInventory () (at Assets/Scripts/Models/Character.cs:1002)
Character.CheckForJobMaterials () (at Assets/Scripts/Models/Character.cs:943)
Character.Update_DoJob (Single deltaTime) (at Assets/Scripts/Models/Character.cs:698)
Character.Update (Single deltaTime) (at Assets/Scripts/Models/Character.cs:336)
World.TickEveryFrame (Single deltaTime) (at Assets/Scripts/Models/World.cs:188)
TimeManager.InvokeEvent (System.Action`1 eventAction, Single time) (at Assets/Scripts/Controllers/TimeManager.cs:198)
TimeManager.Update (Single time) (at Assets/Scripts/Controllers/TimeManager.cs:131)
WorldController.Update () (at Assets/Scripts/Controllers/WorldController.cs:146)
```
each tick.

mikejbrown: Seems like the real problem is in the job handling code but I'm not sure what's really happening without a good way to reproduce the problem.

kd7uiy:
I don't know if you have it planned for a future update, but I find that Dictionaries are really slow when it comes to pathfinding algorithms. You might consider removing those...

Assets/Scripts/Pathfinding/Pathfinder.cs
vogonistic: My way to deal with our style guide not letting me write if-statements without bracket. It would make sense to expand and make available more generally.

Assets/Scripts/Models/Character.cs
vogonistic: I can fix the error log. The job problem is for another PR.

sboigelot: Shoudn't it just be  movementPath.Last**OrDefault**() != null

mikejbrown: @sboigelot That would probably be a good change and would probably stop the log spam, but I think the deeper issue is that somehow two characters are getting the same job in odd circumstances.

Assets/Scripts/Models/InventoryManager.cs
```diff
@@ -114,20 +113,27 @@ public bool PlaceInventory(Character character, Inventory sourceInventory, int a
     /// Gets <see cref="Inventory"/> closest to <see cref="startTile"/>.
     /// </summary>
     /// <returns>The closest inventory of type.</returns>
-    public Inventory GetClosestInventoryOfType(string type, Tile startTile, int desiredAmount, bool canTakeFromStockpile)
+    public Inventory GetClosestInventoryOfType(string objectType, Tile tile, bool canTakeFromStockpile)
```
TomMalbran: objectType should be type.

Assets/Scripts/Models/InventoryManager.cs
```diff
@@ -159,34 +165,59 @@ public bool RemoveInventoryOfType(string type, int quantity, bool onlyFromStockp
...
+        }
 
-        // We can also avoid going through the A* construction if we know
-        // that all available inventories are stockpiles and we are not allowed
-        // to touch those
-        if (!canTakeFromStockpile && Inventories[type].TrueForAll(i => i.Tile != null && i.Tile.Furniture != null && i.Tile.Furniture.HasTypeTag("Storage")))
+        return Inventories[type].Find(inventory => InventoryCanBePickedUp(inventory, canTakeFromStockpile)) != null;
+    }
+
+    public bool HasInventoryOfType(string[] objectTypes, bool canTakeFromStockpile)
```
TomMalbran: Maybe use types here too.

Assets/Scripts/Models/InventoryManager.cs
```diff
@@ -10,6 +10,8 @@
 using System.Collections.Generic;
 using System.Linq;
 using MoonSharp.Interpreter;
+using UnityEngine;
```
TomMalbran: This should give a StyleCop error.

Assets/Scripts/Pathfinding/Pathfinder.cs
TomMalbran: Should we even keep all the debug logs? We can keep this function for testing, but nore sure about having all the logging in master.

Assets/Scripts/Pathfinding/Pathfinder.cs
```diff
@@ -0,0 +1,226 @@
...
+        public static List<Tile> FindPathToInventory(Tile start, string[] objectTypes, bool canTakeFromStockpile = true)
+        {
+            if (start == null || objectTypes == null || objectTypes.Length == 0)
+            {
+                return null;
+            }
+
+            Path_AStar resolver = new Path_AStar(World.Current, start, GoalInventoryEvaluator(objectTypes, canTakeFromStockpile), DijkstraDistance());
+            List<Tile> path = resolver.GetList();
+            if (path.Count > 0)
```
TomMalbran: There are some StyleCop errors in this file.

mikejbrown:
[Conflicts]

vogonistic:
@kd7uiy We should, but I have no plans on working on it short term.

vogonistic:
Merge conflicts fixed.

Assets/Scripts/Models/Character/Character.cs
```diff
@@ -757,31 +736,33 @@ private void Update_DoMovement(float deltaTime)
...
+            if (movementPath == null || movementPath.Count == 0)
             {
                 // Generate a path to our destination.
                 // This will calculate a path from current tile to destination tile.
-                pathAStar = new Path_AStar(World.Current, CurrTile, DestTile);
-                if (pathAStar.Length() == 0)
+                movementPath = Pathfinder.FindPathToTile(CurrTile, DestTile, MyJob != null ? MyJob.adjacent : false);
+                if (movementPath == null || movementPath.Count == 0)
                 {
                     Debug.ULogErrorChannel("Character", "Path_AStar returned no path to destination!");
```
mikejbrown: If I leave the game running for a few minutes game time (about a minute at 4x) I start getting spammed by this every tick.

vogonistic: It shouldn't depend on time. It is a Job, or Inventory that can't be reached.

I can remove the log, but I don't consider it part of the refactoring.

sboigelot:
@vogonistic I would suggest to add a [WIP] tag as long as you feel there is still work on it based on the comment and remove it when you feel it's good to go.

Assets/Scripts/Pathfinding/Pathfinder.cs
vogonistic: I've changed all the calls to DebugLog to be conditional on a define that isn't set, so it's quiet now.

Assets/Scripts/Models/Character.cs
vogonistic: I've changed the code to avoid the null exception, using @sboigelot's suggestion. 

As for booking material, I think that's a whole other PR.

vogonistic:
This isn't a WIP. Unless there are more feedback to implement, this is ready to merge.

Assets/Scripts/Models/Character/Character.cs
```diff
@@ -818,8 +799,8 @@ private void Update_DoMovement(float deltaTime)
         // But when we do the pathfinding system, we'll likely
         // switch to something like Manhattan or Chebyshev distance
         float distToTravel = Mathf.Sqrt(
-            Mathf.Pow(CurrTile.X - nextTile.X, 2) +
-            Mathf.Pow(CurrTile.Y - nextTile.Y, 2));
+                                 Mathf.Pow(CurrTile.X - nextTile.X, 2) +
```
TomMalbran: Just out of curiosity. What is adding so much space before. In MonoDevelop when I press enter after ( it just indents the code 4 space. I also think that having the last ```);``` in a new line might look better, if the indentation is 4 spaces.

vogonistic: MonoDevelop is adding them. It's also correcting white space all over the file. I wonder why it isn't changing it to this for you. I have set the settings to what is in the contribution guide.

sboigelot:
@vogonistic  I'm getting an exception, just start a game, get out of dev mode and try to build a floor:

```
ArgumentOutOfRangeException: Argument is out of range.
Parameter name: index
System.Collections.Generic.List`1[Tile].get_Item (Int32 index) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System.Collections.Generic/List.cs:633)
Character.Update_DoMovement (Single deltaTime) (at Assets/Scripts/Models/Character/Character.cs:764)
Character.Update (Single deltaTime) (at Assets/Scripts/Models/Character/Character.cs:340)
CharacterManager.Update (Single deltaTime) (at Assets/Scripts/Models/Character/CharacterManager.cs:54)
World.TickEveryFrame (Single deltaTime) (at Assets/Scripts/Models/Area/World.cs:189)
TimeManager.Update (Single time) (at Assets/Scripts/Models/Events/TimeManager.cs:131)
WorldController.Update () (at Assets/Scripts/Controllers/WorldController.cs:145)

```

kd7uiy:
Conflicted.

vogonistic:
Fixed conflicts and the errors. The error was when the character already was standing next to a job tile when you pick it up and adjacent is set to true. Good catch @sboigelot!

vogonistic:
@dusho Found a large drop in FPS when the characters run out of jobs. Tagged as WIP until it's found.

koosemose:
I'm removing the WIP from this, since as far as I can tell, things work as they should, and I'd like to get this code into master as soon as possible, and allow others to do any enhancements to it from there, since vog will most likely be unable to keep up with getting this to perfect working condition.

I would however like additional verification that nothing creates any horrendous issues before merging.

dusho:
Can't see any FPS drops anymore, thought there seems to be some other issue now.
It's like characters are doing the jobs (hauling, etc.) , so moving, but they don't carry any inventory.. so when they are at endTile for job, they are keeping their jobs, but setting movementPath to null (Character.cs, ln 730).. then every character will get stuck in this state and no other job can be taken.

koosemose:
I've been able to reproduce this behavior building "ladders" and sometimes walls. So far I'm pretty sure they're finishing the first job but something strange is going on with the inventory they think they have proper inventory in their tile, but can't work their job, so they're never going to look for more. I think it has something to do with working adjacent tiles, but I'm not completely sure. Continuing further investigation.

koosemose:
Ah HAH! I've figured out what exactly is happening. They're going looking for inventory properly, but if they happen to pass over their work tile (as will happen often if you build a bunch at once), Update_DoMovement sees that they're at their jobsite and bails regardless of if it has proper inventory or not. Now I just need to figure out exactly how to fix this.

dusho:
tested the latest.. chars are still somehow ignoring jobs from furniture (reactor needing fuel, furnace needing raw iron, ...) instead all those materials are put into stockpile

koosemose:
I am unable to recreate the issues reported by dusho in the latest version. If anyone else can confirm the presence or absence of this bug in the latest, it would be greatly appreciated.

dusho:
ok so it was confirmed that there is issue that if there is no itemType requested by job elsewhere in world, job gets abandoned. Can be reproduced by spawning 1 power cell -> job for filling power generator gets dropped and power cell is taken to stockpile. Even though HeldInventory of job is checked.. there is none yet.. not sure what was the idea how it should be used.

koosemose:
Held inventory I think is for the inventory that has already been placed in the world. 
The issue was that it was doing the check for any inventory on the map before it checks for the character's inventory (as far as InventoryManager is concerned a character's inventory isn't "on the map" since no other character can access it, and that character should be able to make use of it)

Incoming fix.

koosemose:
As far as I can tell, this last push fixes the reported issues. If folks could give it a look over once more, it would be appreciated.

dusho:
it did indeed fix the problem with 1 item jobs
but there still seems to be problem with inventory for furniture.. for metal smelter (raw iron x20) first went to stockpile.. and only then to furnace. And Metal smelter looks totally funny - one char brings inventory to input, then another takes it away, but then it flips back to input and so on... it's like Locked state on inventory is ignored and chars take inventory.. but then it kicks back in and inventory appears at input again
image

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -282,7 +282,7 @@ function MetalSmelter_UpdateAction(furniture, deltaTime)
         if (furniture.Parameters["smelttime"].ToFloat() >= furniture.Parameters["smelttime_required"].ToFloat()) then
             furniture.Parameters["smelttime"].SetValue(0)
 
-            ModUtils.LogError("MetalSmelter: Placing inventory at :" .. outputSpot.x .. ":" .. outputSpot.y)
+            ModUtils.ULog("MetalSmelter: Placing inventory at :" .. outputSpot.x .. ":" .. outputSpot.y)
```
TomMalbran: My Bad. You can totally remove this line.

koosemose:
Regarding ore going to stockpile first, it's inconsistent, and depends on the order of jobs in the queue, they will sometimes put it in the stockpile first, sometimes into the furnace first. However, that same behavior is present in master as well, the changes to how finding inventory works with the Pathfinder in between class may have changed the frequency with which it happens, but it still happens with both.

dusho:
Ok, then this will get fixed with some other PR, JobManager? Stockpile jobs should have maybe their own priority that no other job will reach.. so stockpiling is just a backup job..(at least for now?)

koosemose:
Merging

Issue #1315, opened 2016-09-14
Working furniture from outside of the furniture


koosemose:
Currently all of our furnitures have their JobSpotOffset within the bounds of the furniture, and while this makes sense for some furniture, it doesn't make as much sense for other furniture to stand on top of it while you work it. 

The system as is handles JobSpotOffsets outside the furniture just fine, but for optimum appearance and user experience a few changes would need to be made. 

The first is to not having the mouse "grab" furniture by it's JobSpotOffset (as this just looks silly when that offset is outside of the furniture), this is an incredibly easy change (just removing the lines of code that check for JobSpotOffset and center the furniture being placed on that, it will return to the default of being placed by the center). 

The second is to have furniture make sure its JobSpot is clear, which is still fairly easy, I've already implemented it in a test branch (currently only checking that it is clear of furniture, it should also check for pending jobs that would interrupt, but highly doubt that check would be any more difficult).

The third is simply showing some indicator of where the JobSpot is, though perhaps only if it isn't inside the furniture (I'm not absolutely sure which is the best option, only show when outside of furniture, or always show).

Later of course, when we have chairs and things we will need to alter the check to only check for furniture that isn't those types of things that would be allowable.

Thoughts on implementing more support for JobSpots outside of the furniture?

TomMalbran:
I think that is basically it. Is what Quill talk about in his videos but never got to implement.

Another thing that we might think of is the spawn job spot, which could later be outside the furniture, and we could make it spawn in some sort of stockpile furniture. Similar could be with an input spot.

Raketfart:
Yeah, spawning onto a conveyor belt. Automation here we come

dusho:
if the furniture uses workshop functionality, you can define input/output slots wherever (offset from origin of furniture)

Raketfart:
Will we need to take that into account when placing the furniture, so you don't spawn on top of a wall or something?

dusho:
no, not when placing.. I was thinking about that.. but it can happen that you will place furniture and then place the wall or other furniture where output slot is
so currently workshop just doesn't provide any output if output slot is not available (Factorio style)

koosemose:
Not talking about the Input/Output, but where a furniture is worked from.

TomMalbran:
I know that you aren't. But the same logic can be applied to all 3 spots.

koosemose:
[...]

PR #1352, opened 2016-09-16
Full support for furniture workspots outside of furniture.

koosemose:
Furniture workspots were already capable of existing outside of the furniture, however there were some oddities when doing so. This PR fixes those oddities and includes other support for a better experience with workspots outside of tile. Mining Drone station changed to be 3x2 and have it's workspot outside for demonstrative purposes (and because it looks better that way), empty space beyond 3x2 was cropped from the sprite to account for the new Furniture Height, with permission from @TheRiven.

* When placing furniture with a Workspot defined, the mouse no longer "grabs" it by the workspot, all furniture will use the default behavior of grabbing the center.

* When placing a furniture with a workspot outside of the furniture's bounds, a circle is shown where the workspot is relative to the furniture.

* Building disallowed when the workspot overlaps another furniture.

* Building disallowed when furniture is being placed on another furniture's workspot (either a furniture to be constructed or one already constructed.

* Workspots are allowed to overlap other workspots.

Caveats:
* Currently no furniture is allowed to be placed on a workspot, in the future when we implement furniture (such as chairs) which we want to exist in a workspot, it will have to be decided how to determine what furniture can, and how to implement it.

* Currently only one workspot is allowed per furniture, and is appropriately accounted for by this PR, however, in the future we may wish to allow a furniture to have multiple workspots (if, for example, where you can sit at around a table is defined as workspots). If this is desired behavior in the future, this feature will have to be modified to account for it.

As discussed in #1315


sboigelot:

Assets/Scripts/Controllers/MouseController.cs
```diff
@@ -219,22 +219,11 @@ private void CalculatePlacingPosition()
...
-            }
-            else
-            {   
-                // Otherwise we use the center.
-                currPlacingPosition = new Vector3(
-                    currFramePosition.x - ((proto.Width - 1f) / 2f),
-                    currFramePosition.y - ((proto.Height - 1f) / 2f),
-                    WorldController.Instance.cameraController.CurrentLayer);
-            }
+            // Otherwise we use the center.
```
TomMalbran: This should say "Use the center of the Furniture"

Assets/Scripts/Controllers/MouseController.cs
```diff
@@ -580,6 +570,7 @@ private void ShowFurnitureSpriteAtTile(string furnitureType, Tile t)
         sr.sprite = fsc.GetSpriteForFurniture(furnitureType);
 
         if (WorldController.Instance.World.IsFurniturePlacementValid(furnitureType, t) &&
+            WorldController.Instance.World.IsFurnitureWorkSpotClear(furnitureType, t) &&
```
TomMalbran: Should we just move this to ```IsFurniturePlacementValid```?

Assets/Scripts/Models/Functions/FunctionsManager.cs
```diff
@@ -112,6 +112,9 @@ public static void JobComplete_FurnitureBuilding(Job theJob)
     {
         WorldController.Instance.World.PlaceFurniture(theJob.JobObjectType, theJob.tile);
 
+        // Let our workspot tile know it is no longer reserved for us
+        WorldController.Instance.World.UnreserveTileAsWorkSpot((Furniture)theJob.buildablePrototype, theJob.tile);
```
TomMalbran: I don't get why you unreserve it here. Shouldn't it still be reserved so that other furniture can't be placed in this spot unless is a special kind of furniture like a chair?

Assets/Scripts/Models/Furniture.cs
```diff
@@ -778,6 +781,9 @@ public void Deconstruct()
         // Update thermalDiffusifity to default value
         World.Current.temperature.SetThermalDiffusivity(Tile.X, Tile.Y, Temperature.defaultThermalDiffusivity);
 
+        // Let our workspot tile know it is no longer reserved for us
+        WorldController.Instance.World.UnreserveTileAsWorkSpot(this);
```
TomMalbran: You can use ```World.Current```

Assets/Scripts/Models/World.cs
```diff
@@ -342,6 +342,44 @@ public Utility PlaceUtility(string objectType, Tile tile, bool doRoomFloodFill =
...
+    /// <param name="furn">The furniture whose workspot will be reserved.</param>
+    /// <param name="tile">The tile on which the furniture is located, for furnitures which don't have a tile, such as prototypes.</param>
+    public void ReserveTileAsWorkSpot(Furniture furn, Tile tile = null)
+    {
+        if (tile == null)
+        {
+            tile = furn.Tile;
+        }
+
+        World.Current.GetTileAt(
```
TomMalbran: We should create a function for this in World. A similar code is used in furniture.Jobs, and a few times here.

Assets/Scripts/Models/Tile.cs
```diff
@@ -41,6 +41,7 @@ public Tile(int x, int y, int z)
         Characters = new List<Character>();
         MovementModifier = 1;
         Utilities = new Dictionary<string, Utility>();
+        ReservedAsWorkSpotBy = new HashSet<Furniture>();
```
TomMalbran: This is a HashSet so that multiple furniture can use the same Work Spot? Right now is used as a boolean.

Assets/Scripts/Models/World.cs
```diff
@@ -342,6 +342,44 @@ public Utility PlaceUtility(string objectType, Tile tile, bool doRoomFloodFill =
         return PlaceUtility(util, tile, doRoomFloodFill);
     }
 
+    /// <summary>
+    /// Reserves the furniture's work spot, preventing it from being built on.
+    /// </summary>
+    /// <param name="furn">The furniture whose workspot will be reserved.</param>
+    /// <param name="tile">The tile on which the furniture is located, for furnitures which don't have a tile, such as prototypes.</param>
+    public void ReserveTileAsWorkSpot(Furniture furn, Tile tile = null)
```
TomMalbran: I think this should be at a Tile. The Furniture can call it as ```Jobs.GetWorkSpotTile().ReserveAsWorkSpot(this)``` and the job as ```Furniture furn = (Furniture)job.buildablePrototype; World.Current.GetTileAtOffset(job.Tile, furn.Jobs.WorkSpotOffset).ReserveAsWorkSpot(furn)```. And the HashSet in Tile can be private. But not totally sure.

Assets/Scripts/Models/Functions/FunctionsManager.cs
koosemose: No, the prototype isn't the same furniture as the instance of the furniture, this is removing the prototype, one using the furniture is added when the actual furniture is placed.

Assets/Scripts/Controllers/MouseController.cs
koosemose: Yeah, I caught this one after I was away from the code.

Assets/Scripts/Controllers/MouseController.cs
koosemose: I was of the opinion that it is conceptually separate from if the placement of the actual furniture is valid, so separated it out.

Assets/Scripts/Models/World.cs
koosemose: Hard to be entirely certain as I find it hard to parse code all on the same line. But the process of what's being done feels more complicated this way. I feel the current way is more straightforward.

Assets/Scripts/Models/Tile.cs
koosemose: I believe you are thinking of ReservedAsWorkSpot, which simply returns true if there is anything in ReservedAsWorkSpotBy (count > 0). A very simple check, but parses better than checking the count being greater than 0 from outside.

ReservedAsWorkSpotBy could perhaps use a lest linguistically awkward name, but I couldn't think of one.

Assets/Scripts/Controllers/MouseController.cs
TomMalbran: The work spot is part of the furniture, and the validity of the placement of the furniture also depends on that spot. So for me, Furniture placement should also check it.

Assets/Scripts/Models/Functions/FunctionsManager.cs
TomMalbran: Oh. I see. Yes I was thinking that if you want multiple furnitures to share the same spot. It will fail at the Job stage when trying to build the same Furniture.

Assets/Scripts/Models/World.cs
TomMalbran: Sorry. Was going to do it as 1 line and went with 2 since it was cleaner. I am not sure as to which way is cleaner. Most things are in Tile, so I figure it might work better there. But both ways are correct.

Assets/Scripts/Models/Tile.cs
TomMalbran: I got it now. ```IsReservedWorkSpot``` is only used when validating the furniture, but not when validating the Work Spot. So ```ReservedAsWorkSpotBy``` can be greater than 1.

koosemose:
Conflicts Resolved

kd7uiy:
There's a bug with this, see hotfix #1403. Try building a heater, or something, and it will not work because the build spot is in the furniture being built, and is already resolved. This might be a result of some of the conflicts that were resolved, some weird combination snuck in.

PR #1354, opened 2016-09-16
A faster overlay texture generation

sboigelot:
Change the overlay texture generation to improve speed

Using the following code to test the speed in the Bake method of overlay: (Not in the PR)
```
Stopwatch sw = new Stopwatch();
sw.Start();
GenerateTexture();
sw.Stop();
UberLogger.Logger.Log("none", this, LogSeverity.Message, "Overlay generation took " + sw.ElapsedTicks);
```

The speed is modified from
`from ~2700000 to ~3200000`
to
`from ~500000  to ~3000000`

So in some case it's not much faster but most of the time it is.


mikejbrown:
Is there someway to test (programmatically) that the same textures are generated?

sboigelot:
The same textures are not generated. I don't know why but the original code was make the texture 400 times too large for what we need.

mikejbrown:
Ahh ok... so it's a resolution reduction, not just a refactoring. (Which is fine if the visuals aren't really degraded of course.)

sboigelot:
Most of the lost time though is lost by the Get and Set Pixels (slow method in Unity)
We could even win more time by storing the `Dictionary<int, Color> colorMapLookup = new Dictionary<int, Color>();` in memory per overlay so we don't have to re-build it on every bake

sboigelot:
@mikejbrown it's both a resolution reduction and a refactor.

The refactor is just:

Instead of copying array of color directly to the texture tile per tile, create first a color array serving as buffer to only write once in the texture (avoid a lot of trouble with unity)

mikejbrown:
I'll take your word on that (this Unity centric stuff is above my pay grade). Can you do the `colorMapLookup` thing on this PR?

mikejbrown:
And to what extent is this just implementing our own custom framebuffer? ;)

kd7uiy:
@mikejbrown I'll certify that the approach he has included is a much better way than we had. It's recommended for sure for this complex stuff;-)

sboigelot:
@mikejbrown  

> Can you do the colorMapLookup thing on this PR?

done

> And to what extent is this just implementing our own custom framebuffer? ;)

Well it is, in a way, a extremely simple one without flipping between frame option. But I have no idea of the speed and how to use it in unity. They are mostly related to camera render, not to texture generation. We only need the texture once, not to flip it every frame. But I'm really not an expert

mikejbrown:
> But I'm really not an expert so don't quote me on that

@sboigelot more expert than I good sir... more expert than I...

PR #1353, opened 2016-09-16
Organizes the controller folder

kd7uiy:
The Controller folder is pretty complex, and this change should help to make that change a bit easier to find things

PR #1356, opened 2016-09-16
Really removes UIController

mikejbrown:
PR #1353 removed the UIController.cs file which contained only an empty
class which was not doing anything in the actual code. This was part of
a reorganisation and simplification of the folder structure for the
controllers. However, not all references to UIController were actually
tidied.

In particular, there was a prefab hanging around and a single reference to
it in WorldController. These have now been cleaned up. There are no
changes to functionality.

@kd7uiy @sboigelot Mind having a look?


sboigelot:
looking now :)

kd7uiy:
That's what was going on... Good catch!

sboigelot:
Everything seems to work for me :) good catch indeed.

mikejbrown:
thanks :+1:

PR #1359, opened 2016-09-16
Walls pumps drawn together

kd7uiy:
This is pretty controversial right now, but I want to put it before a larger group. This is how the default ship plus a few doors/pumps looks now:

capture

Note that walls only connect with other walls, they don't connect up with doors or pumps. That to me leave it looking like it's not a solid connection. The way I have it now, here's how that same scene would change

capture

The biggest concern is how the doors look with the walls. To be honest, I'm not really sure what the right solution is, but I can agree that it does look a bit odd. Still, I wanted to put this forward to see what suggestions I can get to improve it.


sboigelot:
We could go with this and make the door larger

vogonistic:
Maybe the doors shouldn't be linked to the walls. The vents are an improvement but the wall cut-off doesn't work for these kind of doors. If someone makes a thicker door, they can turn it on for that door.

kd7uiy:
If we completely take away doors connecting with walls then we end up with the bottom middle single wall section in the first image, which to me looks horrible... I guess we could change that image to something that would work better, that might help...

vogonistic:
The one tile wall between doors is what we get on master, isn't it?

kd7uiy:
Yes, that is correct. The top image is master, the bottom one is this branch.

vogonistic:
I think the solution is good, but I don't think it should be added to doors with the way they are currently designed. If someone designs a wider door, or that has a frame of some kind, we can just set the link at that points and this PR has already prepared the way for them.

kd7uiy:
There's a lot of weird test cases that I don't think we are doing right with the current build with respect to doors. Here's a few of those:

capture


With this build, this is what it looks like:

capture

I'm not saying mine is perfect, but I don't think we can completely ignore doors either... Not quite sure what is the right thing in this case, but...

kd7uiy:
Okay, in order to improve the speed on this one, I'm going to not connect up the doors. 
capture

Assets/Scripts/Controllers/Sprites/FurnitureSpriteController.cs
```diff
@@ -54,7 +54,7 @@ public Sprite GetSpriteForFurniture(string type)
     {
         Furniture proto = PrototypeManager.Furniture.Get(type);
         string spriteName = proto.GetSpriteName();
-        Sprite s = SpriteManager.GetSprite("Furniture", spriteName + (proto.LinksToNeighbour ? "_" : string.Empty));
+        Sprite s = SpriteManager.GetSprite("Furniture", spriteName + (proto.LinksToNeighbour!="" && !proto.OnlyUseDefaultSpriteName ? "_" : string.Empty));
```
vogonistic: There should be spaces around `!=`

Assets/Scripts/Models/Furniture.cs
```diff
@@ -318,7 +321,12 @@ public List<string> ReplaceableFurniture
     /// Gets a value indicating whether this furniture is next to any furniture of the same type.
     /// This is used to check what sprite to use if furniture is next to each other.
     /// </summary>
-    public bool LinksToNeighbour { get; private set; }
+    public string LinksToNeighbour { get; private set; }
+
+    /// <summary>
+    /// Defines the type of link types.
+    /// </summary>
+    public string LinkType { get; private set; }
```
vogonistic: This is never used.

Assets/Scripts/Models/Furniture.cs
```diff
@@ -24,6 +24,8 @@
 [MoonSharpUserData]
 public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextActionProvider, IBuildable
 {
+    public bool OnlyUseDefaultSpriteName { get { return !string.IsNullOrEmpty(getSpriteNameAction); } }
```
vogonistic: This is not according to the style guide

Assets/Scripts/Models/Furniture.cs
```diff
@@ -373,8 +381,8 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
             // (It will be garbage collected.)
             return null;
         }
-
-        if (obj.LinksToNeighbour)
+        
+        if (obj.LinksToNeighbour!="")
```
vogonistic: There should be spaces around `!=`

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -46,7 +46,7 @@
         <PathfindingWeight>1</PathfindingWeight>
         <Width>1</Width>
         <Height>1</Height>
-        <LinksToNeighbours>false</LinksToNeighbours>
+        <LinksToNeighbours></LinksToNeighbours>
```
vogonistic: It would be cleaner if `LinksToNeighbours` wasn't a required tag.

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -57,6 +57,8 @@
         <DeconstructJob>
             <Inventory type="Steel Plate" amount="1" />
         </DeconstructJob>
+		
+		<GetSpriteName FunctionName="Door_GetSpriteName" />
```
vogonistic: This looks like a tab, convert the tabs to 4 spaces

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -609,6 +609,14 @@ function Accumulator_GetSpriteName(furniture)
...
 end
 
+function Door_GetSpriteName(furniture)
+	if (furniture.verticalDoor) then
+	    return furniture.Type .. " Vertical_0"
+	else
+	    return furniture.Type .. "Horizontal_0"
+	end
+end
+
```
vogonistic: Was this left in by accident? It's also no consistent with space in the string.

kd7uiy:
My bad for not running style cop... Sigh. Working on it now...

Assets/StreamingAssets/Data/Furniture.xml
kd7uiy: LinksToNeighbours isn't a required tag. This was left in there in case we decide that doors should be a part of the logic. I can remove that if you'd like.

vogonistic: We can just as well add it back if we want to link doors. This way it looks like it's not optional or that maybe empty is a special value.

Assets/Scripts/Models/Furniture.cs
kd7uiy: Removed

Assets/Scripts/Models/Furniture.cs
kd7uiy: Fixed

Assets/StreamingAssets/LUA/Furniture.lua
kd7uiy: Fixed (This also explains a bug that I rarely, but occasionally, half-saw that was driving me crazy)

kd7uiy:
@vogonistic I believe this PR has now addressed all of the items you included, can you update your review of it?

vogonistic:
@kd7uiy I've changed the review to approved. Looks good!

Raketfart:
How does this look with the new "overshooting" vertical doors? I imagine it will look weird

kd7uiy:
As it stands, the change doesn't affect door/wall connections, only wall/pump connections.

Raketfart:
Ahh ok 👍

sboigelot:
@kd7uiy could you add the same support for vent in another PR later?

kd7uiy:
I'm not quite sure what you mean... All that is required to add support for this is to add the tag as below to the XML definition of the furniture:

`<LinksToNeighbours>wall</LinksToNeighbours>`

sboigelot:
@kd7uiy Yep that what I ask, just to make it and not forget it, so I can merge it soon

PR #1360, opened 2016-09-16
[Updated] Add resolution change to the options

sboigelot:
updated version:

Add an apply button to the settings menu. Nothing will change until you hit apply or save. (Save will: apply, save settings, close dialogbox)

Refactor the settings dialog box code and fix some issue (FPS counter missing, etc)

Allow the resolution change to work.

image

_resolution & fullscreen changes will only work outside of the editor: to test: build & run_


koosemose:
Can't test atm, but the Resolution change probably shouldn't be applied immediately, anything using the OnXchanged methods don't really work how one would expect a settings to work, they apply the effect immediately, and leave it as is even if you cancel out, even though the settings won't be written for next time.

sboigelot:
@koosemose I agree, I should probably add an Apply method applying all change before save. And also a button to apply before saving

koosemose:
Yeah, myself and others have been using OnClickSave basically as an apply, and you're right it should definitely be farmed out into a separate apply function, along with a button to just apply. And removing the OnXclicked/changed/whatevered functions because in most cases that's not what you need to make use of, and their presence is misleading.

sboigelot:
@koosemose I updated the PR with the changes you proposed

Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
```diff
@@ -163,22 +125,40 @@ private void LoadSetting()
...
 
-    private void CreateResDropDown()
+    /// <summary>
+    /// Create the differents option for the resolution dropdown.
+    /// </summary>
+    private void CreateResolitionDropdown()
     {
-        List<string> resolutionStrings = new List<string>();
-        foreach (Resolution r in resolutions)
+        resolutionDropdown.ClearOptions();
```
bjubes: `CreateResolitionDropdown()` is misspelled

koosemose:
Testing things out, I've realized that applying without saving feels kind of weird, the most notable thing is that with dev mode, the spawn inventory ui comes up, but as soon as it gets covered by the construction menu, it never comes back (which isn't surprising since things are just applied and not saved).

I think in most UI with a separate apply and save, apply is really just save without closing the dialog box. Perhaps we need to either have cancel actively set everything to how it was if it is desired to have it cancel out applied changes, or just have the apply button save settings without closing.

sboigelot:
@koosemose this can be done in a separate PR, this PR focus was to make the resolution dropdown works. We can improve the appy/save behavior later.

bjubes:
I agree @sboigelot in that the apply functionality can be changed later. For now this is ready to merge once you fix that spelling error i mentioned earlier.

sboigelot:
@bjubes ho sorry I didn't see your comment before, I will fix it soon when I have access to my computer!

Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
sboigelot: done :)

koosemose:
Looks good, one minor stylecop issue, I'll just do a quick followup PR to fix it. Merged

PR #1361, opened 2016-09-16
Pivot fixes

kd7uiy:
This works a bit better with graphics like the Cloning Pod, where it covers 2x3, but only occupies 2x2. It doesn't require using a custom pivotX or pivotY term to make things look correct. I believe this will make future improvements easier.


kd7uiy:
Proof that everything still looks fine.

capture

Assets/Scripts/Controllers/Input Output/MouseController.cs
```diff
@@ -591,7 +591,7 @@ private void ShowFurnitureSpriteAtTile(string furnitureType, Tile t)
 
         Furniture proto = PrototypeManager.Furniture.Get(furnitureType);
 
-        go.transform.position = new Vector3(t.X + ((proto.Width - 1) / 2f), t.Y + ((proto.Height - 1) / 2f), WorldController.Instance.cameraController.CurrentLayer);
+        go.transform.position = new Vector3(t.X + ((sr.sprite.pivot.x / sr.sprite.pixelsPerUnit) - 0.5f), t.Y + ((sr.sprite.pivot.y / sr.sprite.pixelsPerUnit) - 0.5f), WorldController.Instance.cameraController.CurrentLayer);
```
TomMalbran: This line is a bit too long, and might need to be replaced by a function?

kd7uiy: I don't know about a function, but I broke out these various transformations in to separate lines.

TomMalbran: I meant a function because you are doing the same calculation 3 times. There could be a function that receives a Tile and a Sprite and returns the Vector3. Maybe a Utility function?

kd7uiy: Fair point. I set up a system now that does that very thing. I set it up slightly different, using Vector3 addition, but it works still just as great!

Assets/Scripts/Controllers/Input Output/MouseController.cs
```diff
@@ -591,7 +591,7 @@ private void ShowFurnitureSpriteAtTile(string furnitureType, Tile t)
 
         Furniture proto = PrototypeManager.Furniture.Get(furnitureType);
 
-        go.transform.position = new Vector3(t.X + ((proto.Width - 1) / 2f), t.Y + ((proto.Height - 1) / 2f), WorldController.Instance.cameraController.CurrentLayer);
+        go.transform.position = new Vector3(t.X, t.Y, WorldController.Instance.cameraController.CurrentLayer) + ImageUtils.SpritePivotOffset(sr.sprite);
```
TomMalbran: Would it work if we use t.Z here?

Assets/Scripts/Controllers/Input Output/MouseController.cs
TomMalbran: It looks good. No code duplication is good, specially on this calculations that might change, and are easy to get wrong on copy/paste.

TomMalbran:
The code looks good now.

PR #1363, opened 2016-09-16
Sort furniture by y position when placed and new vertical door sprites

Raketfart:
This will make vertical doors be able to connect better to the north walls, by overlapping the wall. Just waiting for spirite and xml....


dusho:
so character is still in different layer and will always be in front of any furniture
will it get messy if furniture will use character sorting layer? and same order calculation as character ?

Raketfart:
Yes, character is still on top of furniture. If not, he will suddenly flip behind furniture when walking north.
But now we can do this with our vertical doors:
doors5

Raketfart:
The new vertical door and airlock overlaps the walls now. They seem to slam open and shut, but this is because of the late update described in #1383, and should probably be fixed in another PR.

Issue #1344, opened 2016-09-16
FurnitureWorkshopTest writes to disk


mikejbrown:
These lines in [FurnitureWorkshopTest](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/Assets/Editor/UnitTests/Models/FurnitureWorkshopTest.cs#L61) are bad juju:

```c#
FileStream fs = new FileStream("Workshop.xml", FileMode.Create);
TextWriter writer = new StreamWriter(fs, new UTF8Encoding());

XmlSerializer serializer = new XmlSerializer(typeof(FurnitureWorkshop));
serializer.Serialize(writer, fi);
writer.Close();

FileStream fr = new FileStream("Workshop.xml", FileMode.Open);
var dfi = (FurnitureWorkshop)serializer.Deserialize(fr);
```

These lines create a file `Workshop.xml` which is (a) not cleaned up and (b) not necessary to begin with! Unit tests should avoid touching the disk whenever possible since what is being tested here is the xml (de)serialization logic itself and not the ability to create/read files! This is easy enough to fix in this instance using a `StringWriter` instead of files. See, e.g., [SchedulerEditorTest](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/Assets/Editor/UnitTests/Models/SchedulerEditorTest.cs#L309) for an implementation of this idea.

dusho:
Will fix this when I get to my PC..

mikejbrown:
I'm tempted to add the high priority label simply because the Workshop.xml working tree spam is _really_ annoying me at the moment, but that really probably doesn't justify the label. :grinning:

sboigelot:
@dusho you should just use a MemoryStream

mikejbrown:
Fixed by #1365

Issue #1251, opened 2016-09-10
StyleCop rules and var usage


dusho:
Was just hit by [StyleCop] tag and going through the list I was wondering about some rules:

**System using directives must be placed before all other using directives**
there is already rule to keep it in alphabetical order and Organize using -> sort usings does that.. can we remove this rule and keep only alphabetical one?

**The body of the if statement must be wrapped in opening and closing curly brackets.**
so it wants me to create from 

```
if (IsWorkshop)
    workshop.Update(deltaTime);
```

code like this:

```
if (IsWorkshop)
{
    workshop.Update(deltaTime);
}

```

is there point for this? One liners are often used for events.. and with extra 2 lines it's getting longer. Usually good habit is to fit one method on one screen so you don't have to scroll around if method has some tough logic inside.. so having rules that force you stretch your methods is not a good rule, I think

**general (not a StyleCop) one is not use var**
What is the reasoning behind this? We are already using LUA that also doesn't work with explicit types. And I think everyone uses some smart editor here (MonoDevelop or VS2015) that ships with Unity. Using 'var' is faster to write code and lines are usually shorted and don't need splitting.
`foreach (var toConsume in flaggedForTaking)`
vs.
`foreach (KeyValuePair<Tile, int> toConsume in flaggedForTaking)`
or
`var flaggedForTaking = new List<KeyValuePair<Tile, int>>();`
vs.
`List<KeyValuePair<Tile, int>> flaggedForTaking = new List<KeyValuePair<Tile, int>>();`

UnknownSoldier1:
AFAIK the reason why StyleCop doesn't like one liners is that if you want to add more code to the if statement, it's a lot faster and you don't have to add brackets first.

TomMalbran:
And if you add more code inside the if and you forget to add the brackets, you can end up with a bug that might be difficult to track.

dusho:
hm.. ok. That makes sense. Never seen bug like that, but I guess some people may encounter it..
What about **usings** and **var** rule..

sboigelot:
The way I see the **var** rule, it's to be stay beginner friendly. It's easier to understand for someone that doesn't code all the time and doesn't know that the var will be a typed variable once compiled.

koosemose:
usings I agree with you on, having both one seems a little excessive to me.

[...]

PR #1365, opened 2016-09-16
[small] FurnitureWorkshopTest not de/serializing to disk anymore

dusho:
using StringWriter/Reader instead of TextWriter/Reader (.xml only in memory)
fix for #1344

### DIFF
--- a/Assets/Editor/UnitTests/Models/FurnitureWorkshopTest.cs
+++ b/Assets/Editor/UnitTests/Models/FurnitureWorkshopTest.cs
@@ -7,7 +7,6 @@
 // ====================================================
 #endregion
 using System.IO;
-using System.Text;
 using System.Xml.Serialization;
 using NUnit.Framework;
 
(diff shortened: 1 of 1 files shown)


Assets/Editor/UnitTests/Models/FurnitureWorkshopTest.cs
```diff
@@ -57,18 +56,22 @@ public void TestSerialization()
...
 
-        FileStream fr = new FileStream("Workshop.xml", FileMode.Open);
-        var dfi = (FurnitureWorkshop)serializer.Deserialize(fr);
+        StringReader sr = new StringReader(writer.ToString());
 
+        // if you want to dump file to disk for visual check, uncomment this
+        ////File.WriteAllText("Workshop.xml", writer.ToString());
+        
+        // deserialize
+        var dfi = (FurnitureWorkshop)serializer.Deserialize(sr);
```
mikejbrown: The only tweak I would make is to eliminate this use of var to be in keeping with the style guide. Otherwise this is merge ready.

In fact I'll just make the change myself now...

dusho: same line explicitly contains the type it uses.. so using the var here is better for eyes..
I think people in #1251 generally agreed on 'var' usage if it's clearly visible what it is.
Even you gave 'like' to Grenkin's post when var was used gently :)

mikejbrown: something something consistency something something hobgolbin... :P

Issue #1140, opened 2016-09-04
[Discussion] Switch from Managers to Services and move to a new folder


TomMalbran:
If we stay with the current folder structure (Controllers, Models, Utilities, UI) that tries to mimic the Controller Model architecture, we should think about a new layer that goes between a Controller and a Model. 

In another project I work with I have an extra layers between the controller and the model. The controller is an interface between the ui/view and the next layer, and the model just models an entity. So there is a missing piece that can manage models and/or provide services to the controller, which in other architectures is called a Service. Our manager are mostly Services in that way.

I think that we should make a Services folder, move the managers there, and change the class names to ___Service. We could go for manager, but I think that Service is a better name.

That architecture also specifies a Repository layer which holds the Models, but for now I am not sure if we need to go deeper.

sboigelot:
Service is a better name for me too.

akomm:
The manages are actually a mix of repositories and services. 

- Repositories: Hold a set of models. Query API if needed, add/remove API.
- Services: implementation of reusable code/logic to operate on a model or a set on models, should actually not hold any state information.
- Controllers: handle (user) interactions. Know what to do, but not HOW to do (services job)

PR #1366, opened 2016-09-16
Moved non-Controller files into Models

TomMalbran:
PR #1353 organized the Controllers folder, but in Controller we had several classes that were not Controllers. This PR moves them to the Models folder. And yes, the Models folder is really big, but that is because our Models are also Services and Repositories, and we should probably split the Models folder in those groups. For now, the Controller folder only contains Controllers.

Also, I think that we shouldn't have spaces in in file/directory names since it makes it use quotes in the console and can be an issue for contributors that don't know it.


kd7uiy:
As long as you didn't move anything to the root /Models folder, I'm okay with that;-)

I'm not sure if Camera, SpriteManager, KeyboardManager, and maybe even Settings belong in the Model folder.

TomMalbran:
CameraData is fine in Model since it holds the Camera Data. But I agree with all the others. The issue is that they also didn't belong where they were before. They should be in Services, as long as many of the other managers, but we don't have that yet. There hasn't been much discussion in #1140. I have a branch which moves some of the files, but haven't made a PR yet.

kd7uiy:
Okay, I didn't really look at the CameraData, I see what you mean.

mikejbrown:
:+1:

PR #1369, opened 2016-09-17
Moddable Headlines

TomMalbran:
Moved the Headlines into the PrototypeManager. I had to remove the minInterval and maxInterval from the xml since it wouldn't work with the PrototypeManager. Maybe each event could have its own min and max intervals or use the defaults. Or there can be Headline Sources which organize the Headlines into categories each with an interval and such. Maybe for later.


mikejbrown:
Courtesy ping @kd7uiy (author of headlines #1033)

kd7uiy:
The only issue I have is that the intervals are now hard coded, where they could previously be defined in XML. I think I would have the type be something like "Funny", "Serious", etc, but I guess that kind of stuff will come with time. I think I see where it came from, but...

TomMalbran:
I think that the interval was not defined in the best place, but we can move it to xml if each headline had its own interval or if we go with a sort of HeadlineSource. And it might not be hard to add in a initial form.

```xml
<HealineSources>
  <HeadlineSource type="quill_press" minInterval="5" maxInterval="10">
    <Name>Quill Press</Name>
  </HeadlineSource>
</HeadlineSources>

<Headlines>
  <Headline type="CEO" source="quill_press">The CEO of Quillcorp, has announced that it main and only shareholder is still the main and only shareholder, Quill18.</Headline>
  <Headline type="Chairs" source="quill_press">Notice: Quillcorp has placed an embargo on "Chairs", any Quillcorp Basic Utility Station found in possession of this illegal contraband will be fined 2 million Quillbucks.</Headline>
</Headlines>
```

The type has to be unique to each Headline, It can also be a short name or a title, or even just a number (as a string). Is there just because the PrototypeMap requires it, and is easier to add it to the headlines than to make a special case.

kd7uiy:
So, type is just a short version of the full text? Interesting... Could you have the type be the inner text? Just haven't looked much in to it, so...

TomMalbran:
Type is just a way to identify the Headline if for some reason wanted a particular Headline. In this case, it probably won't be used, but it is required as a key to store the prototype in the prototype map.

I think the inner text might work. The key just needs to be a string and unique. I will try that.

kd7uiy:
Actually, we should be doing localization. We might want to use the type as the localization key or something... Hmmm...

koosemose:
I'm not sure Prototypes is a proper conceptual fit for headlines, and even though it may be more convenient on the code side to access them via the prototype manager, having to have "type" in the xml is somewhat misleading to anyone editing only the xml (i.e. a modder). Also I think having to work within the framework provided by prototypes will be limiting to future expansion, or require things added to all prototypes that won't make sense or be used by anything but headlines.

TomMalbran:
Right now it might seem that way. But I can see this system expand a lot. For example each event could have a chance to appear so that not every headline appears with the same frequency. And I think that a headline could have a cooldown so that after it appears it needs to pass certain amount of time before it can appear again. A headline could appear only under certain conditions, which could be done with a combination of parameters and lua functions. A headline could even be more than 1 text, were each text would be a variation over the same thing, and one is chosen at random, or in some sort of sequence, to give more overall variation. A headline could be part of a Source, and when the headline appears it will also show the source. There can be a headline that could create more than 1 headline event, in a way that it could tell a story. Once we start adding things like this we would need to store the data in an Headline object, and we can even get to a point where we would need to clone the headlines.

The Prototype Manager is basically multiple Dictionaries of object that were added manually or loaded from an XML. It doesn't do anything fancy, but it works for everything that has multiple instances of an object, and you need to load and store. I don't think that we will need to add more to it, in any case we might just need to move some of the xml loading functionality to other utilities to be used in other places.

I agree that right now the type makes no sense, and maybe it should be called name. For now we can even remove it and use the text as the type. The only requirement is to have a Type getter that returns a string, that string doesn't need to be the type it can be anything else that is unique to each object. But at some point it will be better if the headlines had a key so that a mod or some event system could tell the HeadlineGenerator to place a particular headline.

koosemose:
My main issue is just having to define what is essentially the name of the headline under as "type". If that can be straightened out, I'm fine with this.... as long as it doesn't prevent future progress (such as something functioning as a variable in a headline)

TomMalbran:
As I said, it can be name in both the xml and Headlines, and then have Type just returns the name. It can also be string itself the type. Maybe later we can change Type to Key in the IPrototypable, but I don't think that it is needed.

sboigelot:
An additional point to separate the text and the metadata is localisation.

TomMalbran:
So I went ahead and removed the type in the Headlines. We can add it later when it is required.

I did a couple of other small changes.

Assets/Editor/UnitTests/Models/SchedulerEditorTest.cs
```diff
@@ -38,20 +38,18 @@ public void Init()
...
         {
             new FunctionsManager();
-            FunctionsManager.ScheduledEvent.LoadScript(LuaFunctionString, "ScheduledEvent");
         }
 
+        FunctionsManager.ScheduledEvent.LoadScript(LuaFunctionString, "ScheduledEvent");
+
         if (PrototypeManager.SchedulerEvent == null)
         {
             new PrototypeManager();
```
sboigelot: This seems weird to me. Shouldn't it be:

```
         if (PrototypeManager == null)
          {
              new PrototypeManager();
```

Assets/Scripts/Models/Events/HeadlineGenerator.cs
```diff
@@ -40,7 +33,7 @@ public void AddHeadline(string headline, bool displayImmediately = true, bool ke
     {
         if (keepInQueue)
         {
-            headlines.Add(headline);
+            PrototypeManager.Headline.Add(new Headline(headline));
```
sboigelot: I'm not sure about the pattern here. Should we add (and or modify) things to the Prototype? Aren't the Prototype suppose to be a read only in memory version of what is in the xml definition files?

TomMalbran: This is adding a new "Prototype" to the PrototypeMap. You can change each prototype, since is just an instance of a class.

Assets/Editor/UnitTests/Models/SchedulerEditorTest.cs
TomMalbran: Unity doesn't like your suggestion. But we could just instantiate it without checking, since the constructor only creates instances of PrototypeMap, and on the next line it loads it.

sboigelot: Haaa sorry, this is not a stored variable but a static access, this is for another PR but we should really get rid of this for a singleton.

It's good the way it is for now.

Assets/Scripts/Models/Events/HeadlineGenerator.cs
TomMalbran: Since the headline added here will run more than 1 time it made sense to add it to the prototype. But if it should just run 1 time, then HeadlineGenerator should have a Queue of Headlines and grab from them.

sboigelot: Because we can, doesn't mean we should. I don't think it's the best pattern, but I may be wrong.

For me prototype shouldn't change after load. If we want a variance, we should store the data in another form.

Assets/Editor/UnitTests/Models/SchedulerEditorTest.cs
TomMalbran: We can totally remove the if check though.

Yes I agree with you. Static things caused issues already. I am still thinking of a global static class that holds all the current static classes. Just to make them easy to access.

Assets/Scripts/Models/Events/HeadlineGenerator.cs
TomMalbran: This isn't changing any prototype, it is just adding one. We are already doing that in other places, and I think that it should be ok to add more prototypes at run time.

sboigelot: The only place we do it for now is in the ScheduleEvents and I didn't like that already :p

I won't block this PR for that but, we should probably open an issue to discuss about it. Just defining the prototypes would be enough for me.

I wanted to do the same for the trading request but restrained myself until now.

sboigelot: Created the issue

TomMalbran: A prototype is only an instance of a class that you will want to copy so that it preserves the default values and then change the rest. I see no problem in changing it at run time, when we need to change a default for all the new instances created from a prototype or to add new prototypes that we will later copy, or even remove ones that we don't want anymore.

The issue here is that technically we don't need to clone the prototypes, since at the moment they don't hold data that will change at some point. But that will change when we expand the system.

The other issue, is that this function did 2 totally different things depending on a parameter. It should either always add the Headline to prototypes and show it now, or never add it and show it now or show it at a later time. Or have 2 functions.

sboigelot: @TomMalbran let's continue the discussion in #1394 so everyone can see and pitch in

TomMalbran: We can still discuss if we need to change this function from my comment in the third paragraph.

sboigelot: Looking at the code now

TomMalbran: BTW, my change didn't modified the original functionality.

sboigelot: This method isn't used anyway, except in the unit test. I vote to delete unused code.

TomMalbran: I know, but it looks like an useful thing. Do you prefer removing it, or use this method to show a Headline only once?

sboigelot: We can keep the shown only once for now. It will be useful indeed. Or split in 2 method, your call, both is ok for me

TomMalbran: I will make it show once. If you want it to show more than one, you can just add it to the xml or add the prototype.

TomMalbran: done

sboigelot:
Tested, still work, no stylecop issues.

Issue #1328, opened 2016-09-15
Planned Airlock Functionality


koosemose:
In the most basic form an airlock should prevent air from being lost as crew exit the base. This is currently possible in an extremely simplistic form (two airlock doors, and a pump keeping the room a vacuum, with the outer door pressure locked). 

For a more properly functional airlock, it should function with both doors pressure locked, never opening a door with the rooms on either side having different pressure, and preventing any rushes of air. So when the inner door needs opened, it will pump the air inside up to 1 atm pressure (assuming that is the pressure inside the base), the crewmember enters, and then when they need to exit the base it should pump the air back out. 

Optimum functionality (most likely for a later version than 0.1) should have some sort of queuing system to prevent crew from fighting over rather air needs pumped in or out (creating a deadlock), and just for more generally efficient use of the airlock.

The base things needed for a fully functional airlock is a door that can be pressure locked to only open when the pressure is equal on both sides and doesn't leak air, a pump that can pump air in or out of the airlock, and some form of room behavior that can be used to designate a room to function as an airlock, and ensure that it has access to the required items ( 2 airlock doors and a pump at a minimum ), and can be used to handle the interactions between the airlock door and the pump ( primarily activating the pump in such a way to bring the pressure to the proper level to allow the door to open ).

The airlock door is in and can be pressure locked, and the pump is in and functional as well. All that is needed is a Room Behavior of some form. 

As the room behavior/designation is what still needs implemented that will be my primary focus on discussion. 

I think the code for controlling the airlock door should live in the Room behaviors, rather than in the door and pump, as that makes the most logical sense, though I am not absolutely certain at this time if all of the relevant code can function solely from the Room Behavior side. The specific details of implementation will depend on what exactly will work. 

For a first pass for 0.1 Room Behaviors will most likely be assigned purely on a room by room basis, with the possibility for rooms behaviors being able to be designated to a limited area of the room to be investigated in the future. 

Room Behaviors should probably follow a similar prototype system as most of the rest of our system is, with similar Lua hooks, a dictionary of Furniture, keyed by what purpose it serves, and Parameters for any other arbitrary information that needs to either be available to Lua or stored.

What other functionality might be needed for RoomBehaviors (and possibly an alternative name) to implement an airlock, and potentially be useful for other RoomBehaviors than airlock, though initial development will likely focus on what's needed for airlocks, with other things either implemented if they're "adjacent" to airlock functionality, as possible, or just delayed for future refinements.

[...]

PR #1370, opened 2016-09-17
Hotfix for GameEvent double up

mikejbrown:
The two file reorganizations in #1366 and #1328 conflicted, resulting in a
double up of GameEvent.cs leading to a compile error. Somehow it didn't
show as a merge conflict so nothing was visible when testing out the PRs.

I have removed the double and merged the Models/Event directory into
Models/Events.


mikejbrown:
@TomMalbran thanks for testing

PR #1372, opened 2016-09-17
Allow Travis script to run locally in a form useful for a pre-commit or pre-push hook

mikejbrown:
Changes the Travis CI script to allow for it to run locally. This makes it suitable for a pre-commit or pre-push hook. At the moment I am using it as a pre-push hook, which guarantees that I'm not going to push anything which will make Travis builds fail. To use as a pre-commit or pre-push hook simply copy or symlink to the relevant place in .git/hooks/. It is also a faster way to run tests than firing up the Unity editor or IDE.

Running the script locally can also help with PR testing since the remote Travis doesn't always keep up with changes on upstream master. It makes it a shell one liner to checkout a PR branch, merge upstream/master (assuming no merge conflicts) and run all tests.

You can configure your local Unity path using the unityPath env variable. Otherwise an OS based default is assumed. **NOTE**: currently I've only put in the default unityPath of Msys (how git bash for Windows shows up in `uname`). I don't know default paths on other platforms. There is an opportunity here for people on other OSes to contribute!

**NOTE**: Local build is known to fail if you already have the Unity editor running. (Unity refuses to run in batch mode in that case.) The may be a better way of handling this case.

Just having a bit of fun here. :grinning: Thoughts?


mikejbrown:
Is there any way to run stylecop from the script as well?

bjubes:
I don't understand why you can't just run `sh Scripts/test.sh` on your local machine? I fear that making an identical copy of test.sh will cause deviation and then we will end up with different copies of the same script.

sboigelot:
why not run the tests in unity?

bjubes:
well i see running them from command line due to increased speed, but running the travis version is fine, it just will echo `travis_fold:whatever` occasionally which is to make the travis log fold in sections, but is harmless on a local machine

mikejbrown:
Two reasons for a separate script:
- test.sh uses the unity executable assumed to exist in the os x testing envirenment on travis. The new script has a configurable path instead.
- test.sh log spams and also doesn't clean up some files it leaves behind, which is fine when running on travis but spams the working tree with changes when run locally.

I've been thinking of combining the two scripts using some env variables to change the behavior appropriately for running locally/remotely, but this will have to wait till later.

mikejbrown:
I've now combined the scripts. I've tested all the logic locally but not on a Travis server yet.

mikejbrown:
```sh
$ ./Scripts/test.sh --travis
```
That confirms that the Travis CI build is being run with the new version of the test script.

bjubes:
when running `./Scripts/test.sh` on mac i get:
```
uname: illegal option -- o
usage: uname [-amnprsv]
Don't know the path to your local Unity! Bailing.
```
Can you make it so that it defaults to the mac OSX unity path since most people using bash to build a test are on macs? That way this PR has the same base functionality as the original test.sh script.
The `--travis` argument is a much better implementation than before. Keep it up!

mikejbrown:
@bjubes What does `uname -s` return on mac? That might be a better test. I'll change the default as well.

bjubes:
@mikejbrown `Darwin`

mikejbrown:
Yeah this is fine to merge as far as I'm concerned. Can make improvements to the OS recognition logic later if needed.

PR #1376, opened 2016-09-17
Implemented RequestItem for jobs

vogonistic:
RequestedItem is a more lightweight and specific way to describe requirements, including minimum needed. Actual inventory in the job is still carried in Inventory instances.

This doesn't make full use of the RequestedItem, but it allows us to implement it later. 

This functionality was extracted from #1219


Assets/Scripts/Models/Job/RequestedItem.cs
```diff
@@ -0,0 +1,84 @@
...
+        public bool DesiresMore(Inventory inventory)
+        {
+            return AmountDesired(inventory) > 0;
+        }
+
+        public int AmountDesired(Inventory inventory = null)
+        {
+            if (inventory == null || inventory.Type != Type)
+            {
+                return MaxAmountRequested;
```
mikejbrown: I like having separate `AmountDesired` and `AmountNeeded` methods. Very good. I'm just a little confused about the semantics here. If I call `requestedItem.AmountDesired( inventoryOfATotallyIrrelevantType )` it returns the max amount? I would expect it to return zero because it doesn't request any items of that type.

Assets/Scripts/Models/Job/RequestedItem.cs
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            return MaxAmountRequested - inventory.StackSize;
+        }
+
+        public int AmountNeeded(Inventory inventory = null)
+        {
+            if (inventory == null || inventory.Type != Type)
+            {
+                return MinAmountRequested;
```
mikejbrown: Similarly to the comment above.

Assets/Scripts/Models/Job/RequestedItem.cs
TomMalbran: I think that the idea behind min and max will be that a Metal Smelter can say: "Hey, I want 5 Raw Iron to work but you bring me upto 50 since I will use it eventually". The Character would pick a job, and ask: "How many items do you need?", and the answer will be 5 Raw Iron. He goes to look for 5 Raw Iron and finds a stack of 20. So he can then ask: "Are you cool with 20 Raw Iron", and the Job would say, "Yes, he can hold upto 50". So the Character replies: "Cool, I'll take al 20 and we can save on trips later on". Not sure if this was implemented.

Saying that I think that Desires more is a bit confusing, yes it can take more, but it really desires the min amount to start working. Maybe something like CanTakeMore or CanHoldMore, and the other function can be MaxAmountItCanHold or something like that.

vogonistic: @mikejbrown @TomMalbran I'm open to better names on the variables and methods.

I understand that passing in an irrelevant type seems confusing. The way I was thinking about the request you are pointing out is that you aren't asking how many more items similar to what the character carries that the job wants. Instead you are asking a specific requested material, like steel plate, how many are desired. Passing in the inventory allows it to account for what the character is carrying.

If the job wants 20 steel plate and the character is carrying iron ore, the desired amount of steel plate is still 20. If the character is carrying 10 steel plate, then the desired amount is 10.

@TomMalbran Is correct about what the variables are representing. If the min and max are set to 5 and 50 respectively, it needs 5 but desires 50. But as I said. I'm not locked in on the names. I'm going to mark this as `WIP` until we figure out a way to clarify this.

TomMalbran: I cant think of better names. But I now get what @mikejbrown was asking, which was totally different to what I answered. So maybe the names are not the issue. I would still add comments, even if we change the names.

@vogonistic One think that I was thinking for your fsm. In the example, the character picks 20 items, but the job still desires 50. Maybe the character could check if there are more inventory close to the one it found, lets say up to 10 tiles away. If there is, it will also pick that. At the second pick, we can do the same, but we should check that the total distance from the first inventory to the last is less than some amount.

vogonistic: @TomMalbran  The FSM hauler will keep looking for stacks until it has the max it can carry or there is nothing left to get. It doesn't do a distance check right now, but that would be a good improvement. 

I'll add comments to describe how it works.

mikejbrown: @vogonistic @TomMalbran I get what min and max represent. So these functions really return "amount of _my_ type desired(needed) _after_ a possible contribution from the passed inventory has been taken into account." Not sure what a good concise name would be. Perhaps doc comments would be the most helpful here.

Assets/Scripts/Models/Job/RequestedItem.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        public int AmountDesired(Inventory inventory = null)
+        {
+            if (inventory == null || inventory.Type != Type)
+            {
+                return MaxAmountRequested;
+            }
+
+            return MaxAmountRequested - inventory.StackSize;
```
mikejbrown: Also there should probably be a `Mathf.Max(MaxAmountRequested - inventory.StackSize, 0)` here as well.

koosemose:
Sorry, you're conflicted.

Assets/Scripts/Models/Job/RequestedItem.cs
sboigelot: @mikejbrown don't use Mathf for int, use Math

koosemose: Unless there's a nonobvious way to access it, there isn't a Math available to unity, only Mathf. But, if given ints Mathf returns ints.

sboigelot: It's a system class, even in .net 2 
https://msdn.microsoft.com/en-us/library/system.math(v=vs.80).aspx

I can access it without problem even in a unity projet
no obvious way required

image

koosemose: Ahh, every time I've thought to try it I guess it was in a class that wasn't including System. Is there an advantage to pulling System in when it's not already being used and Mathf uses and returns ints anyways?

dusho:
still feels a bit weird that maxAmountRequested (parameter for constructor) drives JobComplete. But maybe just comment to parameter would help there? amount of inventory needed to complete the job
works nice in most cases and also fixes funny behavior that is in #1413 
BUT
if you specify maxAmountRequested bigger than actual maxStackSize of item, job never completes
you can try it by changing in furniture.lua for function PowerGenerator_UpdateAction
`local itemsDesired = {RequestedItem.__new("Uranium", 1, 5)}`
to
`local itemsDesired = {RequestedItem.__new("Uranium", 30)}`
then power generator will never start even though you're giving it all the uranium

mikejbrown:
[Conflicts]

Assets/Scripts/Models/Job/RequestedItem.cs
sboigelot: The main advantage is readability, you directly know it's int. I don't know about performance though, it probably change nothing else.

sboigelot: From the unity forum:

> I highly doubt your 'benchmarks' are accurate, regardless. From what I can tell from the assemblies, Mathf is just a wrapper using casts for the respective System.Math functions.
> 
> So, if anything, Mathf will generally perform slightly slower.

koosemose:
I'm removing the WIP from this, since as far as I can tell, things work as they should, and I'd like to get this code into master as soon as possible, and allow others to do any enhancements to it from there, since vog will most likely be unable to keep up with getting this to perfect working condition.

koosemose:
This seems to be working as intended now, and no significant bugs have been found, further work can take place in a future PR. Merged

PR #1378, opened 2016-09-17
Tweak airlock door and pump

koosemose:
Gave the airlock door a little bit more tolerances when pressure locked ( this may allow a small amount of atmosphere to be lost when exiting the station, but reduces the likelihood of not letting you leave at all. This may need to be tweaked further in the future.)

Prevented air pump from overfilling a room. 

Now also prevent OxygenGenerator from overfilling a room.

Added GetTotalGas, and made pressureLimited versions of several gas moving functions to Room.cs to support the tweaks to the airpump.

Also included is a fix to GetGasFraction (it was causing wildly incorrect percentages to be shown for gas percentages) as it makes use of the GetTotalGas method that is added by this PR.


koosemose:
Actually I think i may have a slightly better approach... [WIP]ing this while I work it out.

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -72,7 +72,7 @@ function OnUpdate_AirlockDoor( furniture, deltaTime )
                 adjacentRooms[count] = tile.Room
             end
         end
-        if(ModUtils.Round(adjacentRooms[1].GetTotalGasPressure(),3) == ModUtils.Round(adjacentRooms[2].GetTotalGasPressure(),3)) then
+        if(math.abs(ModUtils.Round(adjacentRooms[1].GetTotalGasPressure(),3) - ModUtils.Round(adjacentRooms[2].GetTotalGasPressure(),3)) < 0.005 ) then
```
bjubes: should 0.005 be variable/using the mathUtils class for float comparison?

koosemose: The epsilon for for float comparison is drastically smaller than this, and I'm not entirely sure of a proper variable name for this... maybe tolerance?

koosemose:
Conflicts resolved, but there's a bug in master that causes rooms to become room 0 when room enclosing furniture is placed, that got pulled into this while resolving conflicts, which makes this hard to test, since anything to do with pressure doesn't really do anything in room 0.

mikejbrown:
Is there an issue for the bug on master?

koosemose:
I just created it. #1405

koosemose:
Ok, everything should be fixed up in this, and I think any other bugs that needed to be fixed to actually properly test this have been fixed, other than #1411, which needs to be merged before a vent or airpump can replace a wall again.

koosemose:
Minor change, self-merging.

PR #1381, opened 2016-09-17
Closes #1349 , Fixes a GUI problem 

### LINKED ISSUE #1349: Right Click Furniture UI Text-length
So right clicking on furniture has a problem when the name of the furniture object name is too large the text is clipped out of the ContextMenuItem(Clone).
It might have something to do with "vertical layout group component" of the parent of  "ContextMenuItem(Clone)"Game Object

eg. as shown below when the user has right clicked the oxygen generator the deconstruct button is incomplete saying "Deconstruct Oxygen " where it should say "Deconstruct Oxygen Generator"

image

TheAvaren:
Fixes the GUI problem that cause's the name of a furniture object not to show full name.

+Added LayoutElement to ContextMenu.prefab
+Added LayoutElement to ContextMenuItem.prefab

Addition of LayoutElement fixes Issue #1349.


sboigelot:
Either this doesn't work, either my git is confused because you are making a PR from your master. Is this intended?

TheAvaren:
Sorry, at the time i made this i didn't have the correct workflow and it was kinda messy but this should still work ill do some more testing.

PR #1382, opened 2016-09-17
Fix Utilities Context Menu

koosemose:
This applies the same fix used in #1333 to Utilities.


Assets/Scripts/Models/Buildable/Utility.cs
```diff
@@ -724,9 +726,9 @@ private void ClearJobs()
         }
     }
 
-    private void InvokeContextMenuLuaAction(string luaFunction, Character character)
+    private void InvokeContextMenuLuaAction(ContextMenuAction action, Character character)
     {
-        FunctionsManager.Utility.Call(luaFunction, this, character);
+        FunctionsManager.Furniture.Call(action.Parameter, this, character);
```
TomMalbran: It should call Utility

TomMalbran:
If we want to use a lambda, then we can do:


```
yield return CreateContextMenuAction(contextMenuLuaAction);

private ContextMenuAction CreateContextMenuAction(ContextMenuLuaAction contextMenuLuaAction)
{
    return new ContextMenuAction
    {
        Text = contextMenuLuaAction.Text,
        RequireCharacterSelected = contextMenuLuaAction.RequireCharacterSelected,
        return (cma, c) => InvokeContextMenuLuaAction(contextMenuLuaAction.LuaFunction, c)
    }
}
```
And same can be done in Furniture.

If this was broken, is the jobs context menu also broken? We usually only have 1 job so we haven't tested it, but it looks like it might be broken.

Assets/Scripts/Models/Buildable/Utility.cs
koosemose: *grumblegrumble* I really shouldn't try to fix things when I'm tired...

koosemose:
I'm honestly not sure, I was just applying the same fix (as is quite obvious from my forgetting to change it from the copy paste) as I ran across it while doing something else.

Assets/Scripts/Models/Buildable/Utility.cs
koosemose: Fixed now

TomMalbran:
Ok. I could try that later.

sboigelot:
@koosemose please update this branch with master to have the context menu actually working with utility. This is not testable for now.

koosemose:
Fixed the spelling of require that was causing issues with context menu actually being used, pulled in the changes to properly hook the context menus, and added context menus to Power Cables to test that context menu's work.

PR #1388, opened 2016-09-17
Removed old loading code

TomMalbran:
@sboigelot Here it is. I was going to add it to another PR, but I broke the loading somehow and I am looking why. So it is in its own PR now.


sboigelot:
It still works, Thanks for the quick reaction.

PR #1390, opened 2016-09-17
[Small, Fix] GUI Text scaling fix.

### LINKED ISSUE #1349: Right Click Furniture UI Text-length

TheAvaren:
So this fixes #1349,
It does so because i added a element layout group onto:
ContextMenu.prefab,
ContextMenuItem.prefab.
and changing some other values of the other components.

This should be no problem to merge :)
also don't worry about deletion of 
Assets/Resources/Prefabs/ContextMenuItem.prefab.meta and
Assets/Resources/Prefabs/ContextMenuItem.prefab

as you can see they were accidental copies of the original as they were in Resources/Prefabs not the official folder which is Resources/Prefab


sboigelot:
The context menu is now too tall when the text fits on one line
image

TheAvaren:
It is currently static and doesn't adapt to the size of the text :/ i'll see what i can do, i only took into the account of non-intractable furniture.

PR #1393, opened 2016-09-17
Better deconstruct

kd7uiy:
This will force a deconstruct to occur, instead of happening immediately. 

gif

### DIFF
--- a/Assets/Scripts/Models/Buildable/Furniture.cs
+++ b/Assets/Scripts/Models/Buildable/Furniture.cs
@@ -59,6 +59,9 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
 
     private bool isOperating;
 
+    private float deconstructJobTime;
+    private List<Inventory> deconstructInventory;
+
     // did we have power in the last update?
     private bool prevUpdatePowerOn;
 
@@ -611,6 +614,9 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
                 case "BuildingJob":
                     ReadXmlBuildingJob(reader);
                     break;
+                case "DeconstructJob":
+                    ReadXmlDeconstructJob(reader);
+                    break;
                 case "CanBeBuiltOn":
                     tileTypeBuildPermissions.Add(reader.GetAttribute("tileType"));
                     break;
@@ -730,6 +736,30 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
     }
 
     /// <summary>
+    /// Reads the XML building job.
+    /// </summary>
+    /// <param name="reader">The XML reader to read from.</param>
+    public void ReadXmlDeconstructJob(XmlReader reader)
+    {
+        deconstructJobTime = 0;
+        float.TryParse(reader.GetAttribute("jobTime"),out deconstructJobTime);
+        deconstructInventory = new List<Inventory>();
+        XmlReader inventoryReader = reader.ReadSubtree();
+
+        while (inventoryReader.Read())
+        {
+            if (inventoryReader.Name == "Inventory")
+            {
+                // Found an inventory requirement, so add it to the list!
+                deconstructInventory.Add(new Inventory(
+                    inventoryReader.GetAttribute("type"),
+                    int.Parse(inventoryReader.GetAttribute("amount")),
+                    0));
+            }
+        }
+    }
+
+    /// <summary>
     /// Accepts for storage.
     /// </summary>
     /// <returns>A list of Inventory which the Furniture accepts for storage.</returns>
@@ -756,8 +786,25 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
     /// <summary>
     /// Deconstructs the furniture.
     /// </summary>
-    public void Deconstruct()
+    public void SetDeconstructJob()
     {
+        Jobs.CancelAll();
+
+        Job job = new Job(
+            Tile,
+            Type,
+            (inJob)=>Deconstruct(),
+            deconstructJobTime,
+            null,
+            Job.JobPriority.High);
+        job.JobDescription = "job_deconstruct_" + Type + "_desc";
+        job.adjacent = true;
+
+        World.Current.jobQueue.Enqueue( job);
+    }
+
+    public void Deconstruct()
+    { 
         int x = Tile.X;
         int y = Tile.Y;
         int fwidth = 1;
@@ -770,6 +817,14 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
             fheight = furniture.Height;
             linksToNeighbour = furniture.LinksToNeighbour;
             furniture.Jobs.CancelAll();
+        }
+
+        if (deconstructInventory != null)
+        {
+            foreach (Inventory inv in deconstructInventory)
+            {
+                World.Current.inventoryManager.PlaceInventory(Tile, inv);
+            }
         }
 
         // We call lua to decostruct
@@ -941,7 +996,7 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
             {
                 Text = "Deconstruct " + Name,
                 RequireCharacterSelected = false,
-                Action = (ca, c) => Deconstruct()
+                Action = (ca, c) => SetDeconstructJob()
             };
         }
 
--- a/Assets/Scripts/Controllers/InputOutput/BuildModeController.cs
+++ b/Assets/Scripts/Controllers/InputOutput/BuildModeController.cs
@@ -99,7 +99,7 @@ public class BuildModeController
                 // TODO Possibly return resources. Will the Deconstruct() method handle that? If so what will happen if resources drop ontop of new non-passable structure.
                 if (tile.Furniture != null)
                 {
-                    tile.Furniture.Deconstruct();
+                    tile.Furniture.SetDeconstructJob();
                 }
 
                 // Create a job for it to be build
@@ -271,7 +271,7 @@ public class BuildModeController
                     }
                 }
 
-                tile.Furniture.Deconstruct();
+                tile.Furniture.SetDeconstructJob();
             }
             else if (tile.PendingBuildJob != null)
             {
--- a/Assets/Scripts/Controllers/Sprites/JobSpriteController.cs
+++ b/Assets/Scripts/Controllers/Sprites/JobSpriteController.cs
@@ -103,6 +103,10 @@ public class JobSpriteController : BaseSpriteController<Job>
             job_go.transform.position = new Vector3(job.tile.X, job.tile.Y, job.tile.Z);
             sr.sprite = SpriteManager.GetSprite("Tile", "Solid");
         }
+        else if (job.JobDescription.Contains("deconstruct"))
+        {
+            //For now, do nothing
+        }
         else
         {
             // This is a normal furniture job.
--- a/Assets/Scripts/Models/Character/Character.cs
+++ b/Assets/Scripts/Models/Character/Character.cs
@@ -657,7 +657,7 @@ public class Character : IXmlSerializable, ISelectable, IContextActionProvider
         }
         else
         {
-            pathAStar = new Path_AStar(World.Current, CurrTile, DestTile);
+            pathAStar = new Path_AStar(World.Current, CurrTile, DestTile, null, 0, false, false,MyJob.adjacent);
         }
 
         Profiler.EndSample();
--- a/Assets/Scripts/Pathfinding/Path_AStar.cs
+++ b/Assets/Scripts/Pathfinding/Path_AStar.cs
@@ -25,7 +25,7 @@ public class Path_AStar
         this.path = path;
     }
 
-    public Path_AStar(World world, Tile tileStart, Tile tileEnd, string type = null, int desiredAmount = 0, bool canTakeFromStockpile = false, bool lookingForFurn = false)
+    public Path_AStar(World world, Tile tileStart, Tile tileEnd, string type = null, int desiredAmount = 0, bool canTakeFromStockpile = false, bool lookingForFurn = false, bool adjacentOkay=false)
     {
         // if tileEnd is null, then we are simply scanning for the nearest Type.
         // We can do this by ignoring the heuristic component of AStar, which basically
@@ -59,12 +59,20 @@ public class Path_AStar
         // if tileEnd is null, then we are simply looking for an inventory object
         // so just set goal to null.
         Path_Node<Tile> goal = null;
+        List<Path_Node<Tile>> neighbors = new List<Path_Node<Tile>>();
         if (tileEnd != null)
         {
             if (nodes.ContainsKey(tileEnd) == false)
             {
                 Debug.ULogErrorChannel("Path_AStar", "The ending tile isn't in the list of nodes!");
                 return;
+            }
+            if (adjacentOkay)
+            {
+                foreach (Tile tile in tileEnd.GetNeighbours())
+                {
+                    neighbors.Add(nodes[tile]);
+                }
             }
 
             goal = nodes[tileEnd];
@@ -99,7 +107,7 @@ public class Path_AStar
             // If we have a POSITIONAL goal, check to see if we are there.
             if (goal != null)
             {
-                if (current == goal)
+                if (current == goal || (adjacentOkay && neighbors.Contains(current)))
                 {
                     Reconstruct_path(came_From, current);
                     return;
--- a/Assets/StreamingAssets/Data/Furniture.xml
+++ b/Assets/StreamingAssets/Data/Furniture.xml
@@ -15,7 +15,7 @@
             <Inventory type="Steel Plate" amount="5" />
         </BuildingJob>
 
-        <DeconstructJob>
+        <DeconstructJob jobTime="1">
             <Inventory type="Steel Plate" amount="3" />
         </DeconstructJob>
 
@@ -54,7 +54,7 @@
             <Inventory type="Steel Plate" amount="2" />
         </BuildingJob>
 
-        <DeconstructJob>
+        <DeconstructJob jobTime="1">
             <Inventory type="Steel Plate" amount="1" />
         </DeconstructJob>
 
@@ -159,6 +159,8 @@
         <LinksToNeighbours>true</LinksToNeighbours>
 
         <BuildingJob jobTime="0" />
+        
+        <DeconstructJob jobTime="0"/>
 
         <Action event="OnUpdate" functionName="Stockpile_UpdateAction" />
         <DragType>area</DragType>
@@ -179,6 +181,10 @@
         <PowerConnection inputRate="1" outputRate="" capacity=""/>
 
         <BuildingJob jobTime="0" />
+        
+        <DeconstructJob jobTime="2">
+            <Inventory type="Steel Plate" amount="1" />
+        </DeconstructJob>
 
         <Animations>
           <Animation state="idle" looping="false">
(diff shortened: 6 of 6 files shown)


kd7uiy:
For reference I also had to change the A* pathfinding slightly, to allow it to find neighbor paths. This works when the pathfinding cost to get somewhere (Like a wall) that can't be achieved through other means. It will find the neighbor.

I have a hook to put in a sprite for tiles about to be deconstructed, but that hasn't been done yet.

sboigelot:
I'm not sure I understand your gif, what are you deconstructing?

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -756,8 +788,25 @@ public Inventory[] AcceptsForStorage()
     /// <summary>
     /// Deconstructs the furniture.
     /// </summary>
-    public void Deconstruct()
+    public void SetDeconstructJob()
     {
+        Jobs.CancelAll();
```
sboigelot: This will cancel current job but not prevent new job to be created from a lua call for instance.

Could we set a flag, like "public bool RequestedDeconstruction {get;set;}" to true and maybe disable all lua call (I know this will disable animation but...)

Any ideas?

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -772,6 +821,14 @@ public void Deconstruct()
             furniture.Jobs.CancelAll();
         }
 
+        if (deconstructInventory != null)
+        {
+            foreach (Inventory inv in deconstructInventory)
+            {
+                World.Current.inventoryManager.PlaceInventory(Tile, inv);
```
sboigelot: I'm not up to date with the latest inventory stuff, will placing multiple inventory on the same tile work?

Assets/Scripts/Models/Buildable/Furniture.cs
kd7uiy: Good catch! I think I have an idea, will give it a shot.

Assets/Scripts/Models/Buildable/Furniture.cs
kd7uiy: Good question, I'm not really sure what else to do, however. Hmmm...

Assets/Scripts/Models/Buildable/Furniture.cs
kd7uiy: Okay, I'm about to submit an update that I think resolves this. Let me know what you think!

kd7uiy:
I updated to a better gif as well.

Assets/Scripts/Models/Buildable/Furniture.cs
kd7uiy: I'm not really sure if it would work either, but I'm not sure what the solution is to it. It almost seems like the InventoryManager should just pick a neighbor tile if the current one is occupied, but I'm not sure if I want to do something like that in this PR. I could look in to it if you think it's an issue, however.

sboigelot: My solution for the landing pad was to (create and) use the 

```
ile tile = WorldController.Instance.World.GetFirstTileWithNoInventoryAround(6, (int)tradingCoordinates.x, (int)tradingCoordinates.y, (int)tradingCoordinates.z);
Inventory inv = new Inventory(tradeItem.Type, tradeItem.TradeAmount, tradeItem.TradeAmount);
WorldController.Instance.World.inventoryManager.PlaceInventory(tile, inv);
```

Not the best yet, but at least it give more options to the code

Assets/Scripts/Models/Area/World.cs
```diff
@@ -242,14 +242,14 @@ public Tile GetFirstTileWithNoInventoryAround(int maxOffset, int centerX, int ce
             {
                 offsetY = offset;
                 tile = GetTileAt(centerX + offsetX, centerY + offsetY, centerZ);
-                if (tile.Inventory == null)
+                if (tile.Inventory == null && tile.Furniture == null)
```
sboigelot: good catch

Assets/Scripts/Models/Inventory/InventoryManager.cs
```diff
@@ -27,6 +27,11 @@ public InventoryManager()
 
     public bool PlaceInventory(Tile tile, Inventory inventory)
     {
+        tile = World.Current.GetFirstTileWithValidInventoryPlacement(3, tile.X, tile.Y, tile.Z);
```
sboigelot: I'm not sure this should be the default behavior. A machine outputing to a single spot should jam and stop if the direct inventory output is full

Assets/Scripts/Models/Area/World.cs
kd7uiy: So, the TradingController does something that really should be the TradingModel that does. The tile placement shouldn't be figured out there, it should be a model thing. I've got in mind an upgrade that I'm going to work on that should resolve both. I'll leave the model logic in the TradingController for now, but...

sboigelot: Model in most c# pattern are data repositories and shouldn't modify data from other model, this is why we have controller. But I see your point in the way we use them. Feel free to propose an update ;)

Assets/Scripts/Models/Inventory/InventoryManager.cs
kd7uiy: Okay, I've got an update that will resolve that hopefully.

sboigelot:
Hehe too much changes I'm getting confused :) I'll read the code from start to end again

Assets/Scripts/Controllers/Events/TradeController.cs
```diff
@@ -107,9 +107,9 @@ private void TrasfertTradedItems(Trade trade, Vector3 tradingCoordinates)
         {
             if (tradeItem.TradeAmount > 0)
             {
-                Tile tile = WorldController.Instance.World.GetFirstTileWithNoInventoryAround(6, (int)tradingCoordinates.x, (int)tradingCoordinates.y, (int)tradingCoordinates.z);
+                Tile tile = WorldController.Instance.World.GetTileAt((int)tradingCoordinates.x, (int)tradingCoordinates.y, (int)tradingCoordinates.z);
                 Inventory inv = new Inventory(tradeItem.Type, tradeItem.TradeAmount, tradeItem.TradeAmount);
-                WorldController.Instance.World.inventoryManager.PlaceInventory(tile, inv);
+                WorldController.Instance.World.inventoryManager.PlaceInventoryAround(tile, inv);
```
sboigelot: This officially reduce the search radius from 6 to 3, why?
Taking into account the trade possible quantity and the already lost space because of the landing pad, I disagree, and it it is an issue it shouldn't be part of this PR.

Assets/Scripts/Models/Inventory/Inventory.cs
```diff
@@ -98,6 +98,11 @@ public string GetJobDescription()
         return string.Empty;
     }
 
+    public bool CanAccept(Inventory inv)
+    {
+        return inv.Type == Type && inv.StackSize + stackSize < MaxStackSize;
```
sboigelot: inv.StackSize + stackSize **<=** MaxStackSize ?

kd7uiy:
LOL, no problem. I've still got at least one more update I need to make, I just found a bug, so...

sboigelot:
@kd7uiy 

> LOL, no problem. I've still got at least one more update I need to make, I just found a bug, so...

Don't forget to update your title to [WIP] in that case

kd7uiy:
Fair enough. Just resolved the issue, so...

Assets/Scripts/Models/Inventory/Inventory.cs
kd7uiy: Good catch, fixed!

Assets/Scripts/Controllers/Events/TradeController.cs
kd7uiy: Okay, that's fair, I've allowed for the default search radius to be 3, but it can be overridden, and I set the TradeController to use 6.

sboigelot:
I value your changes, and honestly they are good and should be done at one point.
But the change your are making to the Place inventory that check for the inventory maxsize will currently break things.

The trade will not work anymore if you trade more then the stack size. At this point the code allowed to create stack too big for this kind of cases as we do not have anything like a chest, or hopper etc.

Don't get me wrong, this is good changes, but it should not break other features. 

You should either:

- Split this in different PR (but I woudn't)
- Add a bool to allow exceding the inventory max stack size and use it in the trade controller (and probably in the deconstruct as a furniture could give back more then 1 stack of X)
- Come up with a better way to spawn inventory, by splitting the stack too big into smaller stack. Keeping a list of stack that couldn't be spawned on the first pass and try again once some haul happens. Only fiish the deconstruct and the trade when all spawnable stack have been spawned.

Honestly, I would go for the middle option for this PR. So you can have your code improvement but still keep deconstruct and trade working in all cases including cases where it try to spawn a inventory too big.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -417,6 +424,11 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
     /// <param name="deltaTime">The time since the last update was called.</param>
     public void Update(float deltaTime)
     {
+        if (isBeingDestroyed)
```
TomMalbran: This could prevent a machine from working while it will be destroyed, like a door not opening. We could probably just prevent new jobs being added, but let the machines work until they get destroyed.

kd7uiy:
If you attempt to place a stack that is greater than the maximum stack, it will actually still work, it will only place it on an empty stack (No furniture or inventory, and can be entered) 

I agree with the other bit, will modify the isBeingDestroyed accordingly.

kd7uiy:
Okay, I found another bug that I think will improve things considerably. I think this one is worth another shot.

kd7uiy:
Also, updated the GIF again to a new and improved version.

dusho:
just from looking at gif, why is material not placed at same tile as deconstructed furniture?

sboigelot:
Approving the changes for now as I will be away all day and my comments have been answered.

It would be better indeed to set the furniture to null to the tile before placing the items, so the items will have a change to be on the furniture tiles.

kd7uiy:
@dusho That's a fair point. I've just changed it to place the items after the furniture is unplaced, which will allow for them to be placed at the same location as the deconstructed item, if there isn't something there already.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -730,6 +748,29 @@ public void ReadXmlBuildingJob(XmlReader reader)
     }
 
     /// <summary>
+    /// Reads the XML building job.
+    /// </summary>
+    /// <param name="reader">The XML reader to read from.</param>
+    public void ReadXmlDeconstructJob(XmlReader reader)
```
TomMalbran: Would it be better to create the job here as a Job prototype (similar to how is done with the build job) and then just clone it when deconstructing? If not. then why are we doing it for Build and not for Deconstruct, both should be similar.

TomMalbran:
And 1 more thing. You could also add the deconstruct job to the utilities.

Assets/Scripts/Models/Buildable/Furniture.cs
kd7uiy: I don't understand the Job Prototype system well enough to add in something else. It could be done, but I'm not sure of the utility at this point.

TomMalbran: Instead of storing the job params you just create the deconstruct job with a null Tile, and add it to the PrototypeManager (check how is done for the build job). Then when you are going to deconstruct, you grab the job prototype, clone it and set the tile.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -780,6 +843,15 @@ public void Deconstruct()
 
         Tile.UnplaceFurniture();
 
+        if (deconstructInventory != null)
+        {
+            foreach (Inventory inv in deconstructInventory)
+            {
+                inv.MaxStackSize = PrototypeManager.Inventory.Get(inv.Type).maxStackSize;
```
TomMalbran: I don't think that you need this. It should already be set to that amount.

Assets/Scripts/Models/Buildable/Utility.cs
```diff
@@ -80,6 +84,7 @@ private Utility(Utility other)
         typeTags = new HashSet<string>(other.typeTags);
         description = other.description;
         Tint = other.Tint;
+        deconstructInventory = other.deconstructInventory;
```
TomMalbran: Maybe we should actually copy the list here.

Assets/Scripts/Models/Buildable/Utility.cs
```diff
@@ -549,6 +551,15 @@ public void Deconstruct(Utility utility)
             Removed(this);
         }
 
+        if (deconstructInventory != null)
+        {
+            foreach (Inventory inv in deconstructInventory)
+            {
+                inv.MaxStackSize = PrototypeManager.Inventory.Get(inv.Type).maxStackSize;
```
TomMalbran: Same. I don't think it is required.

Assets/Scripts/Models/Prototypes/PrototypeManager.cs
```diff
@@ -22,9 +22,11 @@ public PrototypeManager()
         Inventory = new PrototypeMap<InventoryCommon>("Inventories", "Inventory");
         TileType = new PrototypeMap<TileType>("Tiles", "Tile");
         Furniture = new PrototypeMap<Furniture>("Furnitures", "Furniture");
-        FurnitureJob = new PrototypeMap<Job>();
+        FurnitureBuildJob = new PrototypeMap<Job>();
+        FurnitureDeconstructJob = new PrototypeMap<Job>();
         Utility = new PrototypeMap<Utility>("Utilities", "Utility");
-        UtilityJob = new PrototypeMap<Job>();
+        UtilityConstructJob = new PrototypeMap<Job>();
```
TomMalbran: This should be UtilityBuildJob or FurnitureBuildJob should be FurnitureConstructJob. Either way is fine as long as both use the same build name.

TomMalbran:
And I just realized, should we also deconstruct floor tiles? Maybe in another PR, since the Floor deconstruction is kind of weird.

Assets/Scripts/Models/Buildable/Utility.cs
kd7uiy: This is actually needed. Inventory max sizes aren't available when the Furniture.xml file is read in. This location seemed to be the easiest to squeeze in the max size.

Assets/Scripts/Models/Buildable/Utility.cs
kd7uiy: It actually doesn't matter. This inventory is copied later on. I can't imagine why one particular instance of a type of furniture would need to have different inventory requirements than any other would. As a result, it saves quite a bit of memory to just have a pointer to the proto object for this list, and not actually copy the list.

Assets/Scripts/Models/Prototypes/PrototypeManager.cs
kd7uiy: This is a good catch, will submit an update soon.

kd7uiy:
Floor, well, it could be done, but I think this PR has enough in it, a separate one would be a good thing. Meanwhile, I've made the other changes, or replied with my reasoning for the items that are in place. Let me know if there's anything else I should do!

Assets/Scripts/Models/Buildable/Utility.cs
TomMalbran: Sure. It makes sense. We will need to copy if we later decide to change the deconstruct inventory for an installed furniture. For example, if a furniture can get damage, maybe it will drop less stuff, or even some other type of inventory, some damaged thing, or something. But we can change it later, if we go in that direction.

Assets/Scripts/Models/Buildable/Utility.cs
TomMalbran: Right, I forgot that we shouldn't relay on the prototype loading order. Maybe we shouldn't even create other entities while creating a prototype. An option would be to save just the type and the amount when reading the xml and create the inventory at this point.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -756,8 +802,25 @@ public Inventory[] AcceptsForStorage()
     /// <summary>
     /// Deconstructs the furniture.
```
TomMalbran: Update the comment here

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -756,8 +802,25 @@ public Inventory[] AcceptsForStorage()
...
+        Jobs.CancelAll();
+
+        Job job = PrototypeManager.FurnitureDeconstructJob.Get(Type).Clone();
+        job.tile = Tile;
+        job.OnJobCompleted += (inJob) => Deconstruct();
+
+        World.Current.jobQueue.Enqueue(job);
+    }
+
+    public void Deconstruct()
```
TomMalbran: This could have some comments.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -417,28 +422,31 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
     /// <param name="deltaTime">The time since the last update was called.</param>
     public void Update(float deltaTime)
     {
-        if (PowerConnection != null && PowerConnection.IsPowerConsumer && HasPower() == false)
+        if (isBeingDestroyed == false)
```
TomMalbran: This isn't preventing new jobs to be created. Most new jobs are created from Lua, when OnUpdate is trigger from EventAction. And a furniture should still stop working when it has no power and is set to be deconstructed.

If we want to prevent new jobs to be added, we can do so in FurnitureJobs.Add. I think that a furniture could still work until it gets deconstructed, but maybe it shouldn't be used. For beds, we could not use the ones that will be deconstructed.

kd7uiy:
Okay, comments are added, and I improved the rejection of jobs for furniture being destroyed.

sboigelot:

kd7uiy:
@sboigelot You beat me to it;-) Just about to get the deconflicted update pushed.

Assets/Scripts/Models/Buildable/Utility.cs
```diff
@@ -529,6 +511,26 @@ public void PauseJobs()
     }
 
     /// <summary>
+    /// Deconstructs the furniture.
```
TomMalbran: This doesn't deconstruct the furniture, it creates a job to deconstruct it, if it is not set to be deconstructed already.

Assets/StreamingAssets/Images/UI/CursorCircle.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Sprites>
+	<Sprite name="CursorCircle"     x="0" y="0" w="1" h="1" pixelPerUnit="32" />
```
TomMalbran: Spaces please

kd7uiy:
Okay, I think I've fixed your latest issues. Bad me for not checking after every change Stylecop...

koosemose:
Conflicted.

kd7uiy:
Yeah, I thought that was going to happen... Sigh. Anyways, I think I've got it.

koosemose:
Should perhaps also check if devmode is active (`if (Settings.GetSetting("DialogBoxSettings_developerModeToggle", false))`) and instantly deconstruct if it is.

kd7uiy:
Okay, I've added a dev mode instant deconstruct option.

kd7uiy:
Yeah, for a while I'm going to work on the pre-existing bugs. There are a few of them...

PR #1396, opened 2016-09-18
[Tiny, Fix] Sorting order of furniture now being applied on load

### LINKED ISSUE #1389: Sorting order lost after save/load
The sorting introduced in #1363 is regrettably lost after save/load...

jaselito:
Fixes #1389 

The sorting layer was being assigned before the position of the game object was set. It was defaulting to 0 for all loaded furniture sprites.


Raketfart:
Ahh yes. Thanks!

PR #1397, opened 2016-09-18
This should definitely fix the GUI issue #1349 

TheAvaren:
Found a temporary fix for issue #1349,  possibly till we introduce the scriptable/modular GUI system, but i don't know much about that yet.

+added Element Layout Component and VerticalLayoutGroup to ContextMenuItem.prefab
+added Content Size Fitter Component to ContextMenu.prefab

so if you don't know already issue #1349 was annoying unfortunately i under estimated my ability to fix it.(i have never bothered to mess with unity UI's).

but after alot of frustration here it is a small fix that took too much time.
if you have any question's read below.

1.The UI is dynamically update's its size based on the length and width the text take up using some UI,
Components.

2.There is no scripting done to achieve this, it all just unity built in components based off of this
https://docs.unity3d.com/Manual/HOWTO-UIFitContentSize.html.

3.please ask me anything you want if you want.

### DIFF
--- a/Assets/Resources/Prefab/ContextMenuItem.prefab
+++ b/Assets/Resources/Prefab/ContextMenuItem.prefab
@@ -23,6 +23,8 @@ GameObject:
   - 114: {fileID: 114000011520895160}
   - 114: {fileID: 114000014070300310}
   - 114: {fileID: 114000011943728566}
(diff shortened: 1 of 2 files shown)


mikejbrown:
Why closed? Are you/anyone still working on #1349?

PR #1400, opened 2016-09-18
Add a fast update (every frame)  to furniture to allow smooth animation

### LINKED ISSUE #1383: [BUG] Update is run at 5 fps
I was debugging why furniture animation was so laggy. I figured out that deltatime in lua and furniture update was 0.2 second. This seems to be because TimeManager calls updates 5 times a second, from what I could see.

This means that even though we see 60 fps in the counter, nothing updates at more than 5 FPS.

sboigelot:
Fixes #1383 

Duplicate the EventAction in furniture to add FrameAction

EventAction behavior unchanged
FrameAction will be call every frame, so animation can be smooth

Fix the "door slam open effect"

### DIFF
--- a/Assets/Scripts/Models/Buildable/Furniture.cs
+++ b/Assets/Scripts/Models/Buildable/Furniture.cs
@@ -71,7 +71,6 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
         Tint = Color.white;
         VerticalDoor = false;
         EventActions = new EventActions();
-        FrameActions = new EventActions();
         
         contextMenuLuaActions = new List<ContextMenuLuaAction>();
         Parameters = new Parameter();
@@ -116,11 +115,6 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
         if (other.EventActions != null)
         {
             EventActions = other.EventActions.Clone();
-        }
-
-        if (other.FrameActions != null)
-        {
-            FrameActions = other.FrameActions.Clone();
         }
         
         if (other.contextMenuLuaActions != null)
@@ -207,16 +201,7 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
     /// </summary>
     /// <value>The event actions that is called on update.</value>
     public EventActions EventActions { get; private set; }
-
-    /// <summary>
-    /// Use lightly
-    /// Gets the FrameAction for the current furniture.
-    /// These actions are called every frame. They get passed the furniture
-    /// they belong to, plus a deltaTime (which defaults to 0).
-    /// </summary>
-    /// <value>The event actions that is called on update.</value>
-    public EventActions FrameActions { get; private set; }
-
+    
     /// <summary>
     /// Gets the Connection that the furniture has to the power system.
     /// </summary>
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -638,6 +661,11 @@ public void ReadXmlPrototype(XmlReader readerParent)
                     EventActions.ReadXml(subtree);
                     subtree.Close();
                     break;
+                case "FrameAction":
+                    XmlReader frame_subtree = reader.ReadSubtree();
```
bjubes: variable names should be camelCase not under_scored

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -425,12 +440,20 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
         return obj;
     }
 
+    public void EveryFrameUpdate(float deltaTime)
```
TomMalbran: We should add docs here and update the docs in FixedFrameUpdate, so that is clear why there are 2 update methods.

Raketfart:
👍  Looks good. Should we rename OnUpdate_AirlockDoor and OnUpdate_Door so it's obvious in lua that these are frameactions?

koosemose:
It is completely unnecessary to duplicate EventActions. EventActions aren't tied to a specific update cycle, they just trigger whenever you tell them to, which in the case of OnUpdate just happens to be in the Fixed Frequency Update. Just create different action for EventAction.

For Example:
```xml
<Action event="OnFastUpdate" functionName="OnUpdate_AirlockDoor" />
```
and then in EveryFrameUpdate do:
```cs
EventActions.Trigger("OnFastUpdate", this, deltaTime);
```

sboigelot:
@koosemose you are making sense :D I didn't think about it indeed. I'll change this during the night or tomorow

sboigelot:
Actually this was super easy to change. @koosemose I implemented your change request

TomMalbran:
If the Lua functions OnUpdate are called every frame. Do we also need to call them on a fixed frequency. Since the OnUpate for doors is not linked to any job or power or any of that stuff, we probably don't. And if we don't, then we can also change the Lua function name.

sboigelot:
@TomMalbran I'm not sure I understand what you mean?

TomMalbran:
Sorry, I read `OnUpdate_Leak_Door` as `OnUpdate_Door`, and got confused. We should change `OnUpdate_Door` to OnFastUpdate_Door`. Same for Airlock.

Raketfart:
@sboigelot Don't we need all the funtionality of FixedFrequencyUpdate in EveryFrameUpdate, like checks for power and calling animation.update?

sboigelot:
@Raketfart no we don't really, the fast update should just control the current animation, the slow one will change the current animation state in a best scenario. You may end up in an animation different then the real action for a few frames but what's a few frame, we, slow human will never notice

Raketfart:
Sure, if the animation is smooth, then I'm happy.

Assets/Scripts/Models/Buildable/Furniture.cs
sboigelot: not in the final code

Assets/Scripts/Models/Buildable/Furniture.cs
sboigelot: comment added

Issue #975, opened 2016-08-28
Power Cell Press puts cells in walls


kd7uiy:
![capture](https://cloud.githubusercontent.com/assets/5951269/18033617/ef3e5be8-6cf7-11e6-906f-f545abd17695.PNG)

Not quite sure what the solution for this should be, but this doesn't seem right.

Fatalexceptionz:
The Power Cell sprite looks all sorts of broken too...

Grenkin1988:
Same issue - metal floating near landing pad
![inventory](https://cloud.githubusercontent.com/assets/2979432/18034239/6820be1e-6d40-11e6-86e9-6dc5457d25d2.PNG)

kd7uiy:
I was just about to include something similar;-) Nice work!

sboigelot:
The metal hanging out of the landing pad is still a test feature, we shouldn't worry about it. 

#817 (+following PR) will replace this shortly

Grenkin1988:
This is because existing solution do not care what is there is spawn point, but it should

sboigelot:
Before implementing a solution that care, we should show the player where the job point and spawn points (and all other points) are during the construction or when a furniture is selected

MANDAL0R3:
@sboigelot, What do you think about showing the job and spawn point with something like this? 
![idea](https://cloud.githubusercontent.com/assets/12440052/18041964/9d587da6-6d8b-11e6-9704-3b49a3768c54.png)

sboigelot:
@MANDAL0R3 it looks great to me

MANDAL0R3:
@sboigelot Well if it was supposed to be then its a bug, for me the raw iron is produced where the i put the blue square

sboigelot:
@MANDAL0R3 ho yes sorry, my bad, I confused it with the metal smelter, you are totally correct!

MANDAL0R3:
@sboigelot, I have been doing math home work all night, i honestly thought that i had gone insane for a minute :D

dusho:
so those in/out slots should be checked when placing devices for building, whether there are empty slots (or compatible inventory item placed there already)
There was a discussion on generic factory/device #604 , so it would really make sense to create subclass from Furniture (Factory or Device, ... ) that would contain input and output port positions together with stuff that all processing devices will have

Grenkin1988:
9 days old bug is not good :-(

dusho:
[...]

PR #1409, opened 2016-09-19
Fix for workshop jobs getting cancelled

### LINKED ISSUE #1311: Power Cell Press stops asking for more input after receiving one type of input
While trying to test if #975 was still an existing bug, I discovered this new one. 

To ensure it wasn't a different issue about machines not continuing to request resources when one isn't found, I spawned a stack of steel plates, a stack of uranium, and built enough RTGs to power everything, and only then built the Power Cell Press. It will request both items, and if you have enough crew to start both at the same time, both resources will be grabbed, however, as soon as one of the crew arrives with whichever resources (acts the same with either one, just depends on which arrives first), the second crew will not deliver the inventory and just ends up holding on to it. If you don't have enough crew to carry both towards it at the same time, the second resource will never be grabbed.

dusho:
Power cell should now get all the materials needed:
image
removed old code (evidently not active until recently)
This should fix #1311 and hopefully #975 is now testable again..

### DIFF
--- a/Assets/Scripts/Models/Buildable/FurnitureWorkshop.cs
+++ b/Assets/Scripts/Models/Buildable/FurnitureWorkshop.cs
@@ -259,13 +259,7 @@ public class FurnitureWorkshop
                     furniture.Tile.Y + reqInputItem.SlotPosY,
                     furniture.Tile.Z);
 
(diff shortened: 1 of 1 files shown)


koosemose:
Works properly, merged.

PR #1411, opened 2016-09-19
[Fix] Fixes furniture being unable to replace allowed types.

koosemose:
Stops internal workspots from being reserved (there's no need to, as the tile will already be unable to be built on by another furniture unless that furniture can replace it, in which case it shouldn't care about the workspot)

Also stops a furniture from checking if it's workspot is clear if the workspot is internal (technically it just returns true when checking if the workspot is clear), which again is a problem for when replacing furniture.


mikejbrown:

koosemose:
Conflicts Resolved.

dusho:
wanted to test this, but don't know how actually.. not sure I know what was broken
but check for workspot being free seems to be working ok when placing furniture.. can't be even placed over stockpile (I guess it's ok)

koosemose:
Try replacing a wall with a door or a pump or anything else that should be replaceable by the furniture being placed, it won't work in master currently, but should work in this PR.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -422,6 +422,11 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
         return obj;
     }
 
+    public bool WorkspotIsInternal()
```
TomMalbran: Should this be in BuildableJobs?

koosemose: I don't think so, furniture is, to the best of my knowledge, the only thing that has workspots. Isn't BuildableJobs the jobs for building furnitures?

TomMalbran: `Jobs` is an instance of `BuildableJobs`. `BuildableJobs` holds everything Job related of the furniture. Was called `FurnitureJobs`, but changed it when I added them to the Utilities, since that is the Interface name.

TomMalbran: I still think this should go in `BuildableJobs`.

koosemose: The primary problem is that it also needs to be aware of the size of the furniture (or the furniture itself to get the size from) and it isn't currently, and otherwise doesn't need to be aware of.

TomMalbran: BuildableJobs already keeps a reference to the instance of the class (Furniture/Utility) that creates it, and that class must implement IBuildable, which means they they must have a Weight and a Height property, which means that the if this method was in BuildableJobs, it can be written like:

```
public bool WorkspotIsInternal()
{
    return WorkSpotOffset.x >= 0 && WorkSpotOffset.x < buildable.Width && WorkSpotOffset.y >= 0 && WorkSpotOffset.y < buildable.Height;
}
```

It makes more sense to place it there than in Furniture.

koosemose: Ahhh the buildable  I missed that when I was looking through the class

koosemose:
And that should take care of any remaining concerns.

Assets/Scripts/Models/Buildable/BuildableJobs.cs
```diff
@@ -154,6 +154,11 @@ public bool HasJobWithPredicate(Func<Job, bool> predicate, out Job job)
         return job != null;
     }
 
+    public bool WorkspotIsInternal()
```
TomMalbran: Can we name this `WorkSpotIsInternal`? Or `IsWorkSpotInternal` better? (uppercase S for Spot)
Does it needs docs?

koosemose:
:( It would probably help if I changed everywhere that uses that function...

TomMalbran:
@koosemose Thanks. The code looks great now.

koosemose:
Self-merging this now as it fixes a bug currently on master.

PR #1412, opened 2016-09-19
Furniture progress info and power cell usage

dusho:
Added new function for furniture to show progress (called with GetAdditionalInfo())
Power generator now uses Power cell as fuel, lasts 30 seconds, shows remaining fuel when selected:
image
plus tiny bug fix for clamping in Connection.cs (IndexOutOfRangeException: Array index is out of range.)

**Known issue:** Can't force to use just one power cell ? Looks like it takes 4 to make power generator work.. don't know why, don't see problem in settings (also previously even though it required 5 uranium, it took 20 to start the generator.. something funky is there)

### DIFF
--- a/Assets/Scripts/Models/Buildable/Furniture.cs
+++ b/Assets/Scripts/Models/Buildable/Furniture.cs
@@ -34,6 +34,11 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
     /// </summary>
     private string getSpriteNameAction;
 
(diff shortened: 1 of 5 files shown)


koosemose:
Since you're already changing the burn time of the power generator, could you crank it up a little bit more? It makes it easier to test things if the generator lasts a little longer (I'd go for at least 300, though when I tweak it for my own purposes I prefer somewhere between 1000-2000)

Issue #1081, opened 2016-09-01
[ART]Isometric vs Top Down vs Fully Flat


longtomjr:
This is a discussion thread for Isometric vs Topdown vs Fully Flat
*Examples*
### Top Down (aka angled Top Down)
![](http://liza.io/images/2012/02/creatures-and-castles.png)
### Isometric
![](http://1.bp.blogspot.com/--w3dyXRR5pA/Tt6fXQCDs2I/AAAAAAAABDw/UwFy8XUg9Tw/s1600/display_1.jpg)
### Fully Flat (aka true Top Down)
![](http://1.bp.blogspot.com/-DdgIcm5sohI/UY0vkAR2m9I/AAAAAAAALF8/sP3WdN-7AWw/s1600/Monaco-2.jpg)

Tranberry:
I love isometric, but that would mean that I'm out. I can't make that kind of art (I won't have the time to do that). I vote for keeping it top down at an angle, else you have to suffer my code. 

Another reason to keep it top-down is that most people can make pretty decent sprites that way, imo.

longtomjr:
Here is my 2 cents:  
**TopDown:** Prolly the best for this project. Having things at an angle helps to convey more info about that object and it is simple and clean. This is also the easiest to draw for and have shown success in the building game scene.  
**Isometric:** Hard, really hard. Looks great but I don't think it will work for this project. There will need to be camera changes and being in space will make it necessary to render multiple levels. This would be great if we were a full time studio with dedicated artists, but for this project I don't think it will work.  
**Fully Flat:** This is also a really hard style to pull of because you are limited from one angle. Another problem is that it might be a lot more difficult to convey the necessary info when you can only draw one side.  

**TLDR**(*though you should really read my thoughts)*: I am going for TopDown!

longtomjr:
also, just some pings...  
@ComputerOverlord, @Mizipzor , @dusho , @Tranberry , @longtomjr , @abackwood @alexisfcote

Tranberry:
@TheRiven ^ @longtomjr was lazy.

TheRiven:
That isometric pic looks awesome :+1: 

But I do feel top-down is a good look for us to continue with, although the fully flat could probably work as well. :thought_balloon:

Grenkin1988:
**Isometric** is so beautiful. But it would be hell to do it.
**Top down** I think

vogonistic:
I don't think this should be an isometric game. 

[...]

PR #1416, opened 2016-09-19
Rotating furnitures

Ohda:
**EDITED** This PR allow the rotation of the furnitures.
If the furniture got  `<CanRotate>true</CanRotate>` in the XML, the player can press R and T to rotate the furniture when placing it.

Only Beds and solar panel have been allowed to rotate for now as most of our sprites don't support it.
(shadow issue, perspective issue.

The feature works with save/loading system.
The next step is to manage rotation of Jobspot, workspot.


Tranberry:
I'm not sure any square (1x1u, 2x2u..) needs rotation. This will force us to add 1-3 more sprites to each furniture or revisit the discussion in https://github.com/TeamPorcupine/ProjectPorcupine/issues/1081 .

The artist are few and far between here, and adding more work to this will hinder some progress as things that will break with art won't become apparent in a speedy manner.

**I do think this is a vital feature,** but maybe not allow all furnitures to be rotated until we got some artists actually replacing the needed art (like the bed).

alexfriesen:
the rotation of the bed looks wrong

Raketfart:
We definetly want to be able to rotate some of the furniture - defined i xml. I imagine some furniture has two rotations, and some have four. I agree that most square furniture doesn't need to be rotated.

Most often, we probably don't want the sprite to just rotate, but instead set a "facing/orientation" on the furniture, and then GetSpriteName should fetch, and maybe flip the right sprite. When rotating we should switch the width and height of the furniture, so 2x3 becomes 3x2. 

Solar panels look ok when rotated, if the size of the furniture gets fixed.

sboigelot:
@Raketfart you would rotate square furniture to face the workspots inputs and outputs

Tranberry:
@sboigelot I believe @Raketfart meant that there would be independent sprites on the sprite-sheet for each facing.

Raketfart:
@sboigelot I Agree. I meant things like the RTG or the landing pad, that take up the same space, and has no functional or cosmetic difference if rotated.

Ohda:
If i can finsih this i think i will only add rotation on bed and solar panel for now.

Ohda:
Placement problem are fixed.

Still looking forward for the saving/loading issues.

Ohda:
Remove the WIP tag.

- The furniture rotation is bind to R and T Key.
- If the furniture got <CanRotate>true</CanRotate> in the XML, the rotation is allowed when placing the furniture.
- Bed and Solar panel are active for rotation.
- The furnitures rotation are saved/Loaded from the save file.

Assets/Scripts/Models/Area/World.cs
```diff
@@ -194,6 +194,7 @@ public Furniture PlaceFurniture(string type, Tile t, bool doRoomFloodFill = true
         }
 
         Furniture furn = PrototypeManager.Furniture.Get(type);
+        furn.SetRotation(rotation);
```
TomMalbran: I think that we should clone the furniture before setting the rotation.

Assets/Scripts/Models/Area/World.cs
```diff
@@ -240,10 +241,11 @@ public Furniture PlaceFurniture(Furniture furniture, Tile t, bool doRoomFloodFil
...
     public void JobComplete_FurnitureBuilding(Job theJob)
     {
-        // Let our workspot tile know it is no longer reserved for us
-        UnreserveTileAsWorkSpot((Furniture)theJob.buildablePrototype, theJob.tile);
+        Furniture furn = (Furniture)theJob.buildablePrototype;
 
-        PlaceFurniture(theJob.JobObjectType, theJob.tile);
+        // Let our workspot tile know it is no longer reserved for us
+        UnreserveTileAsWorkSpot(furn, theJob.tile);
+        PlaceFurniture(theJob.JobObjectType, theJob.tile, true, furn.Rotation);
```
TomMalbran: should we just use furn at this point? is already a clone... and we might not even need to clone it again. What do you think?

Ohda: Not sure why i was afraid to doing it. Work like a charm.

Assets/Scripts/Controllers/Sprites/JobSpriteController.cs
```diff
@@ -122,7 +122,8 @@ protected override void OnCreated(Job job)
                 sr.sprite = usc.GetSpriteForUtility(job.JobObjectType);
             }
 
-            job_go.transform.position = job.tile.Vector3 + ImageUtils.SpritePivotOffset(sr.sprite);
+            job_go.transform.position = job.tile.Vector3 + ImageUtils.SpritePivotOffset(sr.sprite, BuildModeController.Instance.GetCurrentPreviewRotation());
```
TomMalbran: Should we only rotate furniture, and maybe use `job.buildablePrototype.Rotation` (if we only check for furniture)?

Assets/Scripts/Models/Area/World.cs
TomMalbran: Now that furn is no longer the prototype, it does work.

Ohda:
All suggestions from @TomMalbran as been implented. Thanks you !

koosemose:
Conflicted

Ohda:
Conflict resolved

dusho:
so in the newest version only bed and solar panels are rotate-able? From what I tried, all works ok.
Would be curious how smelter and other stuff would work - probable can't since offsets are placed specifically in LUA?

TomMalbran:
They should all work. But the graphics look bad when rotating some furniture. We would need to create multiple sprites for the rotations that we want.

The offsets are not set in Lua, but the offset vectors will need to be changed so that it works for the rotations. Is totally doable though.

Ohda:
I'll try to add a Jobspot to the bed to work on the rotation of thoses spots that need rotation.
Can it be another PR ? I'am afraid of the big one !

koosemose:
This does not seem to be working anymore, unless you changed the rotation keys from B and N, in which case please change it in the top post for easier testing.

Ohda:
Yes this is R and T now as suggested.
Top post edited

koosemose:
Looks good to me, Stylecop clean. Merged.

PR #1417, opened 2016-09-19
[Fix] Properly deconstruct utilities

koosemose:
Utilities did not deconstruct previously, both because BMC didn't call deconstruct on utilities, and the utilities deconstruct method didn't handle things properly.

Deconstruction will now first look for a furniture to deconstruct, and then if there is a pending build job, will cancel it (as before) and then deconstruct the last utility built.

This may not be that absolutely best optimal desired behavior (I would like to be able to have things figure out to delete power cables if you've drug deconstruct over power cables, but that would require much more complicated logic, and possibly drastic changes to how the buildModeController performs deconstruction.) but it is functional whereas previously it was not and would completely destroy the utilities hashSet making it impossible for any utilities to be added to the tile or any that were there to do anything.

### DIFF
--- a/Assets/Scripts/Controllers/InputOutput/BuildModeController.cs
+++ b/Assets/Scripts/Controllers/InputOutput/BuildModeController.cs
@@ -276,6 +276,10 @@ public class BuildModeController
             {
                 tile.PendingBuildJob.CancelJob();
             }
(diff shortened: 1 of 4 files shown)


koosemose:
As this fixes a bug currently on master, I will be self-merging this.

PR #1419, opened 2016-09-19
No steel plate from mining (deconstruction fix)

dusho:
mining ores gave you also steel plate (guess from deconstruction)
also made iron mine spot color consistent with iron color
image


kd7uiy:
I guess I got a bit carried away from that... The change looks good, I won't speak to the artwork change.

ComputerOverlord:
It shouldn't be steel plates - it should be unprocessed iron

dusho:
ye.. you will get appropriate resources when mining this 'mining spot', but without this 'fix' you will also get additional 1 steel plate

PR #1420, opened 2016-09-19
World update frequency for visible/invisible furniture

Raketfart:
This keeps track of visible and invisible furniture, so we can manage update frequency of visible furniture. See #1383.

I'll let this one wait for #1400 to be finished, as they are closely related.

### DIFF
--- a/Assets/Scripts/Controllers/WorldController.cs
+++ b/Assets/Scripts/Controllers/WorldController.cs
@@ -131,6 +131,7 @@ public class WorldController : MonoBehaviour
         spawnInventoryController.SetUIVisibility(Settings.GetSetting("DialogBoxSettings_developerModeToggle", false));
 
         cameraController.Initialize();
+        cameraController.Moved += this.World.OnCameraMoved;
 
         // Initialising controllers.
         GameObject canvas = GameObject.Find("Canvas");
--- a/Assets/Scripts/Controllers/InputOutput/CameraController.cs
+++ b/Assets/Scripts/Controllers/InputOutput/CameraController.cs
@@ -6,6 +6,7 @@
 // file LICENSE, which is part of this source code package, for details.
 // ====================================================
 #endregion
+using System;
 using UnityEngine;
 
 public class CameraController
@@ -64,6 +65,8 @@ public class CameraController
         TimeManager.Instance.EveryFrameNotModal += (time) => Update();
     }
 
+    public event Action<Bounds> Moved;
+
     public int CurrentLayer
     {
         get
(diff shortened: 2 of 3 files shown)


bjubes:
@Raketfart #1400 has been merged, so is this no longer WIP?

Raketfart:
It works now, but i still want to change a few things.

Raketfart:
Changed the camera moved to an event.

This optimization can be further optimized when needed, but I think it's enough for now.

Assets/Scripts/Controllers/WorldController.cs
```diff
@@ -131,6 +131,7 @@ public void Start()
         spawnInventoryController.SetUIVisibility(Settings.GetSetting("DialogBoxSettings_developerModeToggle", false));
 
         cameraController.Initialize();
+        cameraController.Moved += this.World.OnCameraMoved;
```
TomMalbran: I would like this on World, since is for the World only.

TomMalbran:
I don't like that you are assuming that a furniture only updates if it has EventActions. Even if a furniture does not have an event action it can still do things. For example, right now the Metal Furnace works only with the Workshop options and has no EventAction.

I think that in the future we can add more of those components, like one for Doors, since all Doors work in the same way, we could have parameters for the speed and such and then have all the code be in C#. You could still do it like now on update for something not supported. It would be a bit of a pain to add the same functions for each type of Door we have. And other components could be added to Power. I think that when we have several pieces of furniture that do really similar things in the Lua functions, we should try to make components. And if those don't fit for certain mod, they still have the option to do it in Lua. 

In the future we could have furniture without event actions that do a fast update. So I would remove the EventActions check for now and just split in visible and invisible furniture for the fast update. We can use it later to distribute the furniture updates in frames.

Assets/Scripts/Controllers/WorldController.cs
Raketfart: Me too, but when I tried, it seems that world is created before cameracontroller, so I'm not sure how

Raketfart:
@TomMalbran You make a good point. I was assuming that we only would use lua updates on furniture, and that could lead to weird bugs, when we expand furniture.
I've rolled back that change, and left all furniture in one master list. I still think it would be nice to ignore furniture like walls when sorting the list, but we can save that for later.

TomMalbran:
We might need to add a property in the furniture that tell us if we need to update it or not. I am not sure if we can assume that, maybe if we check for Events, Power usage, Workshop, Animations and each thing that is done inside the update. Even for walls, we could have at some point some super fancy walls with neon lights, that would need to be updated.

Assets/Scripts/Controllers/WorldController.cs
TomMalbran: We would need to create the CameraController before the World. Maybe later, since we have things to improve in the event driven system.

TomMalbran:
Code looks good. But haven't tested it.

Raketfart:
@TomMalbran I was thinking the same - maybe asking the furniture. But that would still make it easy to make bugs when just working on furniture, because you maybe don't know about this. So I think we should leave this as is, and expand it, when we know more about how furniture works.

TomMalbran:
@Raketfart Yes. Is an improvement that can be done later, if required. At that point we can figure out the best way to do it.

For now. This PR is good.

koosemose:
Looks Good, Merged

PR #1426, opened 2016-09-20
Modable Overlays

TomMalbran:
The overlays now use the PrototypeManager for the prototypes and the FunctionsManager for the functions. Since I was changing it, I also added localization support for them.


Assets/Scripts/Models/Functions/LuaFunctions.cs
```diff
@@ -46,6 +46,16 @@ public void RegisterGlobal(Type type)
     }
 
     /// <summary>
+    /// Determines whether there is a Lua global with the given name.
+    /// </summary>
+    /// <returns><c>true</c> if there is a global with the given name; otherwise, <c>false</c>.</returns>
+    /// <param name="name">The global name.</param>
+    public bool HasGlobal(string name)
+    {
+        return script.Globals[name] != null;
```
mikejbrown: Might want to do a null check on name here.

Assets/Scripts/UI/Overlay/OverlayMap.cs
```diff
@@ -287,10 +270,11 @@ public void SetOverlay(string name)
                 }
 
                 Tile tile = WorldController.Instance.GetTileAtWorldCoord(new Vector3(x, y, z));
-                return (int)script.Call(handle, new object[] { tile, World.Current }).ToScalar().CastToNumber();
+                DynValue result = FunctionsManager.Overlay.Call(descr.LuaFunctionName, new object[] { tile, World.Current });
+                return (int)result.ToScalar().CastToNumber();
```
mikejbrown: If result is null or non-numeric the resulting exception wouldn't easily trace back to the guilty lua function, or would it? (I'm expecting it would be "chunk blah blah blah" or a random NRE or the like.) Might just want to do a check here and throw an error msg with some meaningful information?

mikejbrown: Basically Lua debugging is in a sad state and needs improvement across the board. Might as well make it as good as possible in the files this PR touches.

Assets/Scripts/Models/Prototypes/PrototypeManager.cs
```diff
@@ -114,6 +115,12 @@ public PrototypeManager()
     /// <summary>
     /// Gets the headline prototype map.
     /// </summary>
-    /// <value>The scheduler event prototype map.</value>
+    /// <value>The headline prototype map.</value>
```
mikejbrown: Good catch.

Assets/Scripts/Models/Functions/LuaFunctions.cs
TomMalbran: Good point. Will do.

TomMalbran:
Fixed. And also found some StyleCop errors.

Assets/Scripts/UI/Overlay/OverlayMap.cs
```diff
@@ -271,7 +271,14 @@ public void SetOverlay(string name)
 
                 Tile tile = WorldController.Instance.GetTileAtWorldCoord(new Vector3(x, y, z));
                 DynValue result = FunctionsManager.Overlay.Call(descr.LuaFunctionName, new object[] { tile, World.Current });
-                return (int)result.ToScalar().CastToNumber();
+                double? value = result.CastToNumber();
+                if (value == null)
+                {
+                    Debug.ULogErrorChannel("OverlayMap", string.Format("The return value from the function named '{0}' was null", descr.LuaFunctionName));
```
mikejbrown: :+1:

mikejbrown: Actually just tried testing this by changing
```lua
function oxygenValueAt(tile)
    return "blahblahblah"
end
```
(yeah, I know, but modders may be stupid) and Unity actually freezes.

mikejbrown: I think it's the debug getting called in a tight loop. Commenting out the debug line stops it from freezing.

mikejbrown: Other than this the PR is fine to merge though... so if you like I can merge as in and make a separate issue out of this?

TomMalbran: I guess it makes sense since this callback is called once per tile. If we go back to what it was before. I can make it log once.

mikejbrown: You still probably want the null check, no? Or does CastToNumber already handle it if assigning to an `double` instead of `double?`? Just get rid of the debug here and figure out somewhere else to put it.

TomMalbran: CastToNumber returns a null if it fails to cast. If you then try to cast null to an int you get lots of errors. Is a lot easier to have the debug there.

TomMalbran: I think that logging once will be enough.

mikejbrown:
Very good. And :tada: for fixing the localizations!

TomMalbran:
NP. it so much easier to add localizations now that they are in master.

Issue #1425, opened 2016-09-20
Missing English translations


mikejbrown:
These are just the ones I found in a ~minutes play through. Feel free to add any others you find. (I am not expecting a separate issue for each language. This is only because en_US is the default and all keys should be there in the repo.)

- [x] Translation for Empty in en_US language failed: Key not in dictionary.
- [x] Translation for Floor in en_US language failed: Key not in dictionary.
- [x] Translation for Ladder in en_US language failed: Key not in dictionary.
- [x] Translation for furn_metal_furnace in en_US language failed: Key not in dictionary.
- [x] Translation for furn_vent in en_US language failed: Key not in dictionary.
- [x] Translation for util_power_cable in en_US language failed: Key not in dictionary.
- [x] Translation for util_water_pipe in en_US language failed: Key not in dictionary.
- [ ] Translation for  in en_US language failed: Key not in dictionary. (hmmm... not sure why this one was trying to access an empty key...)

mikejbrown:
Should add to v0.1 milestone?

svmnotn:
probably, we probably want to look at that whole system before doing anything else

svmnotn:
I believe at some point we were considering not having an en_US and just using the xml data as a deafult/fallback (so the xml for tiles would include the human en_US name). But I don't think that ever got much farther than a thought.

svmnotn:
Also I believe we are still waiting for quill to make a locale repo under the organization.

mikejbrown:
@svmnotn I just merged #1418 so en_US is now in the main repo. :D (Thanks @sboigelot!)

And I think having fallback defaults for en_US in the xml makes sense.

koosemose:
Added a PR for some of these, I'm not entirely sure why the floor tiles are looking for Empty, Floor, and Ladder, as that's not what's defined for them in the xml... presumably they don't properly make use of the translation keys?

mikejbrown:
Merged in #1427 which fixed some of the above and a few I had missed in my quick play through. Thanks @koosemose!

mikejbrown:
@koosemose #1429 fixes the localization issue for Empty, Floor and Ladder.

mikejbrown:
Tracking down the empty key one is turning out to be a pain...

koosemose:
[...]

PR #1429, opened 2016-09-20
[small] Fix ConstructionMenu.cs localization for tile jobs

mikejbrown:
Previously the menu lookup for Empty, Tile and Ladder were broken because
they were based off the tileType.Type rather than the localized name.
This fixes the issue by adding a localization lookup in
RenderTileButtions.

Fixes an issue spotted by @koosemose in the comments at #1425.


mikejbrown:
Now sprites are broken... putting on [WIP] until I get it fixed.

koosemose:
Looks Good. Merged.

PR #1430, opened 2016-09-20
LocalizationDownloader fix

### LINKED ISSUE #1428: Stop localization downloader from clobbering en_US defaults
From #1418 by @QuiZr:
> It will break stuff: ClearLocalizationDirectory method - after fresh pr in localization repo (which @quill18 still didn't move to the TeamPorcupine account) en_US.lang will be deleted and replaced by nothing.

QuiZr:
Closes #1428

### DIFF
--- a/Assets/Scripts/Localization/LocalizationDownloader.cs
+++ b/Assets/Scripts/Localization/LocalizationDownloader.cs
@@ -86,9 +86,9 @@ namespace ProjectPorcupine.Localization
             {
                 latestCommitHash = GetHashOfLastCommitFromAPIResponse(versionChecker.text);
             }
-            catch
+            catch (Exception e)
             {
-                yield break;
+                Debug.ULogErrorChannel("LocalizationDownloader", e.Message);
             }
 
             if (latestCommitHash != currentLocalizationVersion)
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Localization/LocalizationDownloader.cs
```diff
@@ -226,9 +226,9 @@ private static void OnDownloadLocalizationComplete(Action onLocalizationDownload
                 // called ProjectPorcupineLocalization-*branch name*. Now we need to move all files from that directory
                 // to Application.streamingAssetsPath/Localization.
                 FileInfo[] fileInfo = localizationFolderInfo.GetFiles();
-                if (fileInfo.Length > 0)
+                if (fileInfo.Length > 2)
```
mikejbrown: Should explicitly check filenames instead??

sboigelot: && fileInfo.All(fi => fi.Name.StartWith("en_US.lang")

(this may have issue as I just wrote it in github; but something like that may do the trick)

mikejbrown: Looks reasonable...

QuiZr:
Ok, it's fixed now. It still makes no sense because en_US will be the most outdated localization..

TomMalbran:
The point is that everyone adding something that requires localization should add the localization in the PR. Which means, that en_US should always be up to date, and every other localization can be based from that one.

Assets/Scripts/Localization/LocalizationDownloader.cs
QuiZr: I really hate lambda expressions - they are really unclear to me and every time I see one I need to spend like 5 minutes figuring out what it does - hence I didn't use it.

PR #1438, opened 2016-09-21
Using regions, grouped similar furniture functions together.

kd7uiy:
This makes it a bit easier to find things inside this file!


dusho:
I'm ok with regions (specially to collapse serialization code), but I remember removing them from my PRs because someone didn't like it.. (something about hiding code)

mikejbrown:
That person needs to learn to use their IDE.

sboigelot:
Most people dislike regions (cf: http://programmers.stackexchange.com/questions/53086/are-regions-an-antipattern-or-code-smell )

I personally like them in some cases.

This version would clash with the stylecop rule of ordering private and public methods. What if you had a private method used only for the #region Save Load 
(According to me the problem is the stylecop rule here, I like this kind of regions)

kd7uiy:
@sboigelot It meets all Stylecop rules as it stands now, I'm not quite sure what you mean...

sboigelot:
@kd7uiy For now yes, but imagine you have to add a private method in the following region

```
    ....
    #region Update and Animation
    public void EveryFrameUpdate(float deltaTime)
    {
       DoSomethingPrivate();
        ...
    }

    ...

    private void DoSomethingPrivate()
    {
       
    }
    #endregion
    
    #region Get Status
    public bool IsExit()
    {
        ....
    }

   ...

```

Then the new method "DoSomethingPrivate" would be on top of the public method bellow the endregiond and stylecop wouldn't be happy with our current config

sboigelot:
I think the rule is http://stylecop.soyuz5.com/sa1202.html

kd7uiy:
That to me would be a different PR than this one, but I think that would be a good idea.

dusho:
on the other hand it's kind of cool to have all public methods bundled somewhere.. and not placed randomly in file... I'm really puzzled with this one..

sboigelot:
We shouldn't merge a PR that will break future modification and confuse everyone, if we accept region we have to turn off this rule at the same time.

But this need more input from others probably

vogonistic:
I don't use code folding and find that regions don't do anything for me. 

I think that @sboigelot has a valid point about the ordering of the functions and this will also become another thing for the reviewers to keep track of. Does that function belong to that region or should it be somewhere else.

We already have a rule that the public functions should be before private. If they are ordered in a confusing way we should fix that instead.

dusho:
instead of regions I had partial classes.. where you could have functionalities grouped nicely (serialization, updates, helpers, ...) and still follow this StyleCop about order of methods.. but partial classes were not liked either
Is there any other way to group things by functionality?

TomMalbran:
I also don't use folding, nor for functions and will not use them for regions. They do have valid points that is not always easy to grupo functions, and having regions for just 1 function seems silly, if we decide that all functions should be contained inside a region.

I sometimes don't like the ordering of having public functions first and privates later. For example when you want to split a public function which will call private functions, then those other functions will most likely not be closer to the public one. But in other cases is good.

To split functionality we can use components, classes which will hold part of the functionality and the main class will call it, or outside clases can call it to. This components can also be shared by more than one class, so that we don't need to repeat ourselves too much. We are already going in this path too.

kd7uiy:
I also don't like the ordering of public functions first, especially if we do something like allow for using regions. Splitting classes in to smaller classes is often, but not always, a good idea, and should be done on a case by case basis.

koosemose:
Conflicted

kd7uiy:
Doesn't surprise me, any changes to Furniture are almost certain to conflict with this, so... I can't believe it lasted so long without conflicting!

bjubes:
Merging for now. When regions conflict with stylecop ordering, well face the problem at that point

PR #1440, opened 2016-09-22
Dialog logic for Save, Load, and implementation of New World in Option

jaselito:
Implemented the New World option in the Options dialog. In order to do this I replaced the AreYouSure dialog with a general purpose PromptOrInfoDialog, which can be used to show simple messages with an OK button or to prompt the user to validate actions, with Yes, No, and/or Cancel buttons.  Also introduced logic to allow multiple dialogs to be shown at the same time.

I went ahead and implemented dialog logic for the Save and Load options and changed their skins to match our UI style. 

Here a couple screenshots:

new world
save

This will introduce a couple of keys not found in the localization file... I left it like that since I don't know if I can edit the file myself or what the current deal with localization is.

Also, some times when opening the save or load dialogs I've found this

list bug

This disappears when refreshing the dialog (for instance moving the scroll bar) but I haven't been able to fix it in code other than in a hacky way, so right now I left it like that in case someone knows the proper fix.


TomMalbran:
a couple of design issues:

**Prompt**
1. I think they should all the have the same width, which could be changed in code. The width should be less than than the new world one, and more than the overwrite one.
2. There should be more padding around the text + buttons, and more space between the text and the buttons.
3. The height of the buttons should be the same as the height of the buttons in the options menu.
4. The buttons could be a little bit wider and occupy the entire width.

**Save**
1. I think that it should use the same design as the options menu. Might be debatable if it doesn't look as good.
2. The dialog could be wider and taller. The bottom buttons and input will look better.
3. Maybe we should close the options menu and open the save menu, and when done open the options menu again, so that there aren't so many dialogs. Or we might need to add a backdrop.
4. There should be more padding. Similar to the settings menu.
5. We could add a bit more space between the list and the buttons/text input
6. Each line in list could be spaced a bit more.
7. The list should have a transparent background (not white), and each even row could have a background color similar to the buttons one

We have room, we can have bigger dialogs with more spacing.
This still looks great and is a great improvement over the old dialogs.

Tranberry:
Text should force the box to be taller not wider, and if the text don't fit it is probably not done in a UX friendly way. But I like that things are moving forward, good work!

jaselito:
Thanks for your feedback!  I'll try to implement these changes.  Right now the message box expands to accommodate the text as is given in the string.  So it will adapt to the width of the longest line and will expand in height if there are multiple lines. I'll see if I can give it a constant width and make the text wrap then.

jaselito:
Ok, suggestions implemented.  Here a preview of the current way things work:

screen

As you can see towards the end of the animated gif, right now if you call the save dialog as a result of choosing to save before loading a new world, the options dialog remains open in the background. I agree with @TomMalbran that it'd be better if it were closed, but if I do that, then the rest of the logic in the New World button function doesn't get called after saving (its component being deactivated due to the closed dialog), and the new world is not created.  Any ideas?  

Also, that strange bug with the file list in the save dialog is still there, and goes away when I scroll the list. Anyone knows how to fix this in a proper way?

I also created localization keys for all prompts/messages. If anyone knows a way to add newline characters to the localization file I'd be all ears (if I just use \n unity doesn't recognize it as an escape sequence and just prints it as \n).

TomMalbran:
The dialogs look much better. I still didn't liked the files list, so I tried a few options. I think this one is the best:
screen shot 2016-09-23 at 1 27 21 am

Basically I change in FileListItem.prefab the sprite for none to get rid of the rounded corners and in `DialogBoxLoadSaveGame` the colors for:
```
public static readonly Color PrimaryColor = new Color32(0, 149, 217, 80);
public static readonly Color SecondaryColor = new Color32(0, 149, 217, 160);
```
It could still be improved, and maybe the active color could be less dark. And later we need to change the delete button for a white x. But we are going in the right direction.

I also don't like the elastic scroll, it bounces too much. I think we removed it from the Construction Menu, maybe we can remove it here too?

jaselito:
Hmmm I do like your colors better but I'm not quite getting the same as in your screenshot... weird, our blues look different... 
Anyway, I will implement your suggestions;  I like both the rounded and the square corners, but square should actually fit better with the overall design

TomMalbran:
Did you copy the full code? I changed Color to Color32, since I could copy the colors from Photoshop.

For this design I think that no rounded corners are better, and more in a list.

jaselito:
Yeah, I copied your lines as they are... but also the blue in the dialog background is different, just look at the screencaps...

TomMalbran:
Is the same background as in the save dialog from the first part of the Gif. Not sure why the colors are different.

I tried to get the same blue as the button, but I don't know which opacity is used, so tried with some. We can go for a closer color too.

jaselito:
The colors you suggested are working for me.

Btw, are you getting the same bug with the list when opening the save dialog after clicking on New World? (it doesn't always happen, maybe try a couple of times)

TomMalbran:
No. Tried
"New World -> Dont save -> Options -> Load"
"New World -> Dont save -> Options -> Save"
"New World -> Yes to save"

And the dialog was always fine. But I tested in my modified version with no elastic scroll.

jaselito:
It keeps happening on my side. Only with "New World -> Yes to save".  I also removed elasticity and it still happens. Weird... 
I can try to force a small scroll on opening to get rid of it but that feels like a dirty fix...

koosemose:
I definitely also get the issue with everything being crammed into one line, on  "New World -> Yes to save".

There's also an issue where, if you delete a save, the sizes don't get recalculated (but that's a preexisting issue)

TomMalbran:
I tested the error. If I have 7 files, which means that I don't need to scroll I get it all the time. If I have 8, it works fine.

I had another weird issue. For some reason I have a file with no name showing in the list (not sure why). If I try to delete it I can't. But then when I try to delete another file I get 3 buttons yes, no, okay. The last issue might have something to do with no resetting the buttons. The other 2 are separate from this PR.

jaselito:
@koosemose Yeah it seems to me as if the layout weren't being updated before being shown...

@TomMalbran Hmmmm I put a check in to avoid saving a file with no name.  Yours must be an old save then.  Anyway, I removed the check to generate such a file and couldn't replicate the problem... could you upload the file somewhere so I can check it?  What I think is happening is that a prompt is being shown before an info box gets closed, so all the information gets combined (the dialog gets reset on close).

TomMalbran:
@jaselito I just saw the image you posted with the new dialogs colors, it might have not loaded before or something. All the dialog colors are slightly different, but they look fine.

The issue with the blank file is because it was finding .DS_Store as a file and showing it. It should check if the file has a .sav extension. So I couldn't delete that file, which could show an error message, which I didn't saw. And then it messed the buttons of the next dialog, because I think that setting the buttons in a bitwise way can easily cause issues. So we probably should create a new instance of the Prompt dialog for each new confirm/message dialog, so that we can also display more than 1, and second, it might be better to set the buttons with a different param for each.

The other issue with the list is weird, it might have to do with the way you are showing the dialogs, or something like that.

jaselito:
Will see about making each dialog a new instance.  I think if we go that way we can leave setting the buttons bitwise; I feel it's much more intuitive that way.

I also think I fixed the list bug (both for showing the save dialog and for the load dialog after deleting a file).  Please test and confirm if you can. Mind you, I still don't know what the actual cause was but the addition I did seems reasonable to have anyway.

TomMalbran:
Is either making a new instance or maybe changing the set buttons, and in another PR we can make it a single instance.

I actually think that settings the buttons as
```dialogBoxPromptOrInfo.SetButtons(DialogBoxResult.Yes, DialogBoxResult.No);```
Is much more intuitive, as each button is just a param. And the code can be easier, since you can cycle through the params and show the required buttons. I think that bitwise operations makes the code harder to read.

**Comment by jaselito** [comment]
I see what you mean... I'm still pretty green and didn't think of using an array of params, thought you were referring to having a list of bools with default values in the arguments.  You're right, will try to do it like that.

**Comment by jaselito** [comment]
Ok, changed to list only .sav files and to set buttons with a params array.

File: Assets/Scripts/UI/DialogBox/PromptOrInfo/DialogBoxPromptOrInfo.cs [hunk 1]
```diff
@@ -28,22 +28,27 @@ public class DialogBoxPromptOrInfo : DialogBox
...
+        Transform buttons = gameObject.transform.Find("Buttons");
+        buttons.gameObject.SetActive(true);
 
-        if ((buttonsToSet & DialogBoxResult.Yes) == DialogBoxResult.Yes)
+        foreach (Transform button in buttons)
+        {
+            button.gameObject.SetActive(false);
+        }
+
+        if (Array.Exists(buttonsToSet, element => element == DialogBoxResult.Yes))
```
TomMalbran: This could be simplified a bit if you had a Dictionary of DialogBoxResult to GameObject. You can then do a single loop and set the active to true.

jaselito: Since regrettably Unity doesn't serialize dictionaries in the inspector, in order to do this I would have to set up each entry for the dictionary manually in a start method. I could do it if you're dead set on it, but it seems ugly...

jaselito:
I added a check to save before loading a game... please test.

koosemose:
Everything looks good, the previous issues seem to be absent, and things work as expected, no stylecop issues. Merged.

PR #1443, opened 2016-09-23
[Art] Walls and floors (wall tileset by Yurgeta and Tranberry - TRBRY)

dusho:
Maybe it's time for new walls and floors? This wall tileset was lurking in discord for some time. Originally made by Yurgeta, tiled and adjusted by Tranberry - TRBRY, further adjusted by me.
Also changed the floors to match walls better (also with dirt tiles)
image
image
image


dusho:
fixed the north ending of tile
image

TomMalbran:
This looks great. I want the new walls and floors.

Tranberry:
I still think that the floors needs to be brought to the space age, the concrete look is odd.

TomMalbran:
@Tranberry Make us some better floors. hehe. I agree that more spacey floors will look good. But this floors look better with the new walls than the old grey ones.

Raketfart:
:+1:

frankitox16:
I really like it, but at the same time it feels like it contrasts a lot with the current UI colors... 
Great job, btw!

Tranberry:
@frankitox16 I don't see that as a problem. Contrast is good sometimes and bad other times.

koosemose:
You really need to stick to one Sprite (or sprite set in the case of something like floors) per PR, it makes it extremely difficult to gauge rather or not replacement sprites should be merged, when we have to look at differing preferences along 2 or more sprites.

TomMalbran:
For this PR I disagree. The Walls with the grey floor look bad. And the issue is not the walls but the floor. We probably don't need to replace the floors, but there are some issue with new types of floors. And we can always change the floor later or add more.

dusho:
ye.. I added floors to this PR just for sake of some consistent look. These new walls and old floors (we had) looked really bad,,

dusho:
so what's with this.. should I revert the tiles to old ones?
this is how it looked like with old tiles:
image

dusho:
old tiles are back in, is this PR more acceptable now?
image

dusho:
image

PR #1445, opened 2016-09-23
[Small] Better Room Splitting and Joining

koosemose:
This fixes a problem with old rooms not being deleted when rooms are joined (Rooms would stay in RoomManager).

It also fixes a problem when splitting rooms where the gas pressure would be directly copied to the new room, effectively causing a loss of total gasses in the rooms (Gas Pressure now goes up a slight amount when rooms are split, due to more of the space the gas would otherwise be in being occupied by a wall, this matches the behavior where the gas pressure drops when a wall is removed.)

Also renamed CopyGasPreasure to SplitGas, to more accurately reflect what is happening.

### DIFF
--- a/Assets/Scripts/Models/Area/Room.cs
+++ b/Assets/Scripts/Models/Area/Room.cs
@@ -356,11 +356,11 @@ namespace ProjectPorcupine.Rooms
             }
         }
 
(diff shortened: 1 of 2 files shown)


koosemose:
This being a small PR that mostly just makes a minor improvement to systems already in place, I'm self-merging.

PR #1447, opened 2016-09-23
[Fix] Cloning pod was broken after refactoring

dusho:
when cloning pod was placed it was throwing error because of properties moving to different classes and some methods being properties now


TomMalbran:
Sorry. I broke it in a PR. Probably after a merge conflict fix done wrong.

I verified that this fixes the issues.

PR #1450, opened 2016-09-25
[Tiny] Fix for empty localization key in construction menu

jaselito:
As seen on #1425


koosemose:
Nothing seems to go terribly wrong, stylecop clean. Merged

PR #1453, opened 2016-09-26
Stylecop fixes

koosemose:


koosemose:
Just a few stylecop fixes missed. Merging

PR #1458, opened 2016-09-27
[Feedback]Furniture components (with workshop as component)

dusho:
Made some basic component implementation (abstract class Component) that will be used as parent for all components.. I refactored Workshop to be a component, seems to be working fine so far.
Current BuildableComponent methods are:
```cs
public void Initialize(Furniture parentFurniture)
        {
            ParentFurniture = parentFurniture;
            Initialize();
        }
public abstract void Initialize()
public virtual void FixedFrequencyUpdate(float deltaTime)
public virtual void EveryFrameUpdate(float deltaTime)
public virtual List<ContextMenuAction> GetContextMenu()
public virtual string GetDescription()
```
more can be added when needed.
Feel free to test it and maybe give some feedback..


koosemose:
Conflicted

koosemose:
Also added [Feedback] to the title, since you requested feedback in the text. Hopefully having it in the title will draw more feedback (also added the label, just to try to draw more attention)

dusho:
- updated with generic way to detect components in code (classes with ComponentName attribute in assembly)
- Initialize() moved to proper location so components can hook on furniture events (Removed, ... )
- when deconstructing or changing production Workshop component will now unlock all the inventory left in input (there are some issues due to jobs still not having proper priorities)
- resolved conflicts

Assets/Scripts/Models/Buildable/Components/Component.cs
```diff
@@ -0,0 +1,119 @@
...
+        private static Dictionary<string, Type> componentTypes;
+        
+        public Component()
+        {
+            // need to set it, for some reason GetHashCode is called during serialization (when Name is still null)
+            Type = string.Empty;
+        }
+        
+        [XmlIgnore]
+        public Furniture ParentFurniture { get; set; }
```
TomMalbran: I think that every component should link to a Buildable, so that we can use some components for both Furniture, Utility and other buildable clases. And we should replace places where Furniture is used (as comments, part of Function names...) to Buildable or just remove it.

Assets/Scripts/Models/Buildable/Components/Component.cs
```diff
@@ -0,0 +1,119 @@
...
+                Debug.ULogErrorChannel(ComponentLogChannel, "There is no deserializer for component '{0}'", componentTypeName);
+                return null;
+            }
+        }
+
+        public virtual void Initialize()
+        {
+        }
+
+        public virtual void FixedFrequencyUpdate(float deltaTime)
```
TomMalbran: A component should also allow to override the EveryFrameUpdate.

Assets/Scripts/Models/Buildable/Components/Workshop.cs
```diff
@@ -0,0 +1,430 @@
...
+                HaulingJobForInputs(prodChain);
+            }
+        }
+
+        public override List<ComponentContextMenu> GetContextMenu()
+        {
+            return WorkshopMenuActions;
+        }
+        
+        private static void PlaceInventories(List<TileObjectTypeAmount> outPlacement)
```
TomMalbran: Not sure why there a private static functions...

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -47,11 +47,8 @@ public class Furniture : IXmlSerializable, ISelectable, IPrototypable, IContextA
     /// These context menu lua action are used to build the context menu of the furniture.
     /// </summary>
     private List<ContextMenuLuaAction> contextMenuLuaActions;
-
-    /// <summary>
-    /// Workshop reference if furniture is consumer/producer (not null).
-    /// </summary>
-    private FurnitureWorkshop workshop;
+    
+    private HashSet<ProjectPorcupine.Buildable.Components.Component> components;
```
TomMalbran: Would be nicer to at least have ```using ProjectPorcupine.Buildable```, so that we can reduce this lines.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -390,12 +381,14 @@ public static Furniture PlaceInstance(Furniture proto, Tile tile)
...
         obj.Tile = tile;
-        if (obj.IsWorkshop)
+
+        // need to update reference to furniture and call Initialize (so components can place hooks on events there)
+        foreach (var comp in obj.components)
         {
-            // need to update reference to furniture for workshop (is there a nicer way?)
-            obj.workshop.SetParentFurniture(obj);
+            comp.ParentFurniture = obj;
+            comp.Initialize();
```
TomMalbran: Should we pass the furniture to ```Initialize```, I don't like that ```ParentFurniture``` is public.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -487,11 +480,14 @@ public void FixedFrequencyUpdate(float deltaTime)
             EventActions.Trigger("OnUpdate", this, deltaTime);
         }
 
-        if (IsWorkshop && IsBeingDestroyed == false)
+        if (!IsBeingDestroyed)
```
TomMalbran: I think that we should probably pass ```IsBeingDestroyed``` to ```FixedFrequencyUpdate()```, or check inside that function since it can access the Furniture. In the future we will have components that will still run after the furniture is set to be destroyed, for example, a component that handles doors.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -1211,13 +1219,13 @@ private bool DefaultIsValidPosition(Tile tile)
     }
 
     #region Private Context Menu
-    private ContextMenuAction CreateWorkshopContextMenuItem(WorkshopContextMenu factoryContextMenuAction)
+    private ContextMenuAction CreateWorkshopContextMenuItem(ComponentContextMenu componentContextMenuAction)
```
TomMalbran: We should rename this method. should it be part of the component?

Assets/Scripts/Models/Buildable/Components/Component.cs
dusho: I agree.. will try to do it as part of other PR as we then need to move events from Furniture to IBuildable so components can hook on those.. also maybe more changes would be required in other classes (Tile.cs) ?
```cs
public Furniture Furniture { get; private set; }
```

dusho:
incorporated comments from @TomMalbran (thanks for great review)

koosemose:
Looks good, Stylecop Clean. Merged.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -120,7 +119,9 @@ private Furniture(Furniture other)
 
         Parameters = new Parameter(other.Parameters);
         Jobs = new BuildableJobs(this, other.Jobs);
-        workshop = other.workshop; // don't need to clone here, as all are prototype things (not changing)
+
+        // don't need to clone here, as all are prototype things (not changing)
```
TomMalbran: We shouldn't assume this. Once we make a component that will have params that change, then this will not be true.

dusho: Actually I would propose to use Params as only thing that can change.. this would be a generic way how to access parameters from other components, easy to mod, and would also solve the persistency (save/load game). So basically no more prototype vs. instance.. only difference will be that instance will have 'live' Params

PR #1462, opened 2016-09-27
[Awaiting #1487] Fire 2.0

bjubes:
_For now this PR is closed. The work is not abandoned, but this should not be implemented until the effects system in #1487 is created or a consensus reached._

Revives #962. I have already started moving and updating code to work with new systems since the original PR was made. For now I will only make enough changes to re implement fire in a working state similar to the original PR.
Because I am updating old code, there is a very high chance something will slip through that is interacting with the new systems incorrectly. Please point these out!

Todo:
- [x] reimplement FilippoLeon's work into the new master
- [x] upgrade to new animation system
- [x] Fix lag spikes
- [ ] make fire not show up in the construction menu (tell me if you know how)
- [ ] Make event spawn in base, not in any random tile in the world
- [ ] make it so when fire is selected, its description shows its strength

and more that anyone thinks of. What will come after this PR

-  flammability of furniture
- a separate "status" layer on a tile where fire exists, so it can coexist on a tile with a furniture instead of replacing it.
- characters interacting with fire


TomMalbran:
I think we should consider having fire as instance of an Effect/Decoration class or similar, which would be a class just like Furniture and Utility, but in a way that we can have in the same tile 1 or multiple Effects, multiple Utilities and a Furniture. Using furniture for fire doesn't allow us to do that. And Fire acts really different than any furniture.

bjubes:
I agree. I dont think a furniture can even be "on fire" in this system since you can't have multiple furniture on a tile

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -693,5 +693,87 @@ function Rtg_UpdateTemperature( furniture, deltaTime)
...
+
+	-- Consume o2
+	if furniture.Tile.Room != nil then 
+	    furniture.Tile.Room.ChangeGas("O2", -0.00001)
+	    furniture.Tile.Room.ChangeGas("CO2", 0.00001)
+	end
+	
+	local spreadProbability = furniture.Parameters["spread_chance"].ToFloat() 
+	
+	-- Spreads only by chance. the lower the strength, the less the canche
```
Geoffrotism: -- Spreads only by chance. the lower the strength, the less the chance

Assets/StreamingAssets/LUA/Furniture.lua
```diff
@@ -693,5 +693,87 @@ function Rtg_UpdateTemperature( furniture, deltaTime)
...
+-- Deregister the temperature update function
+function Fire_UninstallAction( furniture, deltaTime)
+	furniture.EventActions.Deregister("OnUpdateTemperature", "Fire_UpdateTemperature")
+	World.Current.temperature.DeregisterSinkOrSource(furniture)
+end
+
+-- Increase temperature of current tile
+function Fire_UpdateTemperature( furniture, deltaTime)
+	-- Magic numbers for temperature
+	World.Current.temperature.SetTemperature(furniture.Tile.X, furniture.Tile.Y, 700 + furniture.Parameters["strength"].ToFloat() * 100)
```
Geoffrotism: does this line actually work? it should be requesting a Z axis value.

bjubes: Probably not. I only got to making the actual implementation of the fire "furniture" work again in terms of the new furniture system. I haven't gone into the Lua yet

Tranberry:
can you write a shirt description of the sprite, the XML gave me no clearness. I'm guessing it's some variants on how the small to large flames look?

bjubes:
here is the fire png in the diff https://github.com/TeamPorcupine/ProjectPorcupine/pull/1462/files#diff-754755f6e5ddfbf63fdb1f3550e54684
Between the animation inconsistency and the implementation of fire as furniture, i'm debating abandoning the original PR and implementing this from scratch

Tranberry:
yes I saw the sprite sheet, but I wanted to know what each things was, is there something more than just variants of the flames?

bjubes:
each column is a terrible animation. Each row represents a progressivly larger fire size. There is no real number of "sizes", 8 is an arbitrary number for now

Geoffrotism:
Just some thoughts. havent looked at new commits.
- add a check to make sure the fire event doesnt spawn in outer space. 
- resistance to fire/other for certain objects. (instead of more info in the furniture.xml, would it be better to make a table look up?)
- since fire is a furniture, this implies it cant be on a tile that other objects are. this means that fires can NOT pass from room to room.

bjubes:
@Geoffrotism 
1. I agree it needs to spawn in the base, maybe choose a room to spawn in so we know its relavant to the player not an asteroid 100 tiles away catching on fire?
2. Do you mean a seperate xml file for flammability? I think putting flammability in furniture.XML makes the most sense, but I agree its crowded. Probably wont add it in this PR thou since it adds a lot of complexity and touches every furniture.
3. @TomMalbran mentioned this, I think it eventually needs to exist as a "status" layer that is seperate from furniture, so an area can catch on fire, flood, fog over, that sort of thing. Again that is a huge change and needs to discuss that separately before doing so. I agree that fire as a furniture is hacky and dumb, but thats how Im gonna leave it for now until we have a system for dealing with effects on a tile that are not furniture. Hopefully said system will have a "status" class that inherits from furniture so it can be defined the same way and exist how furniture exists, just "on top" of other furniture instead of replacing it.

TomMalbran:
I think that it makes no sense to add this as a furniture, and I think that we shouldn't merge it as such. We can make a PR that is not as big that just adds a new type of "Buildable" called Effect or something like that, and the Fire implementation. The Effect class can start as a copy of Furniture or Utility were we remove everything that is not needed for effects, a manager that can be a copy of the UtilityManager, and a reference in tiles.

We can later add another PR that can deal better with fire and other effects, if that is required.

At some point we can grab the buildables, and maybe their managers and think about using subclassing.

bjubes:
The more I iwork on this the more I agree with you. Ill hold off this PR until I or someone implements an effect layer. That brings up a lot of questions so It will have to be its own issue for discussion and PR.

Tranberry:
so making another fire spritesheet for now is not needed, right?

bjubes:
no i presume the whole effects thing will take a while

PR #1467, opened 2016-09-28
[Feedback] 3d pathfinding

koosemose:
It's not absolutely perfect, but Crew can now navigate Z-Levels, and will work tiles directly above or below them. 

Thanks to @vogonistic and pderuiter(on discord) for the assistance that lead to getting this working.


koosemose:
As this is functionality that is required for Z-levels to mean anything, I am going ahead and merging this. The functionality certainly needs improvements, and it will require pathfinding to have some optimization, as adding a 3rd dimension (quintupling the amount of space searched with just the relatively small 5 levels we currently have).

PR #1468, opened 2016-09-29
Main Menu Manager

TomMalbran:
This is a manager for the bottom main menu. The idea is so that we can add items to the main menu without having to modify the Main Menu controller. In a way it makes the main menu code driven (but using prefabs for the buttons), since the items can be added through code and it will be really easy to add mod support from a lua script if we go in that route.


Assets/Scripts/UI/InGameUI/MainMenu/MainMenuManager.cs
```diff
@@ -0,0 +1,161 @@
...
+        position = MathUtilities.Clamp(position, 0, menuItems.Count);
+        AddMenuItem(new MainMenuItem(key, callback), position);
+    }
+
+    /// <summary>
+    /// Adds a new menu item behind the menu item with the given key.
+    /// </summary>
+    /// <param name="key">The menu item key.</param>
+    /// <param name="callback">The menu item callback. Called when the menu item is clicked.</param>
+    /// <param name="afterKey">PLace the given menu item after a menu item with this key.</param>
```
sboigelot: type PL -> Pl

koosemose:
I'm not sure MainMenu is the best to describe the bottom menu. For me at least, Main Menu sounds like it either refers to the menu that would be on the opening screen, or the menu that is currently labeled as "options" I think (I think for the most part those two should share the same menu items... except for save of course, but that's separate from this.)

TomMalbran:
@koosemose What would you suggest as a name? Bottom menu is a really bad name, since it doesn't describe the menu, and it could easily be moved to the top, and then the name makes no sense. I went for Main menu because is the menu from where everything else is open, is like the top menu in any application, which is the principal menu.

The opening scene is more of an intro scene with an intro menu, than just a menu. The options dialog is an options menu (I am thinking on doing something similar in that menu to make it extensible).

I don't mind changing the name, but I don't know what else to call it. I am open for suggestions.

dusho:
RibbonMenu ?

MANDAL0R3:
UIMenu ?

koosemose:
To me Options sounds like just another name for Settings. And one of the reasons I'd refer to the intro screen (or at least the menu on it) the main menu, is the number of games with the option of "Return to Main Menu" which takes you to their equivalent of that screen.

As far as names, dusho's suggestion of RibbonMenu might work, or just MenuBar (though that inverts the word order which may be undesirable), perhaps PlayMenu (or GameplayMenu) since pretty much all of the menus on it are involved with actually playing the game (aside from the "options menu").

On a secondary note, would it be possible to force the options menu to always be in a particular location (I'd suggest the far right), from a UI design standpoint, it seems like that sort of menu should always be in the same location, no matter what other buttons/menus mods add to the bar. Though of course you may already do this, I haven't had the opportunity to test this yet...

TomMalbran:
* I don't think that RibbonMenu is a good name, as it refers more to a tab style menu, and we use it to open dialogs and other UI elements.
* MenuBar would be fine as in a sense it works just like the MenuBar of any program, but is at the bottom of the screen.
* PlayMenu doesn't sound that good. I would prefer GameMenu or GameplayMenu. I like GameMenu better.

Each menu item can be added at certain index or after an element. I was thinking of trying something like an option like MenuPosition.Last and MenuPosition.First and then sort them each time a menu item is added. But then, if I had that I might have to also sort items that are placed after another one, and that would make it a lot harder to do. So I didn't went with that option.

I think that we should not allow to add menu items at certain positions. If someone adds a mod with a menu more important than construction, or that i replaces it (we could have an option to remove menu items), he could add it first, and same for options. If you want to add a menu item just add it after a menu before the options menu. It is also possible to add an option to add a menu before another. Right now you can only add it after a menu item.

bjubes:
I think gameMenu is the best so far

pderuiter77:
If we make it a bit more generic, couldn't we use this as a controller for each type of menu we want?
Like the main menu, the pause menu, the Game Action Toolbar (my suggestion for what is now called main menu), maybe even for the context menu

Assets/Scripts/UI/InGameUI/MainMenu/MainMenuController.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        menuLeft = GameObject.Find("MenuLeft").GetComponent<MenuLeft>();
+        buttonPrefab = Resources.Load("UI/MenuButton");
+
+        CreateMainManu();
+
+        MainMenuManager.Instance.Added += CreateMainManu;
+    }
+
+    private void CreateMainManu()
```
dusho: typo, Manu -> Menu

TomMalbran:
I renamed the MainMenu to GameMenu, since it was the game I liked most.

What do you all think about adding more ways to add items to the list. You can now add at certain position or after certain element. We could add a way to add before an element, o to make certain items be always first or last (which I am not sure is that good).

@pderuiter77 The GameMenuManager is pretty generic. It could easily became a base class and have other menus extend it, since it is just data. The GameMenuController would need some work, but I am not sure if it needs to be generic. A different controller will be required for a different menu, since the prefabs and other things change. But it could be generalized.

The other issue is that we don't have other menus like this one yet. We need one for time controls, but it will not really need a manager. The options menu is part of a dialog, and I am not sure if we want to make it from XML or with an OptionsMenuManager. The Context menu works really different.

bjubes:
when test this the game menu is missing. Is this still WIP?

TomMalbran:
@bjubes No. I just forgot to save the prefab changes after the rename. Should be fixed now.

PR #1469, opened 2016-09-29
Game Events Manager Update

### LINKED ISSUE #1451: Time for GameEvents continues regardless of game speed or being paused
As it stands GameEvents is a monobehavior attached to the Controllers gameobject, so it gets its update from that, rather than TimeManager, which would prevent it from continuing to spawn events when paused.

TomMalbran:
The Game Events now use the Prototype Manager and the Game Events Manager is now an instanced created in World and updated through World.

Fixes #1451.

### DIFF
--- a/Assets/Scripts/Models/Area/World.cs
+++ b/Assets/Scripts/Models/Area/World.cs
@@ -143,6 +143,12 @@ public class World : IXmlSerializable
     public RoomManager RoomManager { get; private set; }
 
     /// <summary>
+    /// Gets the game event manager.
+    /// </summary>
+    /// <value>The game event manager.</value>
+    public GameEventManager GameEventManager { get; private set; }
+
+    /// <summary>
     /// Gets the ship manager.
     /// </summary>
     /// <value>The ship manager.</value>
@@ -411,16 +417,16 @@ public class World : IXmlSerializable
         FurnitureManager = new FurnitureManager();
         FurnitureManager.Created += OnFurnitureCreated;
 
-        ShipManager = new ShipManager();
-
         UtilityManager = new UtilityManager();
         CharacterManager = new CharacterManager();
         InventoryManager = new InventoryManager();
         jobQueue = new JobQueue();
-        CameraData = new CameraData();
-        PowerNetwork = new ProjectPorcupine.PowerNetwork.PowerNetwork();
+        GameEventManager = new GameEventManager();
+        PowerNetwork = new PowerNetwork();
         temperature = new Temperature();
+        ShipManager = new ShipManager();
         Wallet = new Wallet();
+        CameraData = new CameraData();
 
         LoadSkybox();
         AddEventListeners();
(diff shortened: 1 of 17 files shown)


koosemose:
Looks like you've had failures in the Scheduler tests, you'll need to take a look and see if either the tests need changed if something you've changed has altered what their expectations should be, or see if something in your code needs changed to match expectations.

TomMalbran:
Woops. I change the Getter name in PrototypeManager from SchedulerEvent to ScheduledEvent, since that is the name of the class. Will fix it.

Assets/Scripts/Models/Area/World.cs
```diff
@@ -478,6 +484,7 @@ private void TickEveryFrame(float deltaTime)
     {
         CharacterManager.Update(deltaTime);
         FurnitureManager.TickEveryFrame(deltaTime);
+        GameEventManager.Update(deltaTime);
         ShipManager.Update(deltaTime);
```
koosemose: Do we need to update the GameEvents every frame? As far as I know they're not the sort of thing that needs that degree of granularity, they're just supposed to either just do it (in the case of spawning crew) or start the events going (such as maybe a pirate attack or a trader visiting)

Assets/Scripts/Models/Events/GameEventManager.cs
```diff
@@ -7,123 +7,32 @@
...
-        }
-    }
-
-    private void ReadEventFromXml(XmlReader reader)
+    /// <summary>
+    /// Update the current game events.
+    /// </summary>
+    /// <param name="deltaTIme">Delta time.</param>
+    public void Update(float deltaTIme)
     {
```
koosemose: Capitalization :)

Assets/Scripts/Models/Area/World.cs
dusho: can we somehow start unifying the Update methods? it's getting messy a bit, also it would be good if it can start with UpdateXXX.. UpdateEveryFrame.. UpdateFixedFrequency, .. ?

TomMalbran: It was updated every frame before. So I went with that. We can always update less, but not sure how good it will work.

@dusho You can do that in another PR.

Assets/Scripts/Models/Events/GameEventManager.cs
TomMalbran: Fixed.

Assets/Scripts/Models/Area/World.cs
TomMalbran: @koosemose This system needs to get expanded. So here I just made it so that it worked exactly the same as in master. We should probably discuss this further in another issue.

In a way an event shouldn't have a repetition. There should be a story teller module that will have a list of possible events with information about the amount and conditions as to when the events can occur. At that point the GameEventManager could keep track of 2 loops, one slow to check for upcoming events, and maybe a fast one for the current events, if the event needs it. And once the event ends, it gets removed from the list. For example some sort of meteor shower could be on a fast update so that the animation is handled better, and at the same time could be in the slow update to check for when it will end or other less important things.

PR #1473, opened 2016-10-01
[Wiki] New screenshot (0.1-alpha)

dusho:
new recent screenshot from master for title page..
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 Project Porcupine: A Base-Building Game...in Space!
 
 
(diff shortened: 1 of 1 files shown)


Tranberry:
Hurray for a new screen shot! But why does it say wiki?

dusho:
ehm.. not sure.. just to differentiate that it's not code change.. but only .md change
is there different prefix for that?

bjubes:
No prefix nessicary. If you want you can say "add picture to readme" in the title

PR #1475, opened 2016-10-01
Add RTL localization support

### LINKED ISSUE #1456: Some Languages Require RTL
Hey! I was going to correct the he_IL (Hebrew) translation - and then I noticed - that there is a problem. A few languages like Hebrew, and arabic, write from right to left. So for some reason, unity displays - siht ekil (like this). I'm not that good of a programer - but I think there should be someway to specify RTL or LTR. Until then hebrew and arabic are basicly meaningless.

bjubes:
closes #1456. This implementation is far from perfect, so please leave suggestions.

RTL languages such as hebrew now display correctly from right to left. This is done by inverting the string, which is bad practice since the string should be correct internally, but the UI should invert it when displaying.

to make a language RTL, the first line of the localization file must contain:
`rtl=true` or `rtl=1`

The following files will be changed RTL on the localization repo after this is merged (please add more if applicable):
* he_IL
* ar


Assets/Scripts/Localization/LocalizationTable.cs
```diff
@@ -156,5 +170,17 @@ private static string GetLocalization(string key, FallbackMode fallbackMode, str
                     return string.Empty;
             }
         }
+
+        /// <summary>
+        /// Reverses the order of characters in a string. Used for Right to Left languages, since UI doesn't do so automatically.
+        /// </summary>
+        /// <param name="original">The original and correct RTL text.</param>
+        /// <returns>The string with the order of the characters reversed.</returns>
+        private static string ReverseString(string original)
```
mikejbrown: There's no builtin for this??? :scream:

bjubes: not according to google

mikejbrown: That is genuinely surprising.

bjubes: we could make an extension of string, but thats kinda overkill IMO

koosemose:
Just noting the bug discovered on Discord here. 

This fails with any localization string using parameters (`selection_hp={0} / {1}`), though none of those currently in Hebrew are used in-game.

mikejbrown:
Should we:
1. Merge this anyway and make an issue for the bug, or
2. Bump this off of v0.1, or
3. Delay v0.1 until there is a fix?

bjubes:
Id go with bump, I probably can't implement a fix until tommorow the earliest

mikejbrown:
Done.

bjubes:
My plan is to parse the string so that { and } are found and kept as is during a RTL string reverse. as such this will no longer be a "small" fix

mikejbrown:
It should be specifically {number}. I imagine { and } could legitimately appear in a string in other contexts.

bjubes:
i would presume that anything inside brackets is a placeholder varable and thus a number but technically yes ill check for a number. bring on the regex

bjubes:
@mikejbrown fixed the parsing error for `{#}` in strings. Now its apparent that a lot of the UI skips the localization table, such as the mouse over dialog, so only some strings are reversed. Its most clear when mousing over furniture, the word "furniture:" is reversed since its defined in a lang file, but the actual furniture name isn't ever interacting with localization at all.

bjubes:
This now works with #1509 so long as any langages with RTL needs put `rtl=true` on the first line and `lang=language_name` on the second

Assets/StreamingAssets/Localization/en_US.lang
```diff
@@ -1,3 +1,4 @@
+rtl=true
```
TomMalbran: Remember to remove this before merging.

bjubes: oops thought i removed that. doing so now

mikejbrown:
Not saying it needs fixing in _this_ PR, but it would be nice if the order of the keys in the localization file didn't matter. In terms of a data model I'm thinking of the localization files as a Dictionary or mapping, which are unordered.

bjubes:
I agree that order shouldn't matter, the only problem is that RTL sets the flag for reversing strings. otherwise if RTL is the last entry the entire table for that language has to be flushed and redone with RTL in mind.

`lang` key  should not matter where in the file it is as well but it would take way longer if we had to search through every single lang file just to find the key to localize the language drop down.

So yes, it shouldn't matter, but with that flexibility we lose the speed of the current implementation. A flexable yet fast solution is not very obvious to me at the moment, so even if i wanted to fix this issue in this PR, I wouldn't know how

mikejbrown:
> lang key should not matter where in the file it is as well but it would take way longer if we had to search through every single lang file just to find the key to localize the language drop down.

I thought all the localization files were loaded on start?

bjubes:
your right. the best solution for letting lang=xxx be anywhere is to build a separate dictionary in the localizationTable class which maps codes (en_US) to language names (english) which can then be displayed. The problem is that the localization table holds a dictionary of each language as a dictionary of strings ` Dictionary<string, Dictionary<string, string>> localizationTable` so iterating through all of it to find every instance of lang as a key is waaay inefficient.

Once this is merged Ill work on allowing both tags (lang and rtl) to exist anywhere in the lang files. that way putting them on top can be convention not an awkward exception to the "order doesn't matter" rule

koosemose:
Looks good, no new bugs, stylecop clean, Merged.

PR #1477, opened 2016-10-02
GasConnection as component (+gas requirements)

dusho:
GasConnection is now component and can be used as gas producer or requirement for other furniture.
Currently used only for Metal Furnace (Metal smelter will probably go away, as furnace does all the things)
image
As of now setup for min. required O2 at 0.1 atm.
image
btw. this should also cover what was missing for 0.1
- Atmosphere requirement for some buildings


Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -368,6 +368,18 @@ public bool OnlyUseDefaultSpriteName
...
+
+    /// <summary>
+    /// Whether the furniture has power or not.
+    /// </summary>
+    /// <returns>True if the furniture has power.</returns>
+    public bool HasPower
+    {
+        get
+        {
+            return PowerConnection == null || World.Current.PowerNetwork.HasPower(PowerConnection);
```
bjubes: ` return PowerConnection == null`
do you intend for objects who have no connection to return that they do have power? would they by definition have no power?

dusho: ye.. I haven't changed the logic there. If furniture doesn't have PowerConnection, means there is no need to power it, so it has always power.

bjubes: in that case HasPower should be false when not on the grid, but then alter the check for operation ` IsOperating = HasPower && canFunction;`
also the current system shows non connected unpowered items as green on line 1075

bjubes: Please change the logic so `IsOperating `does not depend on `HasPower `for items that don't require power. That way items accurately refect their powered state

dusho:
- inc. review comment: renamed HasPower to DoesntNeedOrHasPower, removed unused code
- fixed nil in LUA when mine job was assigned multiple times to same tile (should not be allowed to mine same tile multiple times)

PR #1481, opened 2016-10-02
Added MenuClick sound on most of the UI buttons

Jaytheway:
Tried to add a MenuClick sound to all buttons an menus. With the help of @bjubes I figured it could be done by calling `OnButtonClickedSFX` using `WorldController.Instance.soundController.OnButtonClickedSFX();` inside of every OnButton in a `MenuController` and in a `DialogBoxOptions`.

Also there is a sound for the "Mining Drone Station_OneCreated".


bjubes:
so when there is copy and paste duplication, its usually a sign that some sort of loop or inheritance is needed. In this case its hard to see since there is no base button class we have access 
to. However, we can do have access to the original button prefab that all the other buttons are being created form.

 in DialogBoxOptions::RenderButtons you can see the buttons are being created from a specific prefab. 
`        UnityEngine.Object buttonPrefab = Resources.Load("UI/Components/MenuButton");`

in the lines below that we see that we add the OnButton____ methods individually to each buttons OnClick delegate. 
```newWorldButton.GetComponent<Button>().onClick.AddListener(delegate
        {
            OnButtonNewWorld();
        });
```
So the current flow of the buttons in your PR is `create from prefab > instantiate the prefab for each button > add a OnButton____ method to its onClick> add sfx to each delegate`
a much better flow is  `create from prefab > intercept that prefab >add the sfx call in that buttons delegate >instantiate the prefab for that modified button > add a OnButton____ method to its onClick`

that way you only type `OnButtonClickedSFX` once, and it gets applied to all buttons. Rather than appending a common action like SFX to the end of the "flow" where everything is speiclaized for what each individual buttons does when pressed, this approach targets the beginning of the "flow" before the button is duplicated and specialized.

Okay thats a lot to handle^. give that approach a shot, if you need help understanding or implementing the above just @me on discord or respond below

TomMalbran:
Instead of calling the sound controller from every instance, can we do it so that the DialogBoxOptions and the MenuController have a ButtonClicked event. That event will be invoked manually inside the listener. Then The SoundController can listen to those events and handle the sound. If we do it in that way. Later we could add some sort of button factory so that all the on click events are triggered from a single place, and we have utilities to create buttons.

bjubes:
@TomMalbran  I think were both suggesting a different way of looking at the same implementation. Rather than mulitple manual assignments, soundFX should be assigned once in a common area and exist on all buttons without manually writing it out 20 times.

TomMalbran:
@bjubes Yes. Mi point is mainly that the Menus shouldn't know about sounds, but the sound manager will know which events to hook to to reproduce sounds, just like it is done with furniture, but with everything else.

Jaytheway:
@bjubes I went deep into a google search and Unity manuals, as deep as "The High Level API - Custom Spawn Functions" goes... Also I've tried to use `delegate` to AddListener to a `buttonPrefab`. But, honestly, I couldn't figure out correct syntax to `> intercept that prefab >add the sfx call in that buttons delegate >`

It shows and error `UnityEngine.Object does not contain a definition for GetComponent`.
`onClick` and `AddListener` are red too. 

```cs
private void RenderButtons()
    {
        UnityEngine.Object buttonPrefab = Resources.Load("UI/Components/MenuButton");
		buttonPrefab.GetComponent<Button>().onClick.AddListener(delegate {SoundController.PlayClick();});
	

        GameObject resumeButton = CreateButtonGO(buttonPrefab, "Resume", "menu_resume");
        resumeButton.GetComponent<Button>().onClick.AddListener(delegate
        {
            this.CloseDialog();
			
        });
```

Here is `SoundController`s delegate:
```cs
delegate void ClickDelegate();
	ClickDelegate btnsfx;
	public void PlayClick()
	{
		btnsfx += OnButtonClickedSFX;
		btnsfx ();

	}
```

bjubes:
try casing Object as GameObject in a new variable, then passing that variable as the new "prefab"
```
private void RenderButtons()
    {
        UnityEngine.Object buttonPrefab = Resources.Load("UI/Components/MenuButton");
GameObject buttonGO = (GameObject)buttonPrefab;
        buttonGO.GetComponent<Button>().onClick.AddListener(delegate {SoundController.PlayClick();});


        GameObject resumeButton = CreateButtonGO(buttonGO, "Resume", "menu_resume");
        resumeButton.GetComponent<Button>().onClick.AddListener(delegate
        {
            this.CloseDialog();

        });
```
untested code, but hopefully that solves it or gets you closer to a solution

Jaytheway:
It doesn't work, but definitely getting closer. There are no errors, but the sound doesn't play either.

```cs
private void RenderButtons()
    {
        UnityEngine.Object buttonPrefab = Resources.Load("UI/Components/MenuButton");

		GameObject buttonGO = (GameObject)buttonPrefab;
		buttonGO.GetComponent<Button>().onClick.AddListener(delegate {WorldController.Instance.soundController.PlayClick();});


		GameObject resumeButton = CreateButtonGO(buttonGO, "Resume", "menu_resume");
        resumeButton.GetComponent<Button>().onClick.AddListener(delegate
        {
            this.CloseDialog();
			
        });
```
Could it be the case that `onClick.AddListener` on the `resumeButton` overrides its parent's `onClick.AddListener` on the`buttonGO` "prefab"?

bjubes:
it shouldn't be a case of overriding since AddListener should add not overwrite. Can you push your latest commits up so I can the change as a whole?

bjubes:
Im changing the title to include [WIP] since this is not in a merge ready state. you can remove it once you think your code is ready to be considered for merging

Jaytheway:
Pushed my latest commits

bjubes:
You have merge conflicts now. MenuController.cs has since been removed (which may be part of the problem of why this code isn't working as expected) and the confirm quit dialog has conflicted with the sound you put in that code.

try to fix these and then commit the conflict resolution. Until then ill put [conflicts] in the title. If you need help fixing the merge conflicts just ask.

Jaytheway:
The only problem now is that `MenuClick` plays only when some menu opens or closes. So if the button does not open or close any menu (`ShowDialog`/`CloseDialog`) sound won't play, like "Work" button right now. 
Also "Construction" menu has it own `MenuLeft` class which is not a part of a general `DialogBox` class, so it has its own call for the `MenuClick` sound.

Jaytheway:
This is ready to be merged.

Assets/Scripts/Controllers/InputOutput/SoundController.cs
```diff
@@ -28,6 +28,29 @@ public void Update(float deltaTime)
         soundCooldown -= deltaTime;
     }
 
+	//This is a delegate to play "PlayClick" on ShowDialog and CloseDialog from DialogBox script. 
+	delegate void ClickDelegate();
+		ClickDelegate btnsfx;
+	public void PlayClick()
```
bjubes: the purpose of a delegate is to be used by another class without the original class with the delegate needing to know the other class exists. Because this delegate is never referenced by another file it can be deleted and you can just call OnButtonClickedSFX directly inside of PlayClick.

ideally you would create a delegate (or System.Action, their purposes are identical but Action is a bit easier to use) in the dialogBox class and then add OnButtonClickedSFX to it on line 21 of this file, kind of like the rest of the sounds in this file. However, this approach would be very difficult because DialogBox is never created directly, its only a parent to other specific dialog boxes, so you don't have to worry about it in this PR.

Assets/Scripts/UI/DialogBox/Options/DialogBoxOptions.cs
```diff
@@ -36,7 +36,7 @@ public void OnButtonOpenSettings()
     {
         this.CloseDialog();
         dialogManager.dialogBoxSettings.ShowDialog();
-    }
+	}
```
bjubes: I don't think your editor is configured correctly since your using tabs. you could manually replace tabs with spaces, but its easier to change your monodevelop settings by looking at style guildines in the [contributing file](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/CONTRIBUTING.md#style-guidelines)

Assets/Scripts/UI/InGameUI/MenuLeft.cs
```diff
@@ -33,6 +33,9 @@ public void OpenMenu(string menuName)
 
         menu.SetActive(true);
         CurrentlyOpen = menu;
+
+		WorldController.Instance.soundController.PlayClick();
```
bjubes: I don't think you are using stylecop since your code deviates from the project standards. you can [follow this wiki article to install it here](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/CONTRIBUTING.md#style-guidelines)

Jaytheway:
@bjubes Thank you for pointing that out. 

Delegate removed, style, hopefully, fixed.

bjubes:
not all of the style errors are fixed, you may have to left click the file and hit run stylecop for it to start checking.
also just entirely delete the commented out code in SoundController.

Jaytheway:
Done! But why Tabs are not allowed? It is so inconvenient to press a ton of Spaces instead of a few Tabs

dusho:
image

bjubes:
@Jaytheway when you set mono to use spaces, you hit the tab key like normal but it puts spaces in instead. Tabs/spaces doesnt really matter when working on your own, but github gets a bit wierd with tabs, not to mention that my tabs might appear larger on my machine than yours, causing spacing inconsistencies.

Jaytheway:
Well, I hope now it is ready to be merged:)

Issue #1474, opened 2016-10-01
Repetitve Saving in Settings


bjubes:
While trying to make saving asynchronous I came across another issue. After pressing save, each setting in the dialog is set via a call to `Settings::SetSetting` However, each one on these calls in turn calls `Settings::SaveSettings`. That specific function writes the xml, which means when we press save each setting causes a write call, so settings is written and overwritten 7 times in one frame. This is likely the cause for the lagspike during saving settings

mikejbrown:
I imagine it would probably work fine to make `Settings` a cache by making `Settings::SaveSettings` twiddle a "dirty" flag and spawn a deferred `writeSettingsToDisk` (or whatever) event using the `Scheduler`. `writeSettingsToDisk` would only actually do anything if the cache is dirty (and then mark it clean, obviously). This way a maximum of one write to disk happens per frame.

zwrawr:
I've got a fix for the save settings ( sets a flag which a scheduledevent checks). But that doesn't fix the lagspike. It seems the lag is because all the localisations get reloaded (http://prntscr.com/cp1ww0) , there's no need to load these more than once. I'm going to look into this and then make PR that fixes both of these issues.

zwrawr:
This line seems to be what is causing the lag spike. 
[LocalizationTable.SetLocalization](https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/Assets/Scripts/Localization/LocalizationTable.cs#L73 )
when ever you set the localisation it reloads all of the .lang files. These files should never change once the game has started so we shouldn't ever need to reload them. So it should be safe to remove this call to update the localization table, right?

mikejbrown:
I'm not the localization expert, but that sounds reasonable @zwrawr. You could always just try it :grinning:. The only catch I would imagine is if `UpdateLocalizationTable` fires an event that tells something in the UI to redraw.

bjubes:
Yep assume the Lang files don't change. If we want to allow them to change we can add a seperate reload option akin to reloading a mod which will inevitably take longer

mikejbrown:
I reckon #1490 and #1501 close this. Feel free to reopen if you need.

PR #1490, opened 2016-10-04
optimizes saving settings

zwrawr:
All settings now get saved after the schedualar calls save rather than
when ever a setting changes. also fixes an issue where setting the
localization reloads all .lang files

in response to issue #1474


Assets/Scripts/Models/Settings.cs
```diff
@@ -39,9 +39,18 @@ public static class Settings
     private static string userSettingsFilePath = System.IO.Path.Combine(
         Application.persistentDataPath, "Settings.xml");
 
+    private static bool unsavedSettings = false;
+
     static Settings()
     {
         LoadSettings();
+
+        PrototypeManager.SchedulerEvent.Add(
```
mikejbrown: This changed to `PrototypeManager.ScheduledEvent` in #1469.

Assets/Scripts/Models/Settings.cs
```diff
@@ -124,6 +138,14 @@ public static T GetSetting<T>(string key, T defaultValue)
 
     private static void SaveSettings()
     {
+        // if we do not have any unsaved settings then return
+        if (Settings.unsavedSettings == false)
+        {
+            Debug.ULogChannel("Settings", "No settings have changed, so none to save!");
```
mikejbrown: Could probably get rid of this log.

Assets/Scripts/Models/Settings.cs
```diff
@@ -39,9 +39,18 @@ public static class Settings
...
     static Settings()
     {
         LoadSettings();
+
+        PrototypeManager.SchedulerEvent.Add(
+            new Scheduler.ScheduledEvent(
+                "Settings_SaveSettings",
+                (evt) => Settings.SaveSettings()));
+
+        Scheduler.Scheduler.Current.ScheduleEvent("Settings_SaveSettings", 10f, true);
```
mikejbrown: I had originally thought that changing a setting would spawn a once off event with a delay of like .1 seconds or something -- just to bump the saving to the next frame. But this is probably fine too.

Assets/Scripts/Models/Settings.cs
zwrawr: Fixed!

Assets/Scripts/Models/Settings.cs
zwrawr: Now that its not a reoccurring event this log message makes more sense, as it would be unusual for the message to log.

Assets/Scripts/Models/Settings.cs
```diff
@@ -59,7 +66,13 @@ public static string GetSettingWithOverwrite(string key, string defaultValue)
         }
 
         settingsDict.Add(key, defaultValue);
-        SaveSettings();
+
+        // we have justed altered a setting so we have to set the flag saying their are unsaved settings
+        if (Settings.unsavedSettings == false) {
```
bjubes: Since you copy and paste this block of code on line 71 again on line 109, consider putting it in a seperate method and then calling it in those places.

zwrawr: Done

Assets/Scripts/Models/Settings.cs
```diff
@@ -122,8 +131,29 @@ public static T GetSetting<T>(string key, T defaultValue)
...
+    private static void ScheduleSave()
+    {
+        // we have justed altered a setting so we have to set the flag saying their are unsaved settings
+        if (Settings.unsavedSettings == false)
+        {
+            Scheduler.Scheduler.Current.ScheduleEvent("Settings_SaveSettings", Time.deltaTime, false);
+            Settings.unsavedSettings = true;
+        }
+
+        // else we should already be schedualed to save the settings so dont bother scheduling it again
```
mikejbrown: typo "schedualed"

bjubes:
Wow! settings saves at least 4x faster now

Issue #1459, opened 2016-09-27
Make Saving Asyncronous


bjubes:
Currently saves can take a very long time and in a build will cause the game to temporarily crash, causing windows to report the program has stopped responding.
 We need to give a "Saving..." indication and then write to file outside the main thread, even if the game is stuck at "Saving..." until it finishes.

Same goes for loading

QuiZr:
Well, this won't be as easy as I thought :/
![image](https://cloud.githubusercontent.com/assets/9101486/18902282/2f14a6ec-8552-11e6-9371-e384cd1f1ba2.png)

bjubes:
are you using a coroutine or another implementation?

NogginBops:
Doen't nätter, you cant do `Camera.main` in any other place than the Main thread

bjubes:
im pretty sure a coroutine is on the main thread, it just yields each frame.

sboigelot:
@bjubes is correct, coroutines are only in the main thread

kd7uiy:
This could be dangerous... To do it properly, we'd have to create a memory snapshot of the entire game state, or else risk saving at seemingly random times.

One thing we could do to speed up the save process would be to zip the save files. It should be possible,..

UnknownSoldier1:
Also I would say that XML is a rather strange format for saving savegames. With a custom format we could make the savegame files smaller probably. And I don't know how fast the XML Writer is, but maybe a custom format would save faster too.

kd7uiy:
Compression isn't really that difficult, except for the fact that Unity doesn't support the native C# compression... There are options that I'm looking in to, however, and I think this will help!

DanielNetzer:
why not use json and System.IO?

kd7uiy:
Json uses less memory, but it really doesn't matter which path we go, either is an improvement.

QuiZr:
On my machine serializing world etc takes ~58ms, writing to hdd takes 19ms. Zipping is possible (because we already have SharpZipLib for unfamous LocalizationDownloader) but it would just take longer. I can move the writing to hdd operation on a separate thread and that's the only easy fix that I see here. As @kd7uiy said we would need to create a snapshot of current world&stuff and then launch a separate thread.

kd7uiy:
I like the idea of moving the hdd operation to a separate thread, but still doing the serialization on the main thread.

Zipping could still be a good idea, but we could delay it given that change.

[...]

PR #1496, opened 2016-10-04
Saving World to HDD moved to a separate thread

QuiZr:
Simple and only partial fix for #1459, later whole saving operation should be threaded.


bjubes:
the game still can freeze when saving, but this is definitely an improvement

PR #1503, opened 2016-10-05
[Conflicts]some GUI changes (opinionated)

JulianNymark:
perhaps the only good thing in this commit is the attempt to group how the menu buttons are made (or at least... temporarily made). I'm not sure if the construction menu is _supposed_ to be treated differently from the other dialogs, but i'm guessing not.. or not so in the future... either way, it's now all in sorta one place (not really, just what used to be the OnButtonConstruction function is now an anonymous function similar to all the other dialogs upon construction of the main menu)

**opinionated stuff below this**

i don't like modality too much on gui things in games like these, but... turning it off in the Dialog class might be a bit too much... i noticed it kinda breaks the menu -> settings (and you can now reopen menu behind the settings dialog)

I also thought not auto closing the construction menu on certain categories (essentially anything but 'floor' and 'floor removal') would be more user friendly??, if you want to build multiple things at once, it kinda saves you a step / having to reopen the dialog every time.

### DIFF
--- a/Assets/Scripts/UI/DialogBox/DialogBox.cs
+++ b/Assets/Scripts/UI/DialogBox/DialogBox.cs
@@ -18,9 +18,9 @@ public class DialogBox : MonoBehaviour
 
     public virtual void ShowDialog()
     {
-        openedWhileModal = WorldController.Instance.IsModal ? true : false;
+        //openedWhileModal = WorldController.Instance.IsModal ? true : false;
 
-        WorldController.Instance.IsModal = true;
+        //WorldController.Instance.IsModal = true;
 
         gameObject.transform.SetAsLastSibling();
         gameObject.SetActive(true);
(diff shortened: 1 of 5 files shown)


mikejbrown:
Can you add some screenshots showing your changes?

TomMalbran:
The point of having the construction menu being added from the menu left is because the code is in that file. It was done on purpose. We don't need to have another file know about a Menu Left. If we did wanted to join them it would be a lot better to move them all to a Menu file. Most menu items are added from the same file, because there is where the dialogs are and we haven't changed the dialog manager to work better. In the future we could have an Options Menu Manager that will register itself with the Dialog Manager and will register it's menu item.

The point is that we don't have a class that needs to know about all others just to register the required elements. But a class can know that there are specific managers that have functions to register/add items and there are events that will notify it of any change. So the class is responsible of its own code.

mikejbrown:

koosemose:
This PR has been abandoned with no interaction or responses from the author for three weeks and isn't very clearly explained. Closing.

PR #1504, opened 2016-10-05
Update Localization table in Start

koosemose:
I'm not sure why UpdateLocalizationTable needs called from both awake and start, but I tried removing it from awake, and things didn't work.... so... o.O I don't really know why it's still needed in awake... not even trying to touch the other one in the delegate because there's certainly some stuff going on that I don't fully understand... but at least this works as advertised :D


mikejbrown:
Works as advertised. No obvious new bugs. Merging.

Issue #544, opened 2016-08-21
[Discussion] Health system


FilippoLeon:
I saw talking about needs and enemies. But I searched for health related issues and found no mention of it, so correct me if somebody is working on it/if there is agreement on not working on it right not.

That being said, I'd like to start thinking about an health system. I'd like to work on it, but assemble a team of people to do a cooperative effort.

I have ideas and am willing to work on the health system.

Before I write down my opinions, I'd like to ear from you/get a general idea of interested people and what  is the consensus on how to do the system (and if it is not too early to do so).

gunthergun:
I would gladly help :)

MANDAL0R3:
Currently there is this https://github.com/TeamPorcupine/ProjectPorcupine/pull/135. So there's something begin worked on.

FilippoLeon:
@gunthergun, @BeefEX, would you like to plan something with the temperature system?

longtomjr:
Can we get the status of this please?

koosemose:
We have at least some form of a health system. Closing this now

PR #1508, opened 2016-10-05
Initial Health system

Geoffrotism:
This will be a long term PR so any feedback as I work on it will be appreciated.
I think this will be just the health system for characters for now. Object health stuff can be a separate PR.
That being said I think there are some crossover functions that both could use, such as 
Hurt("something", amount). 
any ideas on a good implementation of this?

- [x] Add LUA commands.
- [x] Add damage (do we want to have damage types for resistances?)


mikejbrown:
Related: #544 which (at the moment) is set for some time _after_ v0.2. Of course, if this nears completion in the v0.2 time frame I'm sure there's no reason not to put it in.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -375,7 +375,7 @@ public bool OnlyUseDefaultSpriteName
     public bool DoesntNeedOrHasPower
     {
         get
-        {
+        {
```
koosemose: Looks like some unnecessary spaces crept in here.

Assets/Scripts/Models/Character/Character.cs
```diff
@@ -577,4 +582,18 @@ private void LoadStats()
...
     }
+
+    /// <summary>
+    /// Use the stats of the character to determine various traits.
+    /// </summary>
+    private void UseStats()
+    {
+        // Base character speed on their dexterity.
+        speed = Convert.ToSingle(stats["Dexterity"].Value) / 2;
+
```
koosemose: Perhaps speed should be calculated on the fly so if something later becomes capable of changing stats, speed will also be affected, or perhaps a better route would be to have a function that sets stats, so when it is used and stats are changed they can be properly updated (which for maxHP we probably wouldn't also want to set currentHP, since then crew would be healed when their con changes)

Geoffrotism: That's part of the reason i wanted feedback. Do we want characters to be able to level up or other such things. CAN their dexterity change? and this part of the code was just initializing on character creation. will add more notes.

Assets/Scripts/Models/Character/Character.cs
```diff
@@ -425,7 +429,7 @@ public string GetDescription()
 
     public IEnumerable<string> GetAdditionalInfo()
     {
-        yield return string.Format("HitPoints: 100/100");
+        yield return string.Format("HitPoints: " + currentHP + "/" + maxHP);
```
dusho: this is a weird use of string.Format.. usually you go with 
`string.Format("HitPoints: {0}/{1} ", currentHP, maxHP);`
to prevent creation of small strings that are GCed later.

Assets/Scripts/Models/Character/Character.cs
koosemose: Actual Leveling doesn't sound quite in line with a basebuilding game, but there might be other things that can affect stats, perhaps training(?) or heavy use to raise a stat, or permanent illness or injury lowering a stat. 

Of course, even barring base stats changing there may be traits that temporarily change a stat (this would probably be a more reasonable way to do illness), in which case speed and other derived stats would still change.

Side Note: this is why I didn't put Health System under 0.2, because it hasn't been heavily discussed and can start getting connected with other systems in a potentially complicated way.

Assets/Scripts/Models/Character/Character.cs
koosemose: Additionally there's also a localization string for this, though I don't know if the info window currently uses a TextLocalizer or how the TextLocalizer will handle larger conglomerate texts such as those used in the info window.

Assets/Scripts/Models/Character/Character.cs
Geoffrotism: I had some thoughts and thought it would be cool if QuillCorp was the one who asked us to go on our "whatever" expedition or something. And if we made them enough money, met some sort of mandate (I'm thinking Ceaser III type thing), or some other trigger: we could send someone to get trained at their HQ for either something like a stat increase, or maybe "Gun Trained" which boosts accuracy. I agree this is for something way down the road, but i wanted to start implementing parts of it in to the 2.0 milestone.

Assets/Scripts/Models/Character/Health.cs
```diff
@@ -0,0 +1,179 @@
...
+    public event Action Destroyed;
+
+    public float Value
+    {
+        get
+        {
+            return healthValue;
+        }
+
+        set
```
dusho: did you just implement whole health system into Value.set property? If this holds with others, can you at least check if the value really changed from previous state? if all checks are for nothing if you end up setting same value.

Assets/Scripts/Models/Character/Health.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            if (value > healthValue)
```
dusho: maybe extract those blocks into separate methods HealthIncreased(value),HealtDecreased(value) , to have that property somehow readable ?

MANDAL0R3: If your talking about calling methods from the property then sure thats easy enough but if your not then are we fine with properties calling methods?

bjubes: @MANDAL0R3 its perfectly fine for a property to call a method to increase readabliity

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -691,6 +697,11 @@ public void ReadXmlPrototype(XmlReader readerParent)
                     reader.Read();
                     Height = reader.ReadContentAsInt();
                     break;
+                case "Health":
+                    reader.Read();
+                    health.InitialHealth(reader.ReadContentAsFloat());
+                    health.MaxHealth = reader.ReadContentAsFloat();
```
MANDAL0R3: Wouldn't this be better as `health = new Health(reader.ReadContentAsFloat());` and then wouldn't it make `InitialHealth(float value)` redundant?

Geoffrotism: I've tried that. It silently breaks. The selection display doesn't show health at all.

ComputerOverlord:
So... you're done now?

Geoffrotism:
As an initial release, yes 🎱

Assets/Scripts/Models/Health/Health.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        set
+        {
+            // New health greater than current health.
+            if (value > currentHealth)
+            {
+                EvaluateHealthIncrease(value);
+            }
+
+            // New health less than current health.
```
bjubes: this is a stylecop error, move the comment to above the if statement

bjubes:
can you add some way to cause damage in game, even if its something as dumb as poking a player causes -1 health, that way it is testable and implemented in some way into the game

Assets/Scripts/Models/Health/Health.cs
Geoffrotism: this is a stylecop error. it IS above the if statement. just look one line below ;)

bjubes: yeah but its inside a if..elseif...else logic. stylecop wants it on top of the first if, so line 67.
just prefix it with // Else:

and yeah i agree that the rule is unnessicary and unclear inside of if else statements.

bjubes:
I want to note that heath doesn't persist on save/load, but as an "initial" health system this seems to work well without error

PR #1509, opened 2016-10-05
[Fix, Small] Options Menu now Localizes Languages

### LINKED ISSUE #358: Localization Localization
Probably rather straight forward but the localisation needs Localising. However do we want the name of all the languages to be in the selected language or do we want them always in the language that it is. I prefer the language that it is.

UnknownSoldier1:
The options menu now localizes the languages you can select. It does this at the moment only for US English, but the people who work on localization can add this to other languages too. This is to finally close #358 and this is also a milestone for Version 0.2.

### DIFF
--- a/Assets/Scripts/Localization/LocalizationTable.cs
+++ b/Assets/Scripts/Localization/LocalizationTable.cs
@@ -37,7 +37,7 @@ namespace ProjectPorcupine.Localization
 
         public static event Action CBLocalizationFilesChanged;
 
-        private enum FallbackMode
+        public enum FallbackMode
         {
             ReturnKey, ReturnDefaultLanguage
         }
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Localization/LocalizationTable.cs
```diff
@@ -132,7 +154,7 @@ private static void LoadLocalizationFile(string path, string localizationCode)
         /// <summary>
         /// Returns the localization for the given key, or the key itself, if no translation exists.
         /// </summary>
-        private static string GetLocalization(string key, FallbackMode fallbackMode, string language, params string[] additionalValues)
+        public static string GetLocalization(string key, FallbackMode fallbackMode, string language, params string[] additionalValues)
```
bjubes: this hsould remain private. a public version exists near line 60 `public static string GetLocalization(string key, params string[] additionalValues)` which encapsulates this version

Assets/StreamingAssets/Localization/en_US.lang
```diff
@@ -1,3 +1,4 @@
+en_US=English (US)
```
bjubes: IMO I think that repeating the language code as the key is unnecessary. inside each lang file there should instead be  `lang=English (US)`. that way the keys are consistent across all lang files. in your approach only english will have `en_US` only hebrew will have `he_IL` causing inconsistencies about which keys should exist in which files

bjubes: IMO I think that repeating the language code as the key is unnecessary. inside each lang file there should instead be `lang=English (US)` that way the keys are consistent across all lang files. in your approach only english will have en_US only hebrew will have he_IL, and so on causing inconsistencies about which keys should exist in which files

UnknownSoldier1: Okay, I might be changing that, but I'm busy at the moment.

UnknownSoldier1:
Changed it! Now it's "lang" for every file.

PR #1510, opened 2016-10-05
Correctly wait for saving to be completed

QuiZr:
Before that user would see "saving complete" pop-up, but in reality a separate thread was still writing data.
Also - I disabled dumping raw XML onto editor's console because it was really hanging the editor.

Sorry for not including this in #1496

### DIFF
--- a/Assets/Scripts/UI/DialogBox/FileSaveLoad/DialogBoxSaveGame.cs
+++ b/Assets/Scripts/UI/DialogBox/FileSaveLoad/DialogBoxSaveGame.cs
@@ -93,9 +93,16 @@ public class DialogBoxSaveGame : DialogBoxLoadSaveGame
             dbm.dialogBoxPromptOrInfo.SetPrompt("message_saving_game");
             dbm.dialogBoxPromptOrInfo.ShowDialog();
 
-            yield return new WaitForSecondsRealtime(.05f);
+            // Skip a frame so that user will see pop-up
+            yield return null;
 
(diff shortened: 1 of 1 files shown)


bjubes:
does this now fully prevent the game from saving on the main thread, thus causing the OS to detect a crash?

QuiZr:
Nope, actual serializing is still done on a main thread.

Assets/Scripts/UI/DialogBox/FileSaveLoad/DialogBoxSaveGame.cs
```diff
@@ -93,9 +93,16 @@ public IEnumerator OkayWasClickedCoroutine()
...
+            yield return null;
 
-            SaveWorld(filePath);
+            Thread t = SaveWorld(filePath);
+
+            // Wait for data to be saved to HDD.
+            while (t.IsAlive)
+            {
+                yield return null;
+            }
```
dusho: how does this behave if there is an exception in SaveWorldToHdd.. it still just finishes and assumes that save was a success ?

QuiZr: If there was an exception while saving there is something really wrong so currently exceptions go to the top of a call tree.

bjubes:
Great change, hopefully we can impelement the same type of thing for loading world as well as loading/saving settings

PR #1516, opened 2016-10-06
[Small] Base Character Animation Speed off MovementSpeed

koosemose:


bjubes:
@koosemose is there a possiblity that movement speed can be so low (or even zero) that the animations wouldn't run? maybe floor between this speed and some base animation speed?

koosemose:
Hmm... that's a situation I hadn't considered.

I wouldn't think we'd end up with a situation where a character would have 0 movement speed and still need to animate... but I suppose that might change if we add more animation types... some of which might have cause to play even for a character that can't move.... Of course even then the animations are going to play ridiculously slowly. I would think to properly account for that sort of thing we would need a more complex animation system in general for characters... I was just trying to minimize skating characters with this (and even for that this still needs some tweaking).

bjubes:
yeah it also bring up the problem of a character having some sort of staticly timed animation, like a blinking light on their visor. again its just a matter of the animation system being very simplistic, but there should at least be something like 
` currentAnimation.Update(deltaTime * Mathf.Max(character.MovementSpeed, 0.1f)); `

PR #1521, opened 2016-10-08
[Fix] Don't phase through floors traveling up, and don't work through floor tiles

koosemose:
Currently, characters will travel up through newly place floor tiles, and will work on a tile from below, even if there is a floor tile in the way.

This regenerates the tilegraph around the tile below a newly placed tile, making the tileGraph aware that access to the tile from below is blocked by the floor tile. A separate check is added when getting acceptable tiles to work adjacent from so that a floor tile in the tile that needs to be worked will block adjacency from below.

Additionally minor changes were made to GetNeighbors (vertical adjacency is optional, and off by default, and the vertical adjacency is farmed to a separate method, to allow access to only vertical neighbors for situations when they need to be handled specially. Also convenience methods were added to access the Up and Down neighbors. The Heuristic Cost Estimate was also adjusted to take vertical distance into account, as this wasn't accounted for in the previous PR adapting the pathfinding to z-layers.

To test:
1st Part: Build a room separate from the starting base on the top level, place inventory in the separate room, in master, a crew will phase through one of the floor tiles, in this PR, they won't start moving.
2nd Part: Find an asteroid on the top level, that doesn't have a layer of asteroid below it (alternatively an asteroid on a lower level that has direct access to the top of it blocked by a layer of asteroid above), and attempt to mine somewhere in the middle, in master crew will happily mine from below, in this PR, they won't. (Testing this portion with the new room won't work as inability to bring inventory to a furniture to be built will mask the lack of ability to work from below through the floor)

Caveat: FPS will drop when the crew is unable to reach the inventory in the newly built room, due to a separate issue in the pathfinder (continually trying to path to it)

### DIFF
--- a/Assets/Scripts/Models/Area/World.cs
+++ b/Assets/Scripts/Models/Area/World.cs
@@ -543,6 +543,7 @@ public class World : IXmlSerializable
         if (tileGraph != null)
         {
             tileGraph.RegenerateGraphAtTile(t);
+            tileGraph.RegenerateGraphAtTile(t.Down());
         }
     }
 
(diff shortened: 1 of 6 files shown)


bjubes:
looks good!

PR #1523, opened 2016-10-09
Remove var usage and General Stylecop cleanup

koosemose:


bjubes:
I fixed many of the same issues in the localization table in my PR #1475 . If you could merge that (assuming there are no problems when you test it) before this goes through it will remove the problems and also implement the RTL support without causing tons of conflicts

dusho:
I thought using var if the same line already contains the type (so the type is nicely readable) was accepted. Blindly removing all vars (which is a nice C# addition) is just non-sense. 
I think having multiple names for Update, FixedFrequencyUpdate, Tick, ..  and inconsistent naming for events is causing more headaches than 'lazy' using of var.

bjubes:
i agree for using var when the type is both repeated in the line AND obnoxious, like:
`System.Collection.Generic.Dictonary<string, ProjectPorcupine.Localization.Keyset> keysets =  new System.Collection.Generic.Dictonary<string, ProjectPorcupine.Localization.Keyset>()`
Everyone agrees that
`var keysets = System.Collection.Generic.Dictonary<string, ProjectPorcupine.Localization.Keyset>()`

is easier to understand. However this case is neither verbose or repeated in line so its nessicary. Besides the guildlines for contributing are quite clear.
> Avoid using 'var', even when the type would be clear from context. Verbose typing is best typing.

dusho:
then either update the guideline to use 'var' in smart way (when lines are too verbose) or make it a stylecop to force people not to use it
not sure why quill put such a weird stuff into guidelines

PR #1461, opened 2016-09-27
Intro scene

Ohda:
**UPDATED 14/10** 
WIP Tag removed has i can't find obvious bugs.
I think i take care of all the comments excepts when @TomMalbran ask to move some Managers to GameController. I love the idea, but i don't find solutions for this without braking the game.
It can be a subject for another PR.
Just added the random background loading (but still 1 background for now)
*******

This is a basic intro scene with a main menu.
- Add a new GameController that will initialize some setup and persist between sceen.
- Add a new SceneController for loading, changing scenes.

Both scenes are able to be launched first. The new scene is in a _SCENES folder.
The Background is from @jaselito Thanks !

There is only one background but if there is more images in the Background folder the game will load a random image.

Known bug :
- There is a bug (also on master) when you open the settings screen without the fpscounter. ( #1526 )

PR #1527, opened 2016-10-11
[Small][Fix] Check if FPS Counter is null before applying settings

### LINKED ISSUE #1526: DialogboxSettings error if FPSCounter is hidden.
There is a bug on the dialogbox settings panel coming from this line : 

`public void OnEnable() { 
      fpsObject = FindObjectOfType<FPSCounter>().gameObject; 
...`
The issue is that fpsObject will be null if the fpsobject is inactive.

I think this is a hack, but when our WorldController initialize the scene, it will always activate the FPSCounter (even if set to false on the settings).

I think the proper solution is to keep a reference to the object before hidding it, so the DialgboxSettings panel will be able to active/desactive it properly but i'am not sure where to do that.

This is another issue i'am facing with the intro scene setup. Any help welcome !

You can add thoses lines on the worldController on start and open the dialogsetting to reproduce on master if your preference say to disable it : 
GameObject` fpsCounter = FindObjectOfType<FPSCounter>().gameObject;
fpsCounter.SetActive(Settings.GetSetting("DialogBoxSettings_fpsToggle", true));`

frankitox16:
### Fixes DialogBoxSettings error
Fixes #1526 

### What this PR does
Now DialogBoxSettings checks if the FPSCounter is null, to avoid errors while loading settings.

### DIFF
--- a/Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
+++ b/Assets/Scripts/UI/DialogBox/Settings/DialogBoxSettings.cs
@@ -48,7 +48,11 @@ public class DialogBoxSettings : DialogBox
     {
         LocalizationTable.SetLocalization(languageDropdown.value);
         
-        fpsObject.SetActive(fpsToggle.isOn);
+        if (fpsObject != null)
+        {
+            fpsObject.SetActive(fpsToggle.isOn);
+        }
+
         WorldController.Instance.spawnInventoryController.SetUIVisibility(developerModeToggle.isOn);
 
         // MasterTextureLimit should get 0 for High quality and higher values for lower qualities.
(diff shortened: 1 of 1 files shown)


Ohda:
It's not enougth to fix the issue.
If the fpscounter is hide on start because userpreference say to disable it, the reference to the object willl not be set and the dialogboxsettings will not be able to show it.

frankitox16:
@Ohda roger that. Will download #1461 and try to see what's going on...

Ohda:
@frankitox16 It may be easier to add thoses lines on start of the worldcontroller from master rather than trying on the intro scene PR.
GameObject fpsCounter = FindObjectOfType<FPSCounter>().gameObject;
fpsCounter.SetActive(Settings.GetSetting("DialogBoxSettings_fpsToggle", true));
I update the original issue post with this info.

PR #1530, opened 2016-10-12
[Fix] Set Settings then Save Settings explicitly

### LINKED ISSUE #1525: Settings DialogBox don't save if game paused
There is a bug when the game is paused, the settings panel doesn't save correctly.
The problem come from here : 
`ScheduleEvent("Settings_SaveSettings", Time.deltatime)`
only happen on
`TimeManager.EveryFrameUnpaused`

This timer has been added in the game in #1490.

### Steps to reproduce the behavior

- set on Pause (spacebar) 
- Open settings panel.
- Make a change
- Save 

### Related issues / pull requests
I've got a problem with this issue on the intro scene, because the game is paused.
Feel free to give me tips or do a pull request to fix it. I'll be really happy to merge it on the intro scene !

koosemose:
As described in #1525 currently due to using the scheduler, settings don't save when paused. 

I have opted to fix this by, rather than saving settings or scheduling a save when a setting is changed, removing SaveSettings from SetSetting, and explicitly saving the settings after all settings have been set.
Fixes #1525

### DIFF
--- a/Assets/Scripts/Models/Settings.cs
+++ b/Assets/Scripts/Models/Settings.cs
@@ -14,7 +14,7 @@ using UnityEngine;
 public static class Settings
 {
     // Settings.xml file that is created if none exists.
-    private const string FallbackSettingsXML = @"
+    private const string FallbackSettingsXml = @"
 <Settings>
   <worldWidth>101</worldWidth>
   <worldHeight>101</worldHeight>
@@ -39,16 +39,9 @@ public static class Settings
     private static string userSettingsFilePath = System.IO.Path.Combine(
         Application.persistentDataPath, "Settings.xml");
 
-    private static bool unsavedSettings = false;
-
     static Settings()
     {
         LoadSettings();
-
-        PrototypeManager.ScheduledEvent.Add(
-            new Scheduler.ScheduledEvent(
-                "Settings_SaveSettings",
-                (evt) => Settings.SaveSettings()));
     }
 
     public static string GetSettingWithOverwrite(string key, string defaultValue)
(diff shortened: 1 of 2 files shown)


mikejbrown:
The only potential issue with this is if someone forgets to call SaveSettings.

I assume you checked that nothing else is calling SetSetting that needs to save? (Haven't had the chance to look myself yet.)

Ohda:
Mikejbrown is right, there is some calls from WorldController and LocalizationTable too.
Except that, i feel better with your solution !

koosemose:
In the case of SetLocalization in LocalizationTable, that method is only called from DialogBoxSettings (and probably shouldn't be setting the setting itself anyways.

The WorldController one is... strange, changing a setting with a hotkey feels strange to me in general (and has the feel of an easy development shortcut than anything), Personally I feel fine with having dev mode when toggled with hot key not being saved, since the user isn't explicitly saving it (by hitting the save button for settings) then why should the code explicitly save it.

koosemose:
Actually, the more I think about it, the more I think that we really shouldn't be setting any settings outside of the actual settings menu, and therefore that should be the only place that we should need to save settings from.

bjubes:
I thought the scheduler was used so that settings was only saved once rather than 6+ times when each property was changed, but I could be wrong. does the "one save at the end" implementation still exist in this change?

koosemose:
Yes, it's just done directly, changing a setting no longer saves the settings, they instead have to be explicitly saved afterwards. So the flow is Change all settings, and then save.

So yes, it still exists, but through a method that avoids using the scheduler, because the scheduler doesn't work very well for out of game stuff as it only runs when unpaused.

bjubes:
ok just double checking. Merging

PR #1531, opened 2016-10-12
Implement Vertical rooms

### LINKED ISSUE #1491: Rooms spanning multiple levels, and enclosing rooms vertically.
Currently rooms only care about their floor on room creation, if a floor tile is removed, the room stays intact, and if a "room" is built with a hole in the floor it will be outside, regardless of what lies beneath.

Optimally a hole in the floor should either expose the room to space, or join it to the lower room.

What about going up? Having some type of separate "roof" tile seems needless tedium, and having to build a floor on the next level up feels awkward (In addition to problems on the top level, either enclosed rooms can't be built on top level or the top level is special and doesn't require roofs built with floor, which seems inconsistent).

Perhaps, at least at initially, we just have "magical" roofs if the room above is room 0, we treat it as enclosed, as though it had a roof, but if the room above is an actual room, we merge them as one would expect.

koosemose:
Currently rooms are single level only, and don't change when a tile within them is changed to empty.

This PR makes floodfill for room expand into 3 dimensions, and calls floodfilling when a tile is changed.

Intended Behavior: Rooms on top level have a "magic" roof, and do not require anything above them to enclose them, rooms on lower levels require some kind of floor (i.e. not TileType.Empty) on the level above to act as roof. A room created across 2 z-Levels should have the same room index (and for all intents and purposes be a single room) if the upper floor doesn't completely separate it, and separate into two different rooms if the upper floor has its floor completely filled in.

Closes #1491


frankitox16:
Currently the room's floodfill is preety slow, and causes some FPS drops... Even if this is a nice enhancement, wouldn't this be making it worse? Or did you manage to improve this aswell?

I do think this is necessary as a game feature, but I don't want many performance issues...

koosemose:
I've never had any issues with floodfilling the rooms, even running recording software at the same time I can't notice any hint of an effect on floodfilling, since it bails as soon as it hits a tile that is guaranteed to be outside.

If you notice slowdowns somewhere feel free to do some profiling to see if you can nail down what makes it slow for you, perhaps we can find some areas for optimization.

frankitox16:
I'm testing it in a preety low budget pc (still runs smoothly when some jobs are being done) but whenever I make a room (or remove it) it goes down to 2 fps, that's the reason why I mentioned fps drops. But if it runs smoothly for you, then I guess I should consider improving my hardware...

koosemose:
On the plus side, it should only a noticeable effect if you are creating rooms spanning floors. The one potential issue for you is that it now also floodfills rooms on tile placement, but that is an absolutely required feature, since otherwise a room can be opened to space through its floor and still be considered a room. But in most cases it should relatively quickly bail out of floodfilling for everything but the room being enclosed by floor, and during wall placement it will act virtually the same (since without spanning floors it won't have any other levels to flow into)

bjubes:
Im getting Null Refrence Exceptions when loading from a save.

koosemose:
Urgh, more annoying weirdness from TileType. Since it doesn't function like Furniture or Utilities and is just a change of the tile's Type, there's not anywhere to easily have it not do a roomfloodfill on a load... 

I'll see if I can find a workaround, I think if I check for a null room when doing a floodfill from tile placement... that might do it, as long as we continue to require floor to be under any furniture, or at least enclosing furniture.

koosemose:
Well, that doesn't work, just leads to a bunch of null rooms... which makes sense, since it's still trying to start floodfilling for the first tile, which ends up null it's room. Tried changing the room setting on load until after the tileType is changed, and that leads to everything being outside... which I'm not absolutely certain why it happens like that.

I'm kind of stumped at the moment, and am not really sure how to go about this with TileType working the way it does.

koosemose:
Ok I think this sorts out the problem on loading a save... it feels a bit hackish to me but it works, and I think it's the best I can do with TileType as is.

bjubes:
After making a multi floor room and drilling out the bottom floor's floor, the entire room becomes -1 despite being unenclosed and retains gas information.

tests

koosemose:
Ok, so the results I got are that in drilling a hole through both rooms, the bottom part of the room actually properly becomes room 0, i.e. the outside room, and the upper part of the room becomes room -1, meaning it is a separate room that isn't in RoomManagers list of rooms, which further means that it doesn't get its room changed ( so it would still be the room formerly known as room 7 ), but the room is getting removed from RoomManager's list.

I'll try some more investigation to see why/how exactly this is happening.

koosemose:
It feels a bit hackish, but I as I can't see where exactly it's failing to do what it should, I am not manually assigning the old rooms to the outside room if the new room formed is outside. It seems to work fine, and doesn't mess up it properly joining the rooms if there's another room below rather than open space.

bjubes:
I'm still getting the -1 room problem using the same steps...

koosemose:
So apparently what I did fixed the problem if the holes aren't aligned, but not if they are aligned... o.O

koosemose:
So now I'm going with what is at least a more direct method, if the tile below the tile being removed is outside, I just assign the tiles of the old room to the outside, it seems to cover every situation, without breaking anything else.

bjubes:
looks good. no more -1 room and save/loading rooms works again

PR #1532, opened 2016-10-12
Limit number of autosaves

### LINKED ISSUE #1529: Limiting amount of Autosaves and ordering of save files
Currently our autosave system will just keep making new files (I'm currently up to 225 just from testing various things) and orders them by creation time. 

The potentially large number of autosave files (I imagine it will only grow larger during actual play), plus ordering by creation time makes it impossible to have regularly used save files, as they will quickly get pushed down by autosaves, in addition to the (in my opinion) general undesirability of having 200+ autosave files. 

To remedy this, I propose solving this from two directions, first is limiting the number of autosave files, to a number set in settings (with a proposed default of 5), and when you have this many it will replace the oldest when saving. The second direction would be simply sorting save files (in load and save, and anywhere else they may appear in game) by LastWriteTime, so that your more recently saved to files will be at the top.

On a related note, I would like to know what people think of autosaves being suffixed with the time and date of the autosave, rather than a number (this would also imply explicitly deleting the oldest save file after a successful save, since the actual filenames won't be reused).

I've already started on an implementation of this, as I'm assuming this won't be controversial, but would like to get input as to details (and ensure my assumption is correct). This implementation doesn't include changing the suffix of autosaves, as that may be a less generally agreed upon thing.

koosemose:
Currently autosaves will always create new files, quickly getting to the point where there are hundreds of autosave files, and this is just over the course of testing during development and PRs.

This PR implements a max number of autosaves (default of 5), after which point it will begin saving over the oldest autosave file.

Minor Fixes: SaveWorld was moved to WorldController, so that saving manually and autosaves won't have separate implementations of SaveWorld, save file sorting is based on last file write time, along with displaying the last file write time rather than creation time. A setting for autosaveInterval in the settings dialog was added along with the one for max autosaves.

Note: If you have more autosaves than the current max number of autosaves it will cycle through all of them, rather than just limiting itself to only the max number set.

Closes #1529


Assets/Scripts/Models/Scheduler/ScheduledEvent.cs
```diff
@@ -260,6 +260,21 @@ public void Stop()
             RepeatsForever = false;
         }
 
+        public void SetCooldown(float newCooldown)
+        {
+            if (newCooldown < TimeToWait)
+            {
+                // Decrement TimeToWait by the amount of time that has passed since the last time this event fired
+                TimeToWait -= Cooldown - TimeToWait;
```
mikejbrown: Why not `TimeToWait = newCooldown - (Cooldown - TimetoWait)`?? I'm getting confused by this.
Also, please write unit tests speccing out the desired behaviour for this function. :grinning:

Assets/Scripts/Models/Scheduler/ScheduledEvent.cs
```diff
@@ -260,6 +260,21 @@ public void Stop()
...
+                if (TimeToWait < 0)
+                {
+                    TimeToWait = 0;
```
mikejbrown: Should this fire the event? Maybe instead of reset TimeToWait replace this with a call to update?

koosemose:
Tests written, unconflicted

Assets/Editor/UnitTests/Models/ScheduledEventTest.cs
```diff
@@ -17,6 +17,7 @@ public class ScheduledEventTest
 {
     private Action<ScheduledEvent> callback;
     private bool didRun = false;
+    private int runCount = 0;
```
mikejbrown: Since you're adding this it would be easy enough to get rid of didRun completely. That could be a separate PR after this is merged if you wish.

Assets/Scripts/Models/Scheduler/ScheduledEvent.cs
```diff
@@ -262,14 +262,11 @@ public void Stop()
 
         public void SetCooldown(float newCooldown)
         {
-            if (newCooldown < TimeToWait)
+            // Decrement TimeToWait by the amount of time that has passed since the last time this event fired
+            TimeToWait = newCooldown - (Cooldown - TimeToWait);
```
mikejbrown: :+1:

Assets/Scripts/Models/Scheduler/ScheduledEvent.cs
mikejbrown: Yes, the question is if this is the desired behaviour. I suppose it makes sense to only fire once, as long as the behaviour is specified somewhere.

Assets/Editor/UnitTests/Models/ScheduledEventTest.cs
koosemose: While you can get that the callback ran from checking the runCount, in most of the cases it seems (to me) that what is being looked for is simply that it did run, and it seems like the separate boolean makes it clearer what is being looked at. 

Of course, standard disclaimer that I'm far from an expert on tests of this sort, so I'm not sure exactly what is the preferred way to go about them.

mikejbrown: Well I'm no expert either. :D It kind of makes sense to test also that it's being run the correct number of times. Using a bool could potentially miss bugs where the event is getting run too many times.

koosemose: Good point... Though I'd also prefer it be in another PR... done by someone else... mostly because I'm not absolutely clear on the functioning of the scheduler at this time... I only figured out enough to get what I needed for autosaving to work the way I thought it should.

bjubes:
make sure you stylecop, there are a few violations. also this has some conflicts

koosemose:
Seems I let slip one in the test... It however will take a bit to merge things, git's ideas of where things go is... interesting.

koosemose:
That should take care of the conflicts and there shouldn't be any more stylecop issues

koosemose:
As this has been approved by another GK, and no recent concerns have been brought up, I am self-merging.

PR #1534, opened 2016-10-14
Adjust Charaters speed.

Ohda:
I adjust characters speed with this formula.

speed = baseSpeed +/-30% of baseSpeed depending on Dexterity.

The baseSpeed is 5 tile per sec. So it can go from 3.5 to 6.5.

The original formula is Dexterity / 2 so there is a range of 0.5 to 10. It feel a little too extreme.

### DIFF
--- a/Assets/Scripts/Models/Character/Character.cs
+++ b/Assets/Scripts/Models/Character/Character.cs
@@ -611,8 +611,8 @@ public class Character : IXmlSerializable, ISelectable, IContextActionProvider
     /// </summary>
     private void UseStats()
     {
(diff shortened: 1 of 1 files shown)


dusho:
tested this... it's much better than what is currently in master, though +/-50% still feels too much of a difference.. maybe 20% is good enough?
currently one guy goes for some inventory, but other one sprints though him and grabs it first.. looks un-productive..
speed should be mostly adjusted by tiles characters are walking on?

Ohda:
Updated from 50 to 30% to reduce the effect.
If the caracters runs for the same furnitures, it's an issue with jobs system i guess.

PR #1520, opened 2016-10-07
PowerConnection, Animator components, components integration

dusho:
Introduces new components
- power connection (for all things that require or provide power)
- animator (for all things that animate)

Changes:
- added IPlugable interface , used in PowerNetwork instead of component, PowerConnection implements IPlugable

Generally introduced inter-component communication using furniture's params. 
Things to be still done:
- (DONE) fix the older PowerConnection UTs
- (DONE) way to declare/expose params in furniture.xml
- (DONE) furniture needs map with 'needs' for it to work - same as power icon, -> gas, temperature, input inventory, context selection, ... 
- (DONE) fix stylecop
- nice to have support for LUA functions instead of conditions in .xml (another separate PR)

Create WIP to gather some feedback if the implementation is going in right direction and whether there are better ideas how to handle or enhance things.
Most of things I tested seems to be working - I would ask other people to test it as well..

PR #1537, opened 2016-10-14
[WIP, Feedback] Power through Utilities

koosemose:
Currently Power is universal, this changes power to travel along Utilities. At the moment, all utilities carry power, of course this isn't what we want in the end, however at this stage (without a fluid system or anything else to use for other utility types) are to just have every Utility carry power (as I've chosen to do for the time being), or in multiple places have special cases where the utility type is checked and only has the network built if it's a power cable. 

This is still in somewhat of a state of development, and therefore still has assorted debugging tools in place (such as showing the id of the grid on mouseover of a utility), but it is at a state now that I think I can get useful feedback on where and how things might be made a little more streamlined. 

A furniture should connect to a utility if it is placed on top of that utility, or an event is triggered if a utility is placed in the same tile as a furniture to allow it to plug in. The utility network should be built for a given utility (again, at this time only power, and all utilities act as power cables) such that all utilities that are visually connected share the same grid/network, if two visually separate utility networks are connected (by adding a utility of the same type connecting them) they should join and share power as would be expected. If a network is split, the two visual halves should end up with separate grids, and furniture connected to one or the other should only be able to draw from power available on that network.

There is currently an issue where when grids are split or joined connected furniture "flickers", I'm not certain if this is purely a visual thing or if they are starting/stopping working for a moment. I'm also not certain at this time how/if I can avoid this, if anyone has any thoughts, please let me know.


dusho:
In my PR ( #1520 ), I introduced PowerConnection as component, so it's not so tightly linked with Furniture anymore.
As we discussed on Discord, I introduced interface IPlugable (which PowerConnection implements), so PowerNetwork (NetworkManager) will hold just those. I assume the IPlugable should be extended to have methods PlugIn(Network), Unplug(Network), where UtilityManager will control what furniture is plugged where.

koosemose:
@dusho It will be easier to implement your changes after one of ours is merged, rather than trying to copy things over. And from what I've been able to work, I don't believe, at least at this time, having to have UtilityManager involved is needed. Once the whole system is integrated together, I may take a second look at things to see how it all fits together. For the most part IPlugable just needs to have any of the methods used from it by Grid and the old Connection.

koosemose:
Some things have gone horribly wrong in trying to sort out some issues... closing this, and will open a new PR once things are sorted out.

PR #1538, opened 2016-10-15
Load localization using seperate config file

### LINKED ISSUE #1524: RTL and lang attributes in the localization system
Currently in order for a language to support RTL (hebrew and arabic), they must have `rtl=true` as their first line in the lang file. The `lang=` must come next if a language has a "localized localization", meaning that English (US) comes up in the language dropdown instead of en_US.

In #1475 there was discussion of redesigning the system so that the rtl and lang attributes could appear anywhere in a file. I think this is a much better design considering order doesn't matter in the rest of the file, so it makes sense for the entire file to act as an unordered dictionary.

However, this new approach has a flaw. the localization system will have to read the entirity of every file on start, as opposed to just english and the one selected in settings. Currently we have over 50 languages, so the increase of processing time will be immense.

so here are the possible solutions off the top of my head. Please comment on which one you think is best.
- Keep the system as is, just make it arbirary that the lang and rtl attriubtes must appear first. I will fix up the code in localization table to do this more cleanly, right now its kinda a mess.
- Scan every file on start, but on a seperate thread. This will be similar to how world is loaded, using a Thread object and couroutine which yeilds to it. I don't know how long scanning through 50 files which contain potentially hundreds of strings, but I can make it prioritize english (default) and the langauge chosen in settings first, and the rest will load in the background as the game runs.
- Keep rtl as the first attribute if required and move "localization localization" to a separate file (say lang_codes.lang?) where it maps every code to its name e.g. `en_US=English (US)`

bjubes:
### The issue this fixes
- prevents awkward `rtl=true` and `lang=English` lines in lang files that had to appear at the top and in a certian order, despite the rest of the file being unordered.
- prevents reading each lang file once partially to extract lang and rtl, and then again to get actual language information
- uses config.xml as new home for version of localization instead of curr.ver file.
- Fixes #1524 
- Bonus: some unit tests for the localization reverse feature
### What this PR does

 Adds a config.xml file to localization which defines a language code, its localized name, and whether it is rtl like so:

``` xml
...
<languages>
    <language code="af_ZA"/> 
    <language code="ar" rtl="true"/> 
    <language code="en_US" name="English (US)"/>
    <language code="es_ES"/> 
    <language code="fr_FR"/> 
    <language code="he_IL" name="עברית" rtl="true"/> 
...
```

note that all attributes are optional, except `code`. name defaults to whatever `code` is, and `rtl` defaults to false.

this works by adding a new class `LanguageData` (suggest a better name) which stores the three attributes above. an instance of that class is created for every language and then stored in a dictionary called `localizationConfigurations`. This dict is filled before any lang files are read, so by the time the system comes across Hebrew, it knows it should be RTL, and when the system makes the drop down for languages, it already knows the name of every language without iterating through each lang file.

in addition to the above, config.xml also holds a version element that acts exactly like the curr.ver file does in master.


Assets/Scripts/Localization/LocalizationDownloader.cs
```diff
@@ -255,14 +263,8 @@ private static void OnDownloadLocalizationComplete(Action onLocalizationDownload
                 Directory.Delete(dirInfo[0].FullName);
 
                 Debug.ULogChannel("LocalizationDownloader", "New localization files successfully downloaded!");
-            }
```
bjubes: Note: Removed this because catching the exception only to spit it back out makes it much harder to find the specific line that raised it. letting the exception remain is best until it is handled properly

koosemose:
Is it failing to find config.xml the first time and and erroring on a bunch of languages not being defined the expected behavior? If so perhaps the error message when config.xml isn't found should inform on what is happening/should happen.

bjubes:
technically after this PR config.xml would always exist unless the user manually removed it. But in the rare case of a failure to update the localization properly or of user tampering, Ill add better error catching. 

This problem arises from inconsistencies in what the config says exists what actually exists. They should never lose sync unless there is an IO error or intentional deletion.

bjubes:
after looking at the errors your describing, I think its proper that each language reports that it isn't in the localization file. I changed it so that this only occurs if the config.xml is present, so when its missing only one error occurs.

Removing [wip], but the notes in italics in the OP are still true.

Assets/Scripts/Localization/LocalizationTable.cs
```diff
@@ -120,6 +174,82 @@ public static string[] GetLanguages()
             return localizationTable.Keys.ToArray();
         }
 
+        public static void LoadConfigFile(string pathToConfigFile)
+        {
+            localizationConfigurations = new Dictionary<string, LanguageData>();
+
+            if (File.Exists(pathToConfigFile) == false)
+            {
+                Debug.ULogErrorChannel("LocalizationTable", "No config file found at: " + pathToConfigFile);
```
koosemose: Should this also be a Warning, since it is an expected result at certain times, and is remedied by creating one if it isn't found? Assuming I'm understanding the process of what's going on anyways.

bjubes: Thats assuming everything goes smoothly. If the config doesn't exist, then the localization code will attempt to redownload the latest version which will contain a clean config. However, if the user is offline then the download will obviously fail. In that case the lack of a config will cause a lot of weird behavior, including rtl languages not being rtl, not showing languages in the dropdown (it will fallback to codes), and other minor glitches in the localization system. I don't know if that is considered an error or warning, but regardless the results will be unexpected without a config.xml file.

bjubes: actually the game is a bit too reliant on config.xml, I didn't realize that not having it causes tons of null key errors. Ill relook how the system works when the file is missing. It should probably be an error, but the game should at least still show english or at least localization strings when its missing.

koosemose: Maybe a note of something to the effect about needing to stop and restart in the error message? I don't know exactly, I'm just expecting ending up with lots of bug reports during development any time someone goes to do something while not having config.xml that would be fixed by just stopping and hitting play again after it's downloaded a new one.

bjubes: Well what should happen is that once it is downloaded, everything should work as normal and the localization should reload itself, but that doesnt seem happening.

bjubes:
Well i figured out the issue. When i merged earlier I accidentally nuked some of my changes. Now when config.xml isn't present it rightly reports so as a single error. The system automatically tries to update, and if there is an internet conection will correctly load the latest localization and all is well.
If there isn't any internet then there maybe problems such as lack of rtl support and such because the config is missing, but there are no other errors or problems that break gameplay

koosemose:
You've got some stylecop errors, I would just merge and stylecop them myself but I'm not entirely certain what should be done about the complaints in LocalizationTable.cs

bjubes:
ahh sorry, I reverted back my work and then never restylecoped or separated classes. That should fix it. 

Since this is merge ready im going to revert back to using QuiZr's localization repo. However, **don't merge this** in until [this pr](https://github.com/QuiZr/ProjectPorcupineLocalization/pull/29) goes through or else it will complain endlessly about no configuration file being present.

bjubes:
ignore the "wont work until 29" commit message, as that has [already been merged](https://github.com/QuiZr/ProjectPorcupineLocalization/pull/29).

koosemose:
Alright, looks good, Merged.

PR #1540, opened 2016-10-16
[Fix] Fixes crash when loading save with character carrying inventory

dusho:
Fixes bug where loading character inventory gave Dictionary - keys doesn't exist exception.
To reproduce: save game while character is holding some inventory, load it.
As jobs are not persisted yet, character will hold the inventory, but he 'forgets' what he was doing...

### DIFF
--- a/Assets/Scripts/Models/Inventory/InventoryManager.cs
+++ b/Assets/Scripts/Models/Inventory/InventoryManager.cs
@@ -161,6 +161,10 @@ public class InventoryManager
         {
             character.inventory = sourceInventory.Clone();
             character.inventory.StackSize = 0;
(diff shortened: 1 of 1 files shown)


koosemose:
Strange that this bug turned up again... I fixed what seems to be this exact same bug a couple months ago... I guess some changes to our system caused my check to no longer work right.

koosemose:
Seems to fix the issue, and behavior matches what it was the last time this was fixed... rather or not that was the exact right way to handle it... it's the same :). 

You do have a small Stylecop error however. Once that is fixed this should be good for merging.

koosemose:
Oh good, you had maintainer edit on :). 

So, looks good, Stylecop Clean ;) and Merged.

PR #1542, opened 2016-10-16
Fix for no audio when game is build

### LINKED ISSUE #1533: Sounds don't play when game is build
I build a standard PC game and i can't here no sounds when running the game.

Ohda:
Fixes #1533
Having LoadAudio and GetAudio static in AudioController seems to be a problem when the game is build.

**Edit** : Looks like it's only an uncomplete fix as The Player Settings>Other Settings > Logging>Log level need to be set to Full for the sounds to played. Developpement build is not needed.

### DIFF
--- a/Assets/Scripts/Controllers/InputOutput/SoundController.cs
+++ b/Assets/Scripts/Controllers/InputOutput/SoundController.cs
@@ -36,7 +36,7 @@ public class SoundController
              return;
         }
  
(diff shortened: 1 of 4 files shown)


Ohda:
Hum it was a mistake. Nothing better with this. 
Still the issue if Logs are not enabled in Player Settings.

PR #1544, opened 2016-10-16
Intro scene

Ohda:
This is a new PR for intro scene.
It's the same branch than the original PR with a fix for the issue spoted by @bjubes when he try to merge it.
The game wasn't loading correctly as i forgot to set a value.
Original PR #1461


bjubes:
Loading now works in  _SCENE_. Thanks for the fix!

PR #1546, opened 2016-10-17
Provide gas at proper ratio

koosemose:
PR #1477 changed some of the intended behavior of the Oxygen Generator due to some of the details of how it was implemented. This PR attempts to restore the intended functionality with minimal alteration of GasConnection prior to a hopefully cohesive upgrade to the functionality of the GasConnection component that allows the Oxygen Generator to behave as intended without breaking other aspects of the component and without somewhat hackish workarounds. There was also a small error that would very occasionally crop up due to operating with a rounded version of the rooms gas pressure rather than the actual gas pressure.

The intended functionality was that the Oxygen Generator would have a set rate of gas production, and each gas produced would be produced at a portion of that rate relative to its target portion of gas. As of #1477 it was changed to each gas having a rate set individually. This PR sets the rates for each individually as it would be under the old system (original rate was 0.16, so o2 is 20% of that, or 0.032, and n2 is 80% of that, or 0.128). Future upgrades will hopefully allow the intended but unimplemented functionality (that was not easily achievable with LUA) of having the gas rate consistent, such that if the room was at the limit for N2 (and therefore the generator didn't need to produce any) it would produce O2 at a rate of 0.16.

The issue regarding the rounded gas pressure would cause it to stop producing gas before a room was entirely full as it was rounding the pressure, which would cause irregularities depending on game speed (particularly as game speed was changed) where the pressure would get close enough to the target pressure that it rounded to that amount. This PR removed the rounding so that it is working with the actual gas pressure, and leave the prettified rounding solely to the UI display.


dusho:
Ye, looks right now. Sorry about issues.. Wanted to sync animation with numbers shown in room and evidently did it in stupid way :) (rounding, that affected also the functionality)

koosemose:
Yeah no problem... I wanted to do the gas ratios differently (the whole have a set rate for the machine as a whole (or the gas component anyways) and gasses produced at a ratio of that, that relates to their target pressure, or they just have a ratio and the component has a total target pressure), but I realized half way through I was hacking it up horribly. I'm rather hoping you can revisit it at some point and do what I'm trying to achieve in a proper way.

dusho:
one benefit of having rates separated like this is that you have freedom to give rates to every gas separately and can e.g. create plant that consumes N2 at some rate, and produces O2 at different rates..
there are still thresholds that will ensure correct ratios

koosemose:
On the other hand you can't have something like the O2 generator have a maximum flow rate and use that rate for whatever gases it needs. I realize I'm not explaining that very well, so I'll provide an example. So the O2 gen has a maximum flow rate of .16, let's assume some event suddenly consumes a relatively large amount of oxygen (a massive fire for example), currently (with this PR) even though it doesn't have to use any of it's theoretical maximum flow rate for nitrogen, it is still limited to .032 flow rate of oxygen, optimally, if it doesn't have to use any of its flow rate for nitrogen, it should be able to use all of that rate for oxygen and therefore be outputting oxygen at 0.16.

Would it be feasible to have both as options? i.e. You either list each gas separately as it is now, in which case it produces as it currently does, or list them in a containing element that has a flow rate of its own, and shares that flow rate between them as I described ( possibly with an actual ratio to use when need for both is equal, rather than the old way of just putting them out at a ratio based off of the desired pressure of that gas )

I've got an assortment of other ideas I'd like to be able to do with gasses to properly implement other behaviors but they're further disconnected with the topic of this PR.

dusho:
I see, so something like this would maybe cover all we need:
```xml
<Component type="GasConnection">
    <Provides>
        <GasInfo gasMixture="Air" rate="0.16" minLimit="0" maxLimit="1">
            <GasRatioInfo gas="O2" ratio="0.2"/>
            <GasRatioInfo gas="N2" ratio="0.8"/>
        </GasInfo>        
    </Provides>   
</Component>
```

koosemose:
I think so. Assuming xml as it is now would continue to produce as it does now?

dusho:
not sure.. this new way is more generic and should cover also old way.. if you will just specify new GasInfo.. total limit will be relative_ratio * maxLimit

koosemose:
That sounds like it would be pretty good too, especially if ratio defaults to 1 (or maybe 1 divided by number of gasses just in case) that way for a gasinfo with just one gas, you really just need the name.

koosemose:
As this is a fairly small change and no concerns have been raised, I am self-merging.

PR #1547, opened 2016-10-17
[Small, Fix] Remove Warning from Parameter Autocreation

koosemose:
Parameter Autocreation is an intended feature, and shouldn't generate a warning.


koosemose:
As this is such a tiny change, removing a certainly unneeded logging line, and is hindering development of some others, I am self merging this one.

BraedonWooding:
Yeh I agree, but maybe log something at-least?

PR #1553, opened 2016-10-18
FPS Counter Additions and Rename to PerformanceHUDManager (#1550)

### LINKED ISSUE #1550: [pull request made] [enhancement] FPS Counter Additions
*Currently I'm implementing this.  PR hasn't been made* **yet**
### Expected behavior

- An average FPS (for last 30 seconds)
- A min FPS (the lowest FPS reached)
- A max FPS (the highest FPS reached)
- Current ms (will be 0 till we add networking)
- Total Memory, Allocated Memory and Memory for Mono

Something that looks like:
image

### Actual behavior

Displays the current Frames per Second (FPS)

BraedonWooding:
Closes #1550 
** NOTE: Fully implemented just needs some small-ish cleanup to document **

### What this PR does
This not only adds new performance tracking options (min/max fps, avg fps, current fps, current ms, total mem/allocated mem/mono mem, user computers stats/build, resources used...) but also adds support to easily add any future one (via a subclass), this will allow modders in the future to potentially create their own performance tracking options (if they so desire?).  And it allows us to very easily create ones too.

I've followed a highly modular based approach with it only have a single access point (via a flag to 'dirty' the UI to cause it to clear and re-render on next update), to keep it apart from the actual game.

(NOTE: Image is cut off a little at the top cause bad snipping).
image

image

### TODO

- [x] Implement main structure for Issue (A class based structure)
- [x] Implement the displaying/updating/cleaning of the GUI
- [x] Make it look like it did before but using the new system
- [x] Change dirty flag to a function call (static).
- [x] Implement the other components (just requires a subclass)
- [x] Upload Image for the other components
- [x] Implement a dropdown for FPS Options (with the various options)
- [x] Upload Image for dropdown
- [x] Remove old implementation (standard assets FPS Counter)
- [x] Color changes of current FPS depending on value (< 30 red, >= 30 < 55 yellow, and >= 55 green)
- [x] Cleanup

### How it works
Essentially it all is subclassing, with a relatively short main class (860~ lines total from 8 files; though not fully done yet).
- The main class which is called PerformanceHUDManager, and handles the drawing/cleaning/updating and general 'day to day' things that the HUD requires.  Transform hierarchy wise it is the child of the old FPS object (MenuTop), and it is simple; 
    - it uses another class called PerformanceComponentGroups (which is just a static class to hold PerformanceComponentGroup; which are just read-only public statics that sole purpose is to act kind of like a struct/enum, with more functionality), 
    - it has a current group (and groups are just 'groups of elements' with elements being different components such as current FPS, average FPS, ms, memory and so on...), 
        - there are 5 groups: none (empty kind of like a null case or when you don't want any HUD), basic (just FPS), standard (FPS + ms), extended (FPS, avg FPS, max/min FPS, ms), verbose (all options - that come with base game).  
    - it also has a dirty flag, which is a public static bool
        - When set to true, the next time the app does an Update() call it will clear the UI, re-draw/render then update the components.  This will be used when changing current group/mode.
    - It shows Components in a grid (around 6 total; with 3 on each 'line' or row)
- The star of the show however is BasePerformanceComponents (the class is just called BasePerformanceComponent though), since they act as your various elements (such as FPS, avg FPS, max/min FPS, ms...), they have a few key functions:
    - priorityID (lowest -> highest for elements ordering)
    - nameOfComponent : this is important since it is the name of the prefab that this UI will use (the only option currently is "UI/TextPerformanceComponentUI".  But more could be implemented later (like graphs or idk, any UI element/s)
    - UIComponent : this again is VERY important since its the 'sub-class' job to store the BasePerformanceComponentUI (the element to display) and then return it when prompted, yes this seems odd but its here so the compiler forces you to create this function (else if it wasn't abstract it wouldn't force you and you can't have abstract properties), other than forgetting and having it do nothing by returning a null value or so.  You get your first and only copy of this in the Start virtual function.
    - Update: Just a function that is called once per frame for you to do updates
    - Start: Just a function for initial setup and it also passes your UIComponent that you need to store to return whenever UIComponent is called.

That's it.  Like really that's it... The BasePerformanceComponentUI class has no properties and is an abstract just for sub classing...

Now personally I like subclasses, HOWEVER I understand the glory of protocols/interfaces and I'm happy to implement that if that's deemed to be more 'friendly' or whatsoever.

### Testing

Make sure you test all the options in settings, (currently apply doesn't work but due to how apply normally works it should just do the same thing as save but not quit... so I need a save to settings for it to work)?

### Files Removed

All of standard assets (since it just contained FPS Counter and FPS Counter Prefab) since they are no longer needed due to new system


BraedonWooding:
Be it gets merged would like people's opinions on adding a colour option to the standard fps (with green being 50+; yellow being 30-50 and red < 30) as well as making the apply button in DialogBoxSettings actually save the settings (else the HUD won't update- it retrieves its current from settings) or to use a variable of the selected one in DirtyUI as an 'apply' feature?  As well as general look, font, colours, and so on.  Also, do you like it as [0]: AVG or AVG: [0] ? For all of them like I did some the first way and some the second tell me which one you like better.

Assets/Scripts/UI/InGameUI/PerformanceHUD/FPSRangePerformanceComponent.cs
```diff
@@ -0,0 +1,78 @@
...
+    private int fpsAccumulator = 0;
+    private float fpsNextPeriod = 0;
+
+    private int currentFps;
+    private int lowestFps = 60;
+    private int highestFps = 60;
+
+    private TextPerformanceComponentUI component;
+
+    public override int priorityID()
```
bjubes: All methods should be PascalCase, so `PriorityID`.

Make sure you run [stylecop](https://github.com/TeamPorcupine/ProjectPorcupine/wiki/StyleCop) on your changes, it will show what code doesn't conform to the project's style guidelines

BraedonWooding: Ooops yep, okay I will :D

koosemose:
Please don't add anything along the lines of [Ready To be Merged], if it's not WIP or Conflicted, it will be looked at by a Gatekeeper whenever they're available.

koosemose:
It's kind of hard to clearly read the FPS at times, rather this requires a larger font (as the previous fps had), a different (more clear) font, or a brighter green for OK level, I'm not sure.

Also what is 0: AVG supposed to be? It never changes, I would think maybe average, but would also expect it to do something, so maybe AVG is short for something else?

Also, I notice that this seems to change much more rapidly than the old one, how intensive is calculating FPS, and does it really need to be faster than the old one (Or perhaps this is just a matter of perception, since I can't easily watch them both side by side)?

koosemose:
Ahh I see it changing now, just didn't watch it long enough, perhaps before it has 30 seconds of frames it could show an average with what it has so far? It seems strange for it to have 0 up until then.

And the <number> :<What it is> (such as 0: AVG) seems odd to me.

And (while I realize the previous had it, but of course we want better than what it was) is it necessary to have FPS after the fps? I would think that anyone using it would realize what it is. 

And I don't know for sure if just increasing to the old size will be enough, I'm just guessing at potential options, it's only a problem when it's green (that green doesn't stand out very well when it's overtop of walls and floors)

BraedonWooding:
I will make it so its AVG: X and so it does 0.5 seconds but keeps building on top of it so its a different formula, which means it updates itself more but is still representative over the last 30 seconds.  I think I'm going to make it 18 font (the old size) with a cleaner font and a brighter green.

koosemose:
All three should hopefully do the trick (My recording software uses a full green for FPS and that seems to be consistently visible)

BraedonWooding:
Yep :D.  I made the changes, it should match the criteria now :smile:

BraedonWooding:
Just making a new one so its not on master (I did a stupid)

PR #1561, opened 2016-10-19
Main Menu Button

Ohda:
- Add a new menu when you clic on Quit. So you can quit to main menu.


bjubes:
can you upload a picture?

koosemose:
Perhaps it's just me, but I think just having a Quit To Desktop and Quit To Main Menu button on the options menu, rather than having to use click on a separate dialog after choosing quit

Ohda:
I can do that if you want, but it was feeling good to me to replace the "are you sure you want to quit panel" by this.

BraedonWooding:
I actually quite like this window its a better alternative then the are you sure window.

dusho:
ye.. if there are more than 2 options, it shouldn't use OK/Cancel layout maybe..
maybe just having:
**Resume
Main Menu
Quit Game**
vertically would be better?

Ohda:
I will try something like that tonight.

Ohda:
Update the original post

ComputerOverlord:
Why are you yelling at yourself to update it?

koosemose:
Looks good, stylecop clean, and everything seems to work as expected. Merged.

PR #1565, opened 2016-10-21
[Small, Fix] Delete Metas, reconnect more prefabs

koosemose:
Several Metas were left over from deleted/moved files, along with a few broken prefab connections. This reconnects


koosemose:
As this is just a small fix sorting out Unity metas and a disconnected prefab, I am self-merging.

Issue #1488, opened 2016-10-04
ProjectSettings and Build settings


bjubes:
As 0.1 nears, I thought now would be a good time to put an offical version number on the project as the bundle identifier. However we currently gitignore `/[Pp]roject[Ss]ettings/ProjectSettings.asset` which contains information including:

- build version
- Company Name
- Application Name
- Game Icon
- Whether fullscreen/native resolution are on by default (we keep track of this in settings.xml, so no real worries here)

As a result, none of this information exists on git and can be changed in the official project, only locally. Im sure ProjectSettings is ignored for good reason (please state it if you know specifically), but how are we going to ensure that the above stays consistent when the game is built if its not tracked by git?

TomMalbran:
Adding ```/[Pp]roject[Ss]ettings/ProjectSettings.asset``` to ```.gitignore``` came from issue #17 and then was added in a PR by Quill.

bjubes:
I remember that issue #96. The problem is that Projectsettings contains information we want ot standardize as mentioned above, but also contains information (i think its the window layout or something like that) which we don't want in the repo. Sadly there is no way to ignore half of a file.

mikejbrown:
Makes sense to try figure this issue out before v0.2, so added it to the milestone.

koosemose:
From what I can tell ProjectSettings.asset is almost entirely information we want to be able to standardise, the only thing that it did and I think was the reason it was ignored in the first place was of someone makes changes using a different version of Unity, and we should really try to be using a consistent version of Unity as is, and if someone isn't it should be up to them to not commit their projectSettings and us GKs to try not to merge PRs with accidentally added ProjectSettings (there should be minimal situations in which we need to actually change it, mostly company/product data, and api compatibility, and those only need to change once).

I move that, if no one disagrees, we re-add ProjectSettings. Thoughts on this @mikejbrown  and @bjubes ?

bjubes:
does it also include editor window layout and size? I thought that was an issue as well but I could be wrong

koosemose:
From my understanding it doesn't, it just stores things you set in Edit->Project Settings, plus a productGUID string which may be specific to each unity version. And looking through the file, I can't see anything that suggests layout to me, the closest is defaultHeight and width, which I presume are the default resolution for a standalone build

bjubes:
I'm good with adding it. worse comes to worse we just remove it again later.

bjubes:
[...]

PR #1566, opened 2016-10-21
Change Save format to Json

koosemose:
Per the decision laboriously reached in #1034 this starts the switchover to Json, by including the Json.Net library ( http://www.newtonsoft.com/json ) and changing loading/saving games to use Json. Tests that previously testing saving related xml reading/writing have been updated to use the appropriate Json based methods. Other than the file format saving/loading should behave as it did before.


koosemose:
As far as I can tell Travis is failing to build at all because Json.Net requires Api Compatibility Level of .NET 2.0 (rather than subset), and it's not being built with that, I'm not sure if this is simply because git doesn't have ProjectSettings, or if when running it from the commandline you have to specifically tell it to use .Net 2.0 somehow.

On a related note, not including ProjectSettings means we can't set the project to use .Net 2.0 in general, so any one developing will have to explicitly set the Api Compatibility level on their local ProjectSettings.

dusho:
Why not to use serialization, but explicitly json objects to serialize/deserialize? I would assume save/load files would be some binary format in the end. And using serialization would enable us to choose any formats (serializers - binary, xml, json, ...). Generally you don't want people to poke around your save files with text editor...

koosemose:
It was an intentional decision in the beginning to have save files be non-binary. I disagree about not wanting people to poke around in save files, it opens up a lot of game play opportunities for those interested in hacking around with things (see things people do with Kerbal Space Program saves as examples). Due to the way some of the data is stored (particularly anything involving colors) autoserialization fails miserably, getting stuck in an infinite loop. Doing it this way allows me to closely mirror the data that is presently saved in xml. If, after this gets in, you or someone else wants to start changing things over to automatic serialization and can do so without running into the looping issues I was and keep the same data saved, feel free.

Of course this all may be moot if we can't get travis to a state that it can run tests with Json.Net...

koosemose:
Actually, using autoserialisation/deserialization would be quite bad for save files, it makes them more prone to breaking from a change to the structure of an object. Also, doing it a more direct fashion allows us to only save what we need to save, and rather than deserializing something into, for example, a furniture, we can get the prototype of the furniture, and just set whatever variables need to be changed from the prototype (mostly params, location and rotation), allowing a save to automatically use new versions of objects rather than being stuck using whatever version it was when it was saved.

koosemose:
@bjubes I'm betting your api compatibility level is .NET 2.0 Subset, because that's the error thrown when it is, which is the exact problem Travis has, and where we don't have the project settings (which stores that information) in git, even though I set it to .Net 2.0 it won't be that for anyone else, which is likely why you can't compile either.

bjubes:
so this another drawback in #1488

koosemose:
Closing and reopening to force travis checks to be reran.

bjubes:
Are the save files supposed to be in json now or just support it? They seem to still be xml

koosemose:
Any new save files should be Json... and are for me, and it is no longer capable of loading xml save files.

Are you sure you're looking at a new save from this PR?

bjubes:
I think when i tested I had an unsaved change which cancelled the merge? Whatever happened Its now working properly for me.

koosemose:
It may have been the sound PR, switching back and forth between branches with and without the DLL causes some issues, because Unity doesn't unload the DLL when exit play mode, so git can't delete it when you change branches.

PR #1569, opened 2016-10-21
Dev console

BraedonWooding:
_Will close #1562_
### What this PR does

It adds the ability to have a command line like developer tool.  It's really easy to create commands and it allows parameters to be passed (any data type :D, including primitives, arrays and vectors).  I'll put a section on how it works/how to add new commands below.  It will in the end allow LUA/C# commands (C# will come once C# modding is implemented).

Note: it uses a private singleton instance to manage statics (kinda similar to performanceHUD), it should be completely fail-safe and it checks instances before doing anything.
### Images

1) General Look
image
2) Autocomplete and text
image
3) Developer Settings
image
### How does this work/How to add commands

Adding commands is easy, it's a single function call (you can call AddCommands to add multiple or just AddCommand to add a single one).  

There are multiple different ways to add commands:
1. Title, Method, HelpText
2. Title, Method, HelpMethod
3. Method, HelpText,
4. Method, HelpMethod
5. Title, Method
6. Method

Title is the title of the command (what you type to run it), Method is a delegate function that is run (passing the parameters asked), HelpText is help to display (when the user types Help or x? with x being the title of your command), if you don't include title it'll use the name of the method, HelpMethod is an alternative to HelpText (it's just like method and you can use console log commands to log your help info).

So as you can see it's got no bounds in uses.  However!  There is a limit on how many parameters you can have _kinda_.  Using the standard Command<T0, T1, T2>(...) structure you can have a max of 3 parameters.  The code allows for what I call 'objects' essentially structs/classes through the use of an array indexer '[...]' with ... representing the variables in between, all it does is just return a string, for example if you have [x, y, z] it will return [x, y, z] rather than x, y, z as separate objects, these are obtained from the constructors of the types you pass if they type [], so essentially [] is a constructor call.  So if you say Command<Vector3> then the passed argument [x, y, z] will be automatically converted to a vector3.  The only command available for LUA is LUACommand which passes an array of objects this allows you to technically have more than 3, however they won't be auto typed (moonsharp will autotype some of them however).  To create a LUA command you need to edit the XML document (which will later become a JSON one) and add a command; the fields are:
- Title: string = the title of the command (what to call to run)
- FunctionName: string = the name of the command to run (via the consoleCommands.LUA script)
- Description: string = what does this command do (for the Help :: function)
- HelpFunctionName: string = (optional) what function to call when Title? is called, should mostly just log to console.  If this is "" or left out it will just log Description to the console.
- Parameters: string = a string consisting of all the parameters (should use C# or just general C standard names such as Int, Float, Bool, try to avoid Int16, but definitely don't do Number, Flag and so on), each parameter needs to be separated by a ',' cause it will print them out when either Help or Title? is called so the user knows how to call the function.

To call a command is easy: you first type the command name in the developer console (you activate it by '\', however in Test_DevConsole its automatically activated), then you follow it by a modifier; there are two types the first being an info or help modifier and the second being a parameter modifier.  The help modifier is "?" and then you press enter to retrieve help info about the command.  The parameter modifier is for adding parameters, its "(", and then you follow it with the parameters that are required for that function (with a comma between parameters), then you end the command with ")".

NOTE: objects are a CSharp feature only, you can't do them in LUA due to the fact that they require static typed variables
- Also note objects are single dimensional aka in a constructor call you can't create more objects
E.G.
runParameterlessFunction ()
runParameterlessFunction

runVectorFunctionCSharp ([30, 40, 25]) 
// Creates a vector3 since the runVectorFunctionCSharp definition states it wants a vector3
// If it defined that it wanted a vector4 or any other type that can take 3 numbers then it also would work.

runVectorFunctionLUA (30,40,25)

runRandomClassFunction( [ 30, "Bob Likes Tomatoes, But Jill likes Cats", 20.0 ]
// This will create an object that the class asked for with parameters of a number (it will cast it to the one wanted) a string, and another number.

runFunction ? 
runFunction?

Note: White spaces are ignored (the command name can't have white spaces however, but you can have as many or as few white spaces between variables or between the command name and the modifier.

### Testing

You can obtain a list of all the functions via the command Help, which will return all the functions, don't type Help? since that'll return info about the function Help hehe.
### Controls?
- The open/close key for the console is a keyboard mapped key which is defaulted to Ctrl+'~'.
- To navigate the history (exact same as like terminal history) you just use the arrow keys but autocomplete keys take priority.
- To activate autocomplete press tab; this opens up a list of all the options (second image under image section), then you can press tab again to select the currently highlighted option (yellow), or use arrow keys to navigate the choices, this window will update as you type.  You can close autocomplete window by either pressing backspace or escape.
### Progress
- [x] Add better information on what parameters that function requires
- [x] Add better error support
- [x] Add the ability to change the command input key to a few presets
- [x] Make everything interfaces
- [x] Add History (and enable going back and recalling)
- [x] Add autocomplete
- [x] Add LUA modding support
- [x] Make the colors look good (they currently clash but that's just cause I randomly threw it together)
- [x] Make it work in the normal scene
- [x] Massive cleanup
- [x] Allow 'objects' and improve the speed of processing and fallibility monitors
- [x] Add unique commands for PP
- [x] Add C# Modding
- [x] Fix Conflicts
- [x] Re-nable old Optimisation, and some preliminary
- [x] Optimise (didn't before, cause "Optimisation is the root of evil" kind of thing)
- [x] Make it so that the parameters xml file, will convert the types (info below)
- [x] Cleanup Documentation (tbh this isn't a priority, and it can be pulled in later as it is just moving around documentation, since I put it close to summaries for myself, then remove it later)

### Parameters for modding
- Due to how the C# modding (it being invoke only) was implemented, along with LUA being invoke only.  I can't grab parameters and handle them that way since I can't get a specific reference to them (without a ton of code that is kind of wasteful and slow), so what I will use is the xml file which already has parameters for the displaying upon `FunctionName?` so we can use it for this as well.  They will have to be in format: `Type0, Type1, Type2...` and so on.  I'll have to use reflection, but this will occur on startup so won't affect performance.

### Major Changes
- There is a new settings dialog box (in addition to the old one) just for developer options.  (There are screenshots below), this moves the old PerformanceHUD option as well as DeveloperMode to the new Developer Settings (accessible from the general settings menu through a button called "Developer Settings").
- Added a new class to handle all of the settings for Developer Settings, so that instead of calling:
  `Settings.GetSetting("DialogBoxSettings_developerModeToggle", false)` you now call `CommandSettings.DeveloperModeToggle`.
- Added "DevConsole" LuaFunctions to FunctionsManager
- LuaFunctions now also registers all of the UnityEngine Vectors (2, 3, 4) as well as DeveloperConsole.DevConsole (to allow access to log commands to the console or edit stuff)
- Added UnSafe variants to RunText and Call, both of these methods will not catch errors (so the caller should do a try/catch) and this is needed for Console so we can then process and log the error to tell the user why it failed (not right parameters, syntax or whatever).
- ModsManager has new Prototype called ConsoleCommands.xml.
- KeyboardManager now has UnRegister commands (for inputAction and inputField)
- Settings.cs has a few minor changes to support the new DeveloperSettings (e.g. DialogBoxSettings_performanceGroup is now DialogBoxSettingsDevConsole_performanceGroup).  Also fixed a bug with Autosave interval not being closed.
- DialogBoxManager creates the developer settings dialog box but doesn't store it (it passes it to dialogBoxSettings -general, for it to show it)
- DialogBoxSettings has some toggles/dropdowns removed (developerMode/performanceHUD) to move them to DialogBoxSettingsDeveloper and it also calls DialogBoxSettingsDeveloper when the developer settings button is pressed.
- ModUtils now has LUAVectors (2, 3, 4) to allow you to create vectors without creating a unity vector.
- KeyboardManager now has a function called TriggerActionIfValid that will trigger the action for the supplied input action (you pass a string of the name) if it's valid, thus ignoring the inputField.isFocused thing, this is important to allow DevConsole to close.


koosemose:
I'm consistently getting an NRE
```
NullReferenceException: Object reference not set to an instance of an object
AutosaveManager..ctor () (at Assets/Scripts/Models/Events/AutosaveManager.cs:48)
WorldController.Start () (at Assets/Scripts/Controllers/WorldController.cs:104)
```

BraedonWooding:
I'll have a look, it may have been the mac changes, that broke things.

koosemose:
Ok, now that things have been assorted, my initial thoughts:

Visually it looks kind of unweildy and chunky, and the blues don't seem quite right particularly the cyanish color of the actual console (the part that shows output rather than where you do input), some of those colors are somewhat difficult to read on that background (pretty much any color other than white).

Personally i'd prefer that the input are didn't look so strongly distinct from the output area (more of a separation by something along the lines of a line, and having the same coloration).

For usability it feels like the input should have focus when you bring the dev console down (though of course that would make it slightly more awkward to dismiss if dismissal stays on '\', but in general on non-character key may be better, at least for dismissal)

Also after entering a command, it loses focus, which makes it really awkward to use.

BraedonWooding:
Okay, so colour wise; I'll fix it up (I kinda threw it together), how about I make it so its always in focus (the input field)?  And if you enter the dismissal key (you get a few choices rather than any key; choices like '`' (the ~ key), '\'.  Choices that aren't used in commands at all) then it'll close the console?

koosemose:
I like the always in focus, but I'm not sure if just allowing keys that aren't in commands won't cause issues (e.g. what if I were using the enterLuaCommand command (or whatever it's called) and needed to use backslash? 

Maybe only do that if it's the first and only character? or have an explicit exit command?

BraedonWooding:
I could always do a Ctrl or Alt + Key to exit?  Like then there is 100% no way for the user to accidentally do it ?

koosemose:
Also I'm not a huge fan of the multiple "modes" of entering commands, it just seems like it will lead to confusion, and seems like options that don't really add anything useful. I'd suggest settling on one mode and having that as the only one (my suggestion would be for the Command(args) style as that most closely resembles the two languages we already use (C# and Lua, of course)

koosemose:
That seems reasonable.

BraedonWooding:
The multiple modes are actually really useful though for C# (LUA only has one mode, which has no restrictions on how many parameters since we don't convert them to the types we just split it and pass it), the reason is that you may want a Vector value so instead of you having to build the vector yourself from the 3 parameters and having to raise an error if there aren't three values, the system takes care of that and won't call your function if it can't build the vector.  And the 3 commands for each parameter is just so you can have one, two, or three (I could add more but I doubt you'll ever need more than three), and the string command is kinda needed in some instances like changing the name of someone you may want commas in that or whatever?  Like each function has its own use.  So essentially LUA is great if you want to parse a large amount of commands (like 4+) or if you rather it, while C# has all the parameters passed as types already and is technically a little safer (cause we can't know what types the LUA script wants so its a technical limitation)?  Like the multiple commands are to improve on the strength of the two languages.

koosemose:
I'm not sure I understand how being able to choose between parenthesis, brackets, braces, or a colon to separate a command from its arguments helps.

BraedonWooding:
I just added that since I personally prefer ':' for some reason and others may prefer '()' and so on.  If you call function? then it'll inform you of how to call it using your currently selected option?  So I don't see how it could be detrimental?

koosemose:
It's mostly a matter of consistency, if a dev (or modder in the future) does a document along the lines of "How to do X" or "How to use dev console to improve your deving/modding", there will either end up having to be an explanation of the different ways of giving commands, or just assuming one way, then someone who has a different way set will have to either translate any commands referenced or try to figure out why things aren't working as expected. There's a reason virtually any other console (either an OS console or a game console) don't offer settings for multiple ways to enter commands, it adds nothing but the potential to fragment support.

BraedonWooding:
Okay sure cause of consistency I'll remove it.  So far we have decided on re-doing graphics/UI, removing the ability to change input modes; Ctrl or Alt + the key to exit/enter command line, with the input field always being in focus while command line is open.  Also I'm sorry I misunderstood your initial comment about the command modes I thought you meant like only have one Command<T0> instead of like the three and the string one and the vector one.  My bad, regardless; Is there any more issues?

koosemose:
Is it possible to make the autocomplete resize to the number of commands in it (it looks weird when it pops up with only one command).

I think that covers everything I've noticed. 

I must say though, despite all my critiques, I find this quite impressive.

koosemose:
Conflicted

BraedonWooding:
Haha thanks, yeh its quite cool.  I can make it resize and I will :D (hopefully, I should be able to atleast).

BraedonWooding:
I'm assuming from the three thumbs up that this has received an 'all okay' feedback?

BraedonWooding:
Note: I just found a slight weird thing with string literals and brackets, fixing that.

bjubes:
Thumbs up is more of a "good idea/approach" type of thing, usually when people test it they leave a comment indicating so.

bjubes:
Im getting this error repeatedly when loading a new world:
```
Attempted to access a setting that was not loaded from Settings.json:	DialogBoxSettingsDevConsole_developerModeToggle
UnityEngine.Debug:LogWarning(Object)
UberLogger.Logger:ForwardToUnity(Object, LogSeverity, Object, Object[]) (at Assets/UberLogger/UberLogger.cs:505)
Debug:ULogWarningChannel(String, String, Object[]) (at Assets/UberLogger/Debug.cs:278)
Settings:GetSetting(String, Boolean) (at Assets/Scripts/Models/Settings.cs:127)
MouseController:Update() (at Assets/Scripts/Controllers/InputOutput/MouseController.cs:133)
MouseController:<MouseController>m__20(Single) (at Assets/Scripts/Controllers/InputOutput/MouseController.cs:61)
TimeManager:InvokeEvent(Action`1, Single) (at Assets/Scripts/Models/Events/TimeManager.cs:198)
TimeManager:Update(Single) (at Assets/Scripts/Models/Events/TimeManager.cs:125)
GameController:Update() (at Assets/Scripts/Controllers/GameController.cs:77)
```

BraedonWooding:
Yep thats a super quick fix, I must have forgot to update the Settings.json file after it changed from xml.

BraedonWooding:
Okay @bjubes, the json file does have that setting on my side, and I'm not getting that error?  Can you confirm that it has 	'DialogBoxSettingsDevConsole_developerModeToggle' : false under Settings.json.  On your side?

bjubes:
Its only before I open the dev mode dialog and change it, so i guess its a incompatibility between the old settings.json from master and your PR, which is fine.

on another note, SetFontSize(18) is returning an error in the console: `cannot access internal field SetTextSize of userdata<DeveloperConsole.DevConsole> `

BraedonWooding:
Oops, yep that was just some LUA code trying to access an old function (it used to call the CSharp variant of SetFontSize, now it just gets the text object and sets the size that way), it works now for me :D.  All fixed now :D

bjubes:
Im getting the a similar error for `ShowTimeStamp(false)` : cannot access field ShowTimeStamp of userdata<DeveloperConsole.DevConsole>

BraedonWooding:
Which I just fixed hehe, the way it was doing it was hacky anyway.  LUA isn't really meant for any directive based commands (like changing settings or using instances) since well you would need to register a lot of types to be able to access them.

bjubes:
I got that error after pulling In your latest commits.

BraedonWooding:
I pushed new changes that fixed it for me like 10s before your comment. Can you try to redo a pull?

bjubes:
Ive tested this for a while functionality wise and it seems to work very well, with the exception that too many messages in the console without clearing slows the game down a bit.
I think this is ready to be merged, but being so big its hard to thoroughly review all of the code and test every edge case. Hopefully anyone who has time can test this out a bit or look over some of the code, as well as @koosemose and any other gatekeepers.

koosemose:
I've mostly been waiting to do any kind of more involved look over until everything is checked off.

I think the slowdown from too many messages is something that needs looked at, tracking down exactly what part of how the messages are handled is causing it.

bjubes:
So a few things:
why is Assets/Resources/UI/black-and-white-city-skyline-buildings.jpg  included? was that by acccident. Also is Desktop.ini needed as well and whats its purpose?

I would also prefer that you don't put dots in folder names either, since it gets confusing whether they are folders or files.

hcorion:
desktop.ini is a frivolous addition, it would be good if you deleted it. I was very surprised when I saw the amount of useless Info messages when I opened ProjectPorcupine. I was using some debug messages in a VR game and it *drastically* dropped the frame rate.

BraedonWooding:
Yeh, slowdown exists but is only cause I was testing something, it won't exist in the next version. The error messages are sadly harder to make clearer, but the only ones you should see would be the wrong parameters (the others are when runtime errors occur like referencing).  I just want to make it a little faster (and changeup the structure), since how C# modding was made it meant that it was harder to present parameters since the C# stuff is invoking so I can't grab the method and do reflection.  Tbh i have no clue why those files are there, ill make sure I delete them, I did a merge recently to fix conflicts so maybe that's how they came to be?  Regardless yeh I'll remove.  Sorry for the long time with this, it just faced a few iterations to improve the structure, since I was a little bit to much of a perfectionist.

BraedonWooding:
@bjubes When you have time can you test the speed of the console?  I just tweaked a few settings, and some better optimisation on type casting.  It should make it faster :D.

bjubes:
Its much faster now even when spamming help and then changing font size.
The scrolling is kinda slow however(at least for me), maybe just increase the sensitivity a bit?

BraedonWooding:
Easily done, I'll increase it a little.

BraedonWooding:
@bjubes.  It's now a developer settings option, and I increased the default.  For me it's about perfect but I like it slower, so you can make it faster (with 15 being the max which is 3 times the old speed, so a large range).  I'm happy with it and it's all ready to be merged unless any more issues persist :D

bjubes:
You have a merge conflict again. 
Other than the scrolling, I think this is ready. Ill let this sit for one or two more days so other gatekeepers and anyone who wants to help test can look over the code and test themselves.

BraedonWooding:
Ugh, conflicts.  Okay all fixed :D

PR #1571, opened 2016-10-22
Power through utilities

koosemose:
Currently Power is universal, this changes power to travel along Utilities. At the moment, all utilities carry power, of course this isn't what we want in the end, however at this stage (without a fluid system or anything else to use for other utility types) are to just have every Utility carry power (as I've chosen to do for the time being), or in multiple places have special cases where the utility type is checked and only has the network built if it's a power cable.

A furniture should connect to a utility if it is placed on top of that utility, or an event is triggered if a utility is placed in the same tile as a furniture to allow it to plug in. The utility network should be built for a given utility (again, at this time only power, and all utilities act as power cables) such that all utilities that are visually connected share the same grid/network, if two visually separate utility networks are connected (by adding a utility of the same type connecting them) they should join and share power as would be expected. If a network is split, the two visual halves should end up with separate grids, and furniture connected to one or the other should only be able to draw from power available on that network.

There is currently an issue where when grids are split or joined connected furniture "flickers", I'm not certain if this is purely a visual thing or if they are starting/stopping working for a moment. I'm also not certain at this time how/if I can avoid this, if anyone has any thoughts, please let me know.

Note: This is a resubmit of #1537, due to some issues with the branch

I consider this in a complete state for this stage of progress, further changes needed involve expanding grid to handle multiple types of grid (fluid grid, and such), and there being other types of utilities (for not Water Pipes are effectively just an alternative power cable. A fluid system of some sort will need to be in place to expand grid in any reasonable way.


koosemose:
Please remember to comment when a PR is conflicted @bjubes there's apparently no notification on a title change. Only reason I noticed it was because I expected the conflict.

bjubes:
oops. I think I hit back before it actually posted...

koosemose:
Conflicts resolved, and fully integrated with #1520.

Assets/Scripts/PowerNetwork/Grid.cs
```diff
@@ -35,6 +35,21 @@ public bool IsEmpty
             get { return connections.Count == 0; }
         }
 
+        /// <summary>
+        /// Gets the connections to this grid.
+        /// </summary>
+        public HashSet<IPlugable> Connections
```
dusho: why do you need to expose this. with this, anyone can modify set from outside
 it's used only in UTs now

Assets/Scripts/Models/Buildable/Utility.cs
```diff
@@ -257,21 +269,37 @@ public static Utility PlaceInstance(Utility proto, Tile tile)
...
+            obj.UpdateGrid(obj);
+        }
+        else
+        {
+            // If we're skipping the update, we need a temporary grid for furniture in the same tile to connect to.
+            obj.Grid = new Grid();
+            World.Current.PowerNetwork.RegisterGrid(obj.Grid);
+        }
+
+        if (obj.Tile != null && obj.Tile.Furniture != null && obj.Tile.Furniture.GetComponent<PowerConnection>("PowerConnection") != null)
```
dusho: maybe create IEnumerable<IPlugable> GetPlugableComponents() on Furniture class and use it here

dusho:
I was not able to make this work.. I put cable from RTG to oxygen generator, but generator was still without power
Also getting 999+ exceptions:
```
InvalidCastException: Cannot cast from source type to destination type.
Job.CancelJob () (at Assets/Scripts/Models/Job/Job.cs:337)
ProjectPorcupine.State.JobState.AbandonJob () (at Assets/Scripts/State/JobState.cs:106)
ProjectPorcupine.State.JobState.Interrupt () (at Assets/Scripts/State/JobState.cs:90)
ProjectPorcupine.State.State.Interrupt () (at Assets/Scripts/State/State.cs:47)
ProjectPorcupine.State.HaulState.Update (Single deltaTime) (at Assets/Scripts/State/HaulState.cs:64)
Character.Update (Single deltaTime) (at Assets/Scripts/Models/Character/Character.cs:347)
CharacterManager.Update (Single deltaTime) (at Assets/Scripts/Models/Character/CharacterManager.cs:95)
World.TickEveryFrame (Single deltaTime) (at Assets/Scripts/Models/Area/World.cs:493)
TimeManager.Update (Single time) (at Assets/Scripts/Models/Events/TimeManager.cs:131)
GameController.Update () (at Assets/Scripts/Controllers/GameController.cs:77)
```

Assets/Scripts/Models/Buildable/Components/PowerConnection.cs
```diff
@@ -206,10 +218,14 @@ protected override void Initialize()
                     CurrentAccumulatorChargeIndex.SetValue(0);
                 }
             }
-            
+
             IsRunning = false;
 
-            World.Current.PowerNetwork.PlugIn(this);
+            foreach (Utility util in ParentFurniture.Tile.Utilities.Values)
```
dusho: just call OnReconnecting() ?

koosemose:
Regarding having to connect to 2 tiles of furniture, it's the "Actual Tile" of the furniture that matters, I need to get furniture to look in all of its tiles for larger than 1x1 furnitures. The InvalidCastException is a standing bug on master, but I will most likely look to fix it in a separate PR.

koosemose:
I'm not absolutely certain this still continues to work properly in all cases after conflict resolution, please don't merge until I've had time to test it properly

koosemose:
Ok, everything should be good again.

bjubes:
you have merge conflicts again.

koosemose:
Unconflicted.

koosemose:
Whoops.. Unconflicted this a day ago..

koosemose:
Merged

PR #1572, opened 2016-10-22
[Fix] Fixes power icons update on furniture placement

dusho:
Fixes issue when power icon was not appearing after placement of new furniture (initialize) and there was lack of power. Machines were not operational though (correct behavior).. so only icons were not synced.
image


koosemose:
Works great, bug no more. Merged

PR #1573, opened 2016-10-22
[Fix] Fixes context menu for workshops with 1 prod. chain

dusho:
Fixes 'Argument cannot be null' when trying to get context menu on workshops with one production chain (power generator, power cell press)


koosemose:
Solves the problem, no Stylecop issues. Merged!

PR #1574, opened 2016-10-23
Readd ProjectSettings.Asset

koosemose:
Per discussion in #1488 this readds ProjectSettings.Asset, which will allow us to set build version, Company Name, Application Name, Game Icon, and api compatibility level.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@
 /Assets/StreamingAssets/Data/Mods/*
 !/Assets/StreamingAssets/Data/Mods/README.md
 !/Assets/StreamingAssets/Data/Mods/README.md.meta
-/[Pp]roject[Ss]ettings/ProjectSettings.asset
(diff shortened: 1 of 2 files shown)


NogginBops:
How will different unity versions be handled with this?

dusho:
isn't Unity version held within ProjectVersion.txt?
I have there:
```
m_EditorVersion: 5.4.0f3
m_StandardAssetsVersion: 0
```

koosemose:
If there is anything that changes based on version used to open, we'll just have to rely on people not committing it if they don't intentionally change it, there's just too many settings we need to have set consistently for the project to leave it out.

NogginBops:
@koosemose @dusho :+1:

bjubes:
Going ahead with this merge. Project settings is needed to consistently set settings within unity. Unfortunately some other information such as editor version can also make its way into the file. If we standardize on a unity version, it shouldn't be an issue.

Issue #1452, opened 2016-09-26
ArgumentException in Tile::PlaceUtility when placing furnitures already installed


Ohda:
There is a bug when placing some utilities furnitures.
Some caracters can try to call place some furnitures already installed.
This throw an error :
Utilities.Add(objInstance.Name, objInstance);
ArgumentException: An element with the same key already exists in the dictionary.
This happen pretty often ( 1 per 20 mabye.)
This was already on master before https://github.com/TeamPorcupine/ProjectPorcupine/pull/1350

`Tile.PlaceUtility (.Utility objInstance) (at Assets/Scripts/Models/Buildable/Tile.cs:237)
Utility.PlaceInstance (.Utility proto, .Tile tile) (at Assets/Scripts/Models/Buildable/Utility.cs:247)
UtilityManager.PlaceUtility (.Utility prototype, .Tile tile, Boolean doRoomFloodFill) (at Assets/Scripts/Models/Buildable/UtilityManager.cs:61)
UtilityManager.PlaceUtility (System.String type, .Tile tile, Boolean doRoomFloodFill) (at Assets/Scripts/Models/Buildable/UtilityManager.cs:49)
UtilityManager.ConstructJobCompleted (.Job job) (at Assets/Scripts/Models/Buildable/UtilityManager.cs:86)
Utility.<ReadXmlBuildingJob>m__2B (.Job theJob) (at Assets/Scripts/Models/Buildable/Utility.cs:469)
Job.DoWork (Single workTime) (at Assets/Scripts/Models/Job/Job.cs:263)
Character.Update_DoJob (Single deltaTime) (at Assets/Scripts/Models/Character/Character.cs:730)
Character.Update (Single deltaTime) (at Assets/Scripts/Models/Character/Character.cs:347)
CharacterManager.Update (Single deltaTime) (at Assets/Scripts/Models/Character/CharacterManager.cs:95)
World.TickEveryFrame (Single deltaTime) (at Assets/Scripts/Models/Area/World.cs:537)
TimeManager.InvokeEvent (System.Action1 eventAction, Single time) (at Assets/Scripts/Models/Events/TimeManager.cs:198)
TimeManager.Update (Single time) (at Assets/Scripts/Models/Events/TimeManager.cs:131)
WorldController.Update () (at Assets/Scripts/Controllers/WorldController.cs:144)
`

koosemose:
[...]

Issue #1513, opened 2016-10-06
Localize _all_ of the UI elements


mikejbrown:
From @bjubes over at #1475:

> Now its apparent that a lot of the UI skips the localization table, such as the mouse over dialog, so only some strings are reversed. Its most clear when mousing over furniture, the word "furniture:" is reversed since its defined in a lang file, but the actual furniture name isn't ever interacting with localization at all.

## Missing localizations

- [ ] Furniture names in construction menu (Test: mouse over dialog with reversing on)

If you find any more instances of missed or unapplied localizations say below and I'll add it to the list.

bjubes:
also adding to this, some things are partly localized. for example in the mouse over dialog the text "Furniture: Steel Wall" appears, but only "Furniture" is passing through the localization system; Steel wall is outputted directly from Furniture.Name so it will not be localized. This specific instance is partially fixed in my PR but many still exist in the project

TomMalbran:
We should also localize all the construct and deconstruct job descriptions. The description key is there but is a different one for each furniture instead of a more generic one and is no in the lang file.

koosemose:
#1765 is localized now

PR #1575, opened 2016-10-24
Improve system of reserving tiles for construction jobs

koosemose:
Currently on master, pending build jobs on a tile aren't well suited for queuing up multiple build jobs on the same tile (such as building a furniture and a utility or building multiple utilities) as pendingBuildJob only holds one job. This leads to several errors (such as #1452).

This PR changes pendingBuildJob to pendingBuildJobs, a HashSet of all jobs pending on the tile, and adapts the DoesBuildJobOverlapExistingBuildJob (as DoesFurnitureBuildJobOverlapExistingBuildJob) and implements DoesUtilityBuildJobOverlapExistingBuildJob allowing both types of build jobs to overlap as intended (furniture doesn't overlap, and cancels a previous furniture build job if it is allowed to replace a furniture of that type, utility can only have one utility of that type per tile). 

As a side effect, this has fixed a reversion that came about in #1416, where furniture could no longer replace a furniture type it was supposed to be able to replace (due to that PR making a clone of the prototype, which didn't previously copy ReplacableFurniture, as it wasn't previously needed), by copying ReplacableFurniture in the clone constructor.


Assets/Scripts/Models/Buildable/Tile.cs
```diff
@@ -561,32 +561,31 @@ public IEnumerable<string> GetAdditionalInfo()
...
                     {
-                        c.PrioritizeJob(PendingBuildJob);
+                        pendingJob.CancelJob();
                     }
                 };
+                if (!pendingJob.IsBeingWorked)
+                {
+                    yield return new ContextMenuAction
+                    {
+                        Text = "Prioritize " + pendingJob.GetName(),
```
bjubes: Can you add a localization entry for this in en_US as per #1513 instead of hardcoding? something like:
`prioritize=Prioritize {0}`
and then call `LocalizationTable.GetLocalization("prioritize", pendingJob.GetName());` instead

koosemose: Yeah, at the time I was just expanding what was already there to work for multiple jobs per tile, but that doesn't look difficult at all to do.

koosemose:
I've got the localization in place, and a conflict that popped up (for some
reason after a rebase) should be sorted out, since as of my present I can't
view PRs on github.

PR #1577, opened 2016-10-24
C# modding

dusho:
Introduces possibility to use C# to write mods. 
- There is new CSharp folder next to old LUA one that will contain .cs files with methods.
- defining methods and hooks from .xml is still the same
- if there are multiple methods with same name LUA and CSharp, one that is loaded first is used (controlled by ModManager.cs)
- Used currently to visualize status of power cell press.
- requires .net 2.0 instead of .net 2.0 subset for Unity project settings (this may cause failure in Travis until #1574 is resolved and project settings are not changed)

Known stuff:
- currently uses reflections (cached MethodInfo) to invoke methods (which should be slow - but is still 5x faster than current LUA calls)


koosemose:
You say it's slow, is it slower than Lua? Unless it's noticeably faster than Lua, I don't really see the point of this.

dusho:
I want to run some tests and see. 
There are ways to create delegates and cache them instead of using reflection, then it's super fast (similar to method call in C#), but I'm failing on finding delegate that would reflect all our method calls. Somehow object Method (params object[]) doesn't fit all the method signatures. Will keep WIP until I can find a way...

NogginBops:
Why not have the the mods be .dll s? That would make loading faster and the code easier.

dusho:
that would be easily doable.. question is how to integrate it with events from furniture and other systems - you would still need to identify class and method to execute

NogginBops:
True, is there any other moddable games that use pre-compiled mods we could look at? KSP comes to mind but I don't know if the mods are compiled though...

dusho:
ye, I think I saw that KSP and Cities have ability to take external .dll. I do have KSP, but haven't done any modding for it..

dusho:
Also I made some tests regarding performance.
code:
```cs
private CSharpFunctions csharpFunctions;
    private LuaFunctions luaFunctions;

    private string testCSharp = @"
    public static class FurnitureFunctions
    {
        public static string PowerCellPress_StatusInfo(Furniture furniture)
        {     
            float perc = 0f;    

            return string.Format('Status: {0:0}%', perc);
        }
    }
    ";

    // The test code
    private string testLUA = @"
        function PowerGenerator_FuelInfo(furniture)
   
            local perc = 0	

            return 'Status: ' .. string.format('%.1f', perc) .. '%'
        end
    ";

    [SetUp]
    public void Init()
    {
        Debug.IsLogEnabled = true;
        csharpFunctions = new CSharpFunctions();
        luaFunctions = new LuaFunctions();
    }

    [Test]
    public void CompareLUAvsCSharp()
    {
        bool resultCSharp = csharpFunctions.LoadScript(ReplaceQuotes(testCSharp), "FurnitureFunctions");
        bool resultLua = luaFunctions.LoadScript(ReplaceQuotes(testLUA), "PowerGen");

        Assert.IsTrue(resultCSharp);
        Assert.IsTrue(resultLua);

        int iterations = 1000;

        List<string> cache = new List<string>(iterations * 2);

        Stopwatch sw1 = new Stopwatch();
        sw1.Start();
        for (int i = 0; i < iterations; i++)
        {
            cache.Add(csharpFunctions.Call<string>("PowerCellPress_StatusInfo", new Furniture()));
        }
        sw1.Stop();

        Stopwatch sw2 = new Stopwatch();
        sw2.Start();
        for (int i = 0; i < iterations; i++)
        {
            cache.Add(luaFunctions.Call<string>("PowerGenerator_FuelInfo", new Furniture()));
        }
        sw2.Stop();

        Debug.Log(string.Format("Iterations: {0}", cache.Count / 2));
        Debug.Log(string.Format("CSharp calls: {0} ms", sw1.ElapsedMilliseconds));
        Debug.Log(string.Format("LUA calls: {0} ms", sw2.ElapsedMilliseconds));

    }

    string ReplaceQuotes(string text)
    {
        return text.Replace("'", "\"");
    }
```
Results:
image
So even when using C# cached MethodInfo and reflections, method calls are 5x faster than LUA.

NogginBops:
When we have this in I think we should think about increasing the update rate on a lot of things in the game. Great job @dusho !

koosemose:
Closing and reopening to force travis checks to be reran.

dusho:
after trying out some solutions and even trying direct method I was not able to make it much faster than current implementation which is already 5x faster than LUA
fixed stylecop

FraserKillip:
I think we should only support both .cs and .dll files. Using source files with make dev super easy, but distributing dlls would be way easier for end users. Also, we don't have to put effort into compiling them, which would improve load times :)

BraedonWooding:
How about we make another PR for .dll files, and so we have two different systems.  I have some time and my latest PR is just about done so I'll give .dll a go :D

BraedonWooding:
Now based upon my previous comment, I actually think that adding dlls may be somewhat detrimental to us in the future, most likely we will want some kind of 'package' system to pack multiple elements of a mod together (via a zip or whatever) then the game unpacks this 'package' that will consist of a variety (obviously not limited to)  sound, images, other graphics and code.  Now dlls kinda break this idea since they make it hard for users to go and edit their favourite mods if they wish?  It also is nice to have just C# scripts since LUA isn't assembled so to have C# follow that same path is consistent.  Also from my research we can do dlls but it essentially is the exact same process (skipping the compiling that @dusho does), and often still requires some kind of 'invoke'.  So overall I'm very impressed with dusho's implementation :D, I actually think it's a much better alternative to dlls (even if it came about cause dlls were weird or whatever) and the code is built well and nice object structures :D.

NogginBops:
@BraedonWooding The main reason for wanting .dll s is because they are fast to load. C# code is slow to compile. If you have a lot of mods those compile times are going to be really long.

NogginBops:
I also like the idea of a package thing, but why can't the package contain the dll? It makes a lot more sense to me than bundling all of the source with the mod.

dusho:
well, .dll mods should be integrated differently, I guess. As currently we want to hook some events and actions from .xmls, it's nice to specify method name from your C# script.
Dlls should be used for large mods that would override some functionalities, so for that we would introduce some interfaces and some classes modders can override. So core code would have some calls to  methods that will trigger functionality from modded .dlls.
As we have some parts of our code in flux, it would be hard to place those hooks without some .dll mod actually using it and testing it. Same that happened to our quest system.. no one noticed it was broken for weeks.
I think .dll mods can wait..

NogginBops:
What would dll be able to do that C# code wouldn't? I don't really see any difference. All you are doing with the C# code anyways is compiling it to a dll in memory.

dusho:
as I mentioned .dll should be integrated differently, so you can even implement new 'çore' systems, e.g. new components and use them.. with C# currently you can't cause you're limited to simple method calls and event triggers.. Sure, you can probably hack a lot with it.. but it doesn't compare to nice clean new component in modded .dll

NogginBops:
I still don't understand the difference, they are effectively and code-wise the same so why should .dll be special? It would just make development a lot harder if the C# code and the .dll where different.

koosemose:
Conflicted

dusho:
unconflicted

bjubes:
can csharp fuctions written in this new PR use unity and .net using directives directly, or do they still have to rely on modUtils to call things like Debug.ULog()?

dusho:
yes, just tried to put
```cs
Debug.ULogWarningChannel("test", "Status: {0:0}%", perc);
```
into FurnitureFunctions.cs and got spammed with
image

Assets/Scripts/Models/Buildable/Components/BuildableComponent.cs
```diff
@@ -191,7 +191,7 @@ protected bool AreParameterConditionsFulfilled(List<ParameterCondition> conditio
         {
             componentTypes = new Dictionary<string, System.Type>();
 
-            foreach (Assembly assembly in AppDomain.CurrentDomain.GetAssemblies())
+            foreach (Assembly assembly in AppDomain.CurrentDomain.GetAssemblies().Where(asm => !CSharpFunctions.IsDynamic(asm)))
```
bjubes: You don't include `using System.Linq` so this is causing a compiler error.

BraedonWooding:
This should get accepted before my #1569 (DevConsole), so I can add the ability to have C# modding and make sure files are rightly merged.

Assets/Scripts/Models/Buildable/Components/BuildableComponent.cs
dusho: in BuildableComponent, ln.12 I see:
```cs
using System.Linq;
```
also, it was compiled several times. I've removed unused using from the file to trigger new build.

koosemose:
Looks good, stylecop clean, I will be making a small followup PR to move mcs to its own folder, so it can be more clearly labeled (as mcs isn't very indicative of what it is).

PR #1578, opened 2016-10-25
Add clarification in issue template

bjubes:
Adds the following to the top of the issue template:

> _this template is for reporting bugs or issues in the current game. You don't need to use it if you are suggesting a feature or starting a discussion_

Some feature requests and discussions are using the template despite it being intended for bugs. Hopefully this change will clear that up.


koosemose:
Merged

PR #1579, opened 2016-10-25
[Tiny, Fix] Fix error on building tiles on lowest layer

koosemose:
When building a tile on the lowest layer, an error will occur due to a null tile (the down tile from the tile being built) being passed to RegenerateGraphAtTile. This simply returns from RegenerateGraphAtTile if passed a null tile.


FraserKillip:
LGTM

koosemose:
Is Tiny and a fix, so Self-Merged.

PR #1564, opened 2016-10-21
Changes Scenes location

Ohda:
- Both scenes are located in `_Scenes_` folder.
- Main scene file name is `_World.unity`
- Intro scene file name is `MainMenu.unity`

PR #1583, opened 2016-10-27
[Small][Fix] Update EditorBuildSettings with new scenes path

Ohda:
Looks like i forget this when i change the scenes location in #1564.

The paths was still on the old scenes location.


koosemose:
Nice and simple. Merged.

Issue #1576, opened 2016-10-24
Positional Sound


Jaytheway:
### Expected behavior

Sound should play from the position of its emitters.
Unity has this capabilities. Not sure to what extent they're applicable in our situation.

- **Min Distance**	Within the MinDistance, the sound will stay at loudest possible. Outside MinDistance it will begin to attenuate. Increase the MinDistance of a sound to make it ‘louder’ in a 3d world, and decrease it to make it ‘quieter’ in a 3d world.
- **Max Distance**	The distance where the sound stops attenuating at. Beyond this point it will stay at the volume it would be at MaxDistance units from the listener and will not attenuate any more.
- **Rolloff Mode**	How fast the sound fades. The higher the value, the closer the Listener has to be before hearing the sound. (This is determined by a Graph).
- **Logarithmic Rolloff**	The sound is loud when you are close to the audio source, but when you get away from the object it decreases significantly fast.
- **Linear Rolloff**	The further away from the audio source you go, the less you can hear it.
- **Custom Rolloff**	The sound from the audio source behaves accordingly to how you set the graph of roll offs.

As I understand, Custom Rolloff is not an option.

Unity Manual page for the reference: https://docs.unity3d.com/Manual/class-AudioSource.html
Unity Scripting API: https://docs.unity3d.com/ScriptReference/AudioSource.html

### Actual behavior

For now sounds play exactly in the center (in a 2D space (where the camera is)) ignoring position of the emitter relatively to the camera.

**Examples**: 
- If aliens attack from the north, you would here sound directly in your face.
- If the fire started you would here the sound in your face, without any directional indication.
- All furniture placements - in your face.

All these resolves not only in an aesthetic nonsense, but also in a muddy mix. (Especially without any means to mix audio) 

### Steps to reproduce the behavior

- Step 1: Get positions of sound emitter and camera (perhaps, the tile, which is directly below the camera)
- Step 2: Calculate the distance between sound emitter and camera 
- Step 3: ...

Also in Uniti Scritping API there is: "You can play a clip at a static position in 3D space using PlayClipAtPoint." It might be a way to do it(?)

### Related issues

This is one of the basic features that our sound system need listed here: #1535

I am by no means a programmer, so: 
-If you know how to do it
-If you are willing to put time in something that will determine how the game feels
[...]

PR #1584, opened 2016-10-28
New Sound System

### LINKED ISSUE #1533: Sounds don't play when game is build

### LINKED ISSUE #1535: [Ongoing] Sound System. And sound hookups.
Since we are not using a third party sound engine, we need to create some functionality for ours. 
I'm not sure what basic Unity Sound features we can already use, so correct me if there is something from this list.

Also here might be a general list with places that should have a hookups for sounds (alternatively we could use a google spreadsheet for this).
### Basic features
- [ ] **Randomize pitch** in range from -X to +Y (semitones, float) every time the sound is played
- [ ] **Randomize volume** in range from -X to _No change_ (dB, float) every time the sound is played (see `dB to Volume` conversion below)
- [ ] **Random container** that randomly choose one sound from the list to be played when event X happens 
- [ ] **Sequence container** that plays next sound from the list when event X happens
- [x] **Distance** volume **attenuation** and **panning** based on the position of a sound emitter

**dB to Volume conversion**
- 6 dB = Double
- -6 dB == 1/2 Volume
- +18 dB == 8x Volume

``` cs
float dBToVolume(float dB)
{
     return powf(10.0f, 0.05f * dB);
}

float VolumeTodB(float volume)
{
    return 20.0f * log10f(volume);
}
```
### Spreadsheet

Here is a spreadsheet that should help to organize our work on the Sound, if anyone wants to contribute in this department: [Sound Design Spreadsheet](https://docs.google.com/spreadsheets/d/1TXwJnrMUbtvmQkYY0zqw5KlBm-eqd-JOlEiWSUjAbOo/edit?usp=sharing)

koosemose:
This replaces our current usage of the builtin Unity sound system with the FMod Low Level API, along with helper functions to simplify usage of FMod Low Level, and some basic features added during implementation. The primary purpose is to fix an issue where unity doesn't handle loading audio from streamingassets very well in a standalone build.

Closes #1533
Closes #1535 
Partial Implementation of #1576 

Features:
Positional Sound
5 Sound Channels: master, UI, gameSounds, alerts, and music
Sound groups: A sound group is defined by sound files named in the format of `SoundName.#.ogg`, files named that way will be grouped together, and can be returned from a SoundClip object, by default they will be returned randomly, but the SoundClip object can be set to return them sequentially.
Individual sound Cooldown: based on SoundClip object, so all grouped sounds will share the same cooldown.
Variable pitch: a sound can be given a range in semitones, and will play from -semitones to +semitones in pitch.
Variable volume: a sound can be given a range in decibels, and will play from -decibels to no change.

This is stage one, there are several areas with room for improvement (allow setting of minmax for individual sounds/sound groups, variable roll offs, special handling of music (handling changing tracks, looping, and crossfade). However, this PR alone includes some pretty large changes, and I would like to get feedback and get others making use of this system before expanding it further, so unneeded features aren't implemented.

Note: Floor Tile change has two new sounds added, which will play randomly as tiles are placed, these are primarily for purposes of demonstration, and can either be gotten rid of at a later date or repurposes.


BraedonWooding:
I've had a quick play and it seems to be all in order, I will have a look at the code structure and test some edge cases later tonight :D.  Amazing job so far really good implementation.

bjubes:
Do we also need fmod credit in game? It would be any easy addition in another PR, maybe as some credits option from the main menu. Just making sure so if we need one its made before the next version.

koosemose:
Their license page says credit in documentation or in program is fine. So it might be nice to have credits in game somewhere, for now I believe credit in Readme qualifies for their required credit line.

bjubes:
sounds are working in both editor and builds, but personally I don't like how much directionality(?) the sounds have. Furniture sounds like it is being placed to the literal left of the player, which is not the same as a spot on the left side of the screen. It almost sounds like someone else is playing the game in my ear.

koosemose:
I've dampened it somewhat (it takes being further away to be noticeably in a direction than the default settings), I can see about dampening it further, I think it may be possible to set 3Dness on a scale of 0..1

Part of the problem is that due to z-layers there's no "space" for the listener to be that is similar to the apparent camera position so it has to listen from the focus point of the camera (otherwise sounds on layers above you will be closer than ones on the layer you are on). I could expand the distance as far as audio is concerned between layers, but that also means that layers away from your current layer are going to sound that much further away, which I am not certain is optimal.

koosemose:
See if this does the trick for you, it seems to be a little less directly in an ear for headphones but without squashing it too far for frontwards speakers.

BraedonWooding:
We should have multiple options for various sounds like one for earphones, headphones, mono/stereo...

koosemose:
That can come later, at this point I just want to get a properly functional sound system back in place.

Side note however: earphone and headphone? Aren't those the same thing?

bjubes:
Thanks. Ill test it when i get a chance. I think that the audio was acting like a 3d game where it would make for sounds to come from their origin spot in game, so that if the camera is fifty units away it would play "building a wall on the left" as a sound 50m to the left of the players ear.  Since there is no physical place in the game where the "player" is, this just creates a weird audio illusion.

BraedonWooding:
Earphone and headphone should have different sound location settings since earphones are more about being in ear and are more muddy, where as headphones are more surround sound based, and closer to a set of speakers.

bjubes:
IMO 3d sounds should be turned off completely, just try building a bunch of walls in a row from right to left while wearing headphones. Its sounds like someoone in metal boots is walking by my moniter :)

Im merging anyway since this PR isn't about the nitty gritty but 3d sound should be addressed later.

PR #1586, opened 2016-10-28
Visuals component, mine spots as overlays

dusho:
Animator component got renamed and repurposed as Visuals component.
In Visuals component you can set sprite names:
- just from name (useName)
- from function (fromFunction)
- from furniture parameter (fromParameter)

Overlay sprite was added (same way as power indicator icon) to have way define visualization of sprite 'above' normal sprite - currently used for showing ores sprites over astro_wall furniture (instead of old tint color).
image
WorldGenerator class was bit refactored - made into singleton, got rid of some deserialization code and static methods.
Sprites for ores were hacked in fast way, if someone has better ones, feel free to drop them here, I can add them..


BraedonWooding:
The only thing is that ores on corners look odd, maybe reduce to 3/4 of the size that way it doesn't cover the '2.5D' edge?

dusho:
ye, I thought it would look better. I may have to scrap whole overlay idea, unless it will be useful for something else and use here system similar to wall tiles - so sprite sheet with all possible connections.. though it will be kind of pain.. have some for each ore.. hm.. unless I can use this overlay again and use tint to color it...
nice talking to you.. and my self.. :)

BraedonWooding:
Well don't be discouraged cause it still looks really good!  Maybe u make the ore always smaller and centred?  Or make it so its like a bar of iron or rod of uranium (or what the item mined looks like) and make that centred?  Like so many options and I would suggest to try a few.

dusho:
ok. I'm too lazy to do the whole sprite maps with the ores.. how about this one, it's a bit more colorful than first version:
image
detail
image

BraedonWooding:
That looks really good, and it doesn't conflict with corners!  Just update the PR request image with those!  Great job they are quite stunning :D.

dusho:
it does look like candy factory though.. here is new iteration, I like this much more
image

ComputerOverlord:
Nice! Quite different from RimWorld, too. What are all the colors? :P

koosemose:
I'm not satisfied with the actual sprites used, but the sprites themselves aren't really the point of this PR. Otherwise, it looks good, Stylecop clean, tests clean. Merged.

Tranberry:
👍

PR #1594, opened 2016-11-02
[Small] Change furniture placement error to warning

bjubes:
When creating a new world, asteroids that spawn on top of the base raise an exception. This PR changes the call to a warning since its not a problem and fairly common.


koosemose:
Merged

PR #1595, opened 2016-11-02
Converts Settings.cs to use json

zwrawr:
Settings.cs now uses json rather than xml as was decided in #1034


bjubes:
Merging. Note that settings.xml will not be removed by this patch, so remain in your appdata (or equivalent) unused unless it is deleted manually.

PR #1596, opened 2016-11-02
Code Cleanup

koosemose:
Removes unused variables and using statements. Also stylecops code that made it into master unstylecopped.


koosemose:
As this only changes formatting and removes unused variables and usings, I am self merging this.

PR #1598, opened 2016-11-02
Fmod for mac and linux

### LINKED ISSUE #1597: [BUG] FMOD can not load in mac OSX
I'm using max OSX , Unity 5.4
when I play in editor,  console output error:

   `DllNotFoundException: fmod64`

koosemose:
Currently only the fmod dlls for windows are included, this includes libraries for linux and mac.

This is intended to fix #1597 but requires further testing on said systems.


BraedonWooding:
I will have a check tomorrow; to see if it works on mac, since I own a mac.  However, I don't own a linux machine so I can't test that.

bjubes:
Works on my mac. Don't have Linux.

koosemose:
I would consider the mac fix the primary focus so if it works, IMO it should be good for merge. Regarding Linux, I'm pretty confident it doesn't work on master (unless mono/unity does some magic on linux to load windows dlls), so worse case scenario is that Linux continues not to work, and we/I can look into fixing it when someone reports if that is the case.

bjubes:
Agree. Regardless of whether this fixes linux, its better than the current master which doesn't work on either. Merging

PR #1604, opened 2016-11-03
[Fix] Reconnect ContextMenuItem to ContextMenu

koosemose:
ContextMenuItem got disconnected from ContextMenu again


koosemose:
Selfmerging fix to bug in master

Issue #1539, opened 2016-10-15
furniture type naming scheme... again.


koosemose:
According to the current Contributing guidelines, furniture types (and other things of a similar nature) are supposed to be named as type_material, or for something that produces something whatItIs_whatItProduces.

However, that's quickly breaks down if you have multiple types of furniture that do similar things. For example, we have three ways to generate power. Solar Panels, the Power Generator (fission reactor), and the RTG, they all DO the same thing, and Produce the same thing... so by that scheme they should all be Generator_Power... Or perhaps you change it to type, so we have SolarPanel_Power, PowerGenerator_Power (FissionReactor_Power might be slightly better), and RTG_Power. What purpose does having _Power on the end serve, even if you break the technical wording of the guidelines. Or something like the Metal Smelter Smelter_Metal? That just seems like it's just complicating things for the sake of it. Even Wall_Steel serves no real purpose, it's not like we're sorting anything anywhere by their types so that that serves some form of purpose, if we're grouping things we want to do it more explicitly than by what they happen to be named and have some things manage to be in the right spot due to the naming scheme.  How about Airlock Doors, are the Door_Airlock (of course with the naming scheme that implies they're made of airlocks), or AirlockDoor_Steel? If we're naming them this way so they can be sorted, shouldn't airlock doors be at least in the same vicinity as Doors?

As an replacement proposal, I would just say to keep it simple, and have the type as a keyed version of what it is. Steel Wall is SteelWall (or perhaps steel_wall) Airlock Door is AirlockDoor. At the most IF an item could concievably have subtypes then the subtype could be a postfix, so perhaps in the case of wall it would in fact be Wall_Steel, but I don't see a really good reason for that, if we want to group similar objects together in some way we should just have a direct way of doing it, rather than naming things awkwardly.

Though to be honest even though it's unnecessary, I kind of liked the furn_SteelWall style that wall has, and just use the object type as a localization code for furnitures (and other such things), but that could of course be handled by localization code returning it prefixxed with furn_.

dusho:
with the categorization (for building stuff) maybe we can get away from having unique names? It will be a problem to keep unique names if you want to have open modding anyway.
So we can have wall and wall.. but one from folder structures/steel/ , other from structures/wood/ .Then we can make name out of it like structures_wood_wall to have it in game unique.. 
modded things will get some prefixes from author of mod or something

bjubes:
[...]

PR #1607, opened 2016-11-03
Sanitize types in xml

koosemose:
Per discussion in #1539, this attempts to sanitize the types listed in xml. Currently there is no consistent scheme to them. This PR updates Furniture, Utility, and RoomBehaviors to match the simple scheme of the us english item name all lowercase, separated by underscores. 

Additionally a few localization keys that matched the old type were changed to match the new types (changed in en_us as well).

Additionally, as many sprite names are based off the type, sprite files were renamed and xml altered to match the new types. Additionally, as many of them needed changed to match anyways, animation sprite names were changed to a consistent scheme as well, of item_name_state_framenumber, leaving out state or frame number if there aren't separate states or animation frames.

Inventory types weren't touched, as I am unsure at this time if there are any places in code that would need to be changed, and this PR has already extended beyond its original intention due to other elements connected to item type.


Assets/StreamingAssets/Localization/en_US.lang
```diff
@@ -82,14 +80,16 @@ tile_Floor=Floor
 tile_Floor_desc=Meets the highest non-skid standards.
 tile_Ladder=Ladder
 tile_Ladder_desc=High-tech space movement system. Effectiveness reduced when held by weak noodle arms.
+roombehavior_airlock=Airlock
```
bjubes: I think room behavior is part of your other PR, but considering this one is big, it will likely be in master before this gets merged.

BraedonWooding:
Since JSON is being phased in and XML is being phased out, is this necessary?  Or should we just commit these changes to the JSON files as we add them?

koosemose:
I don't know the time at which prototype files will be changed to JSON, this way we can be sure they're all on the same naming scheme now, rather than either doing it piecemeal as they're transferred to Json, or waiting until the Json switch completes to sanitize them.

koosemose:
No errors when merged with master, Merged

PR #1612, opened 2016-11-04
Roombased Pathfinding

koosemose:
Primarily this implements astar pathfinding based on the groomGraph, making use of GoalEvaluators and Heuristics similar to what is used for tile based path finding.

It has similar helper functions living in Pathfinder, and is currently used to find closest inventory and furniture of certain types, as it only needs to pathfind through a smaller number of nodes (rooms instead of tiles) it is drastically faster. Pathfinding for both now find the closest room with the wanted object and then determines which inventory/furniture is closest and returns a tile based path to that. Even though it's technically pathfinding twice, it is still drastically faster to do a dijkstra on the roomGraph, and then use that to find a particular inventory and get a non-dijkstra tile path to the inventory.

Aspects that are missing that I would like to implement before this is ready (If you have any thoughts on how these can be implemented, feel free to share):
- ~[ ] Way to measure distance between rooms (that doesn't require iterating through each tile in a room and doing some measurements to determine a center, as that would be too slow to use in pathfinding)~
- [x] Have Room Pathfinding check smaller rooms first, this would allow it to avoid going outside when there is inventory/furniture in a room that's just as close.

It no longer seems necessary to measure distance now that room size is being utilized properly and the outside is reporting a large enough size.


koosemose:
Now that I'm treating room sizes properly (and returning a sufficiently large number for the outside room), This seems to work quite well without a distance measurement (which most likely would be a more complex and therefore less performant option.

koosemose:
Still runs clean when merged with master, all tests pass. This can likely be used in more places, but will already drastically improve searching for inventory and furniture. Merged.

PR #1614, opened 2016-11-04
[Bug] Fixed StyleCop Errors

BraedonWooding:
### The issue this fixes
Some stylecop issues must have leaked through the last few commits so here is the fixes to those issues.

### What this PR does
Fixes all the spaces after some of the headers (the license headers).  It also renames zLevel to currentLayer, since it's called currentLayer everywhere it is used, and zLevel is not a good naming convention (raised a stylecop error even) and currentLayer is more descriptive.


Assets/Scripts/Models/InputOutput/CameraData.cs
```diff
@@ -15,14 +14,14 @@ public struct Preset
...
-    public int zLevel;
+    public int currentLayer;
 }
 
 public class CameraData
 {
     public Vector3 position;
     public float zoomLevel;
-    public int zLevel;
+    public int currentLayer;
```
bjubes: Does anyone have better potential names for this variable? I think "Layer" is a bit to similar to "overlay", but currentLevel isn't great either?
maybe depthLayer or depthLevel?

koosemose:
currentLayer is incorrect, it isn't the current Layer.

BraedonWooding:
Then what should it be called?  zLevel is a bad name?

bjubes:
#1617  fixes the hungarian notation error so revert the variable name to `zLevel`. All the other changes are valid stylecop violations.

BraedonWooding:
Made the changes requested 😄

PR #1616, opened 2016-11-05
Bored Characters

Geoffrotism:
Made idle characters move randomly 1 tile in any direction ever time the idle state gets called.


Assets/Scripts/State/IdleState.cs
```diff
@@ -26,11 +29,41 @@ public IdleState(Character character, State nextState = null)
...
+                        endTile = World.Current.GetTileAt(startTile.X, startTile.Y + 1, startTile.Z);
+                        break;
+                    case 2:
+                        endTile = World.Current.GetTileAt(startTile.X - 1, startTile.Y, startTile.Z);
+                        break;
+                    case 3:
+                        endTile = World.Current.GetTileAt(startTile.X, startTile.Y - 1, startTile.Z);
+                        break;
+                }
+
```
koosemose: This seems like overcomplication, when you could easily just use Tile.GetNeighbors(). You could even do Tile.GetNeighbors(true), to allow diagonal wandering (which seems weird that they wouldn't when diagonal movement is possible).

Assets/Scripts/State/IdleState.cs
```diff
@@ -26,11 +29,41 @@ public IdleState(Character character, State nextState = null)
...
+            Pathfinder.GoalEvaluator GTE = Pathfinder.GoalTileEvaluator(endTile, true);
+
             timeSpentIdle += deltaTime;
             if (timeSpentIdle >= totalIdleTime)
             {
-                // We are done. Lets look for work.
-                character.SetState(null);
+                // We are done. Execute move then look for work.
+              character.SetState(new MoveState(character, GTE, path));
```
koosemose: First, GTE is a bad variable name, it tells us pretty much nothing. Second it isn't necessary to create a variable here. The following would avoid creating an unnecessary variable, and also avoid the uninformative variable name.

```cs
character.SetState(new MoveState(character, Pathfinder.GoalTileEvaluator(endTile, true), path));
```

Assets/Scripts/State/IdleState.cs
koosemose: This seems unnecessary. Setting endTile to a random value from startTile.GetNeighbours() would be a much simpler way, rather than having a switch statement. You could even use startTile.GetNeighbours(true) to allow idle wandering to diagonal tiles, since we allow diagonal movement.

It would be as simple as:
```cs
Tile[] neighbors = startTile.GetNeighbours();
endTile = neighbors[Random.Range[0, neighbors.Length];
```

It would also allow you to skip the null check as GetNeighbours() will never return a null Tile (unless you pass it the third argument as true).

Assets/Scripts/State/IdleState.cs
koosemose: GTE is a bad variable name, it tells us nothing about what it contains. Also you could easily do `character.SetState(new MoveState(character, Pathfinder.GoalTileEvaluator(endTile, true), path));` which is much more direct and clear as to what is being passed (since it isn't used for anything but passing to SetState), avoids the non-informational variable name, and is also how GoalEvaluators are passed in every other place in code (that I am aware of).

Assets/Scripts/State/IdleState.cs
```diff
@@ -26,11 +29,41 @@ public IdleState(Character character, State nextState = null)
 
         public override void Update(float deltaTime)
         {
+                Tile startTile = character.CurrTile;
```
koosemose: Is there any use of assigning CurrTile to another variable that's never changed?

Assets/Scripts/State/IdleState.cs
Geoffrotism: I believe that Random.Range returns a float. So you would need to truncate that.

BraedonWooding: I personally like the first one, but they both the same really in readability so yeh choose whatever one you want really.

koosemose: Truncate is unnecessary with Random.Range, if both parameters are ints, it should return an int.

Also for your use, when doing code, if you do 3 backticks, a line break, then the code, end with a line break and another 3 backticks and it will properly include the linebreaks in your code.

koosemose:
I am unsure what qualifies this as WIP, I have made a change seeing to my one remaining concern (an overcomplicated method to convert the float returned by Random.value * 4 to an int, when Random.Range(0,4) will generate the same results, without converting an interim conversion to a string.

@bjubes  As it has been untouched in nearly a month, I believe we have to assume it is abandoned (or at least unlikely to be seen to in the near future), and determine if we think it is viable for merging, and if not, what leaves it not viable, and is it fixable.

Geoffrotism:
It was wip because of the reason you stated. There was a change request and I didn't get around to doing it. Otherwise it's in a perfectly good working state as of my last commit. If there aren't any conflicts should be good to go

bjubes:
In that case I'll remove WIP and get around to testing it.

koosemose:
Cool, wasn't sure.

BraedonWooding:
Stylecop

koosemose:
Well, it's certainly amusing to watch, and highlights at least one other issue (doors closing on crew). I wonder what the difficulty would be of implementing something so that they're less likely to wander into more difficult to walk tiles (doors for example), presumably based on the PathfindingCost, and also making the wander go farther afield each idle cycle (walking like 1-4 tiles for example).

Geoffrotism:
I think those are great ideas. I wanted to keep it simple as I thought I had heard something about idle people meeting st a bar or something.

bjubes:
Looks good. Merging

PR #1617, opened 2016-11-05
Standalone Stylecop settings

koosemose:
Currently our stylecop settings are merged with system defaults, however, apparently there are differences between the defaults of Visual Stylecop and Monodevelop Stylecop, one recent example is the differences in the exception list for the Hungarian rule (Monodevelop Stylecop has x, y, and z as exceptions, whereas Visual Stylecop doesn't).

This sets our settings as standalone (not merged with defaults), so differences between the defaults won't cause issues. One notable exception is Monodevelop Stylecop won't raise a violation when there is a space after the license header (as shown in following), whereas VIsual Stylecop will. 
```
#region License
// Project Porcupine Copyright(C) 2016 Team Porcupine
// This program comes with ABSOLUTELY NO WARRANTY; This is free software, 
// and you are welcome to redistribute it under certain conditions; See 
// file LICENSE, which is part of this source code package, for details.

#endregion
```

This doesn't seem to be based off of any setting, just a difference between the two.

### DIFF
--- a/Assets/Settings.StyleCop
+++ b/Assets/Settings.StyleCop
@@ -1,0 +1,5 @@
+<StyleCopSettings Version="105">
+  <GlobalSettings>
+    <StringProperty Name="MergeSettingsFiles">Parent</StringProperty>
+  </GlobalSettings>
+</StyleCopSettings>
--- a/Assets/Settings.StyleCop.meta
+++ b/Assets/Settings.StyleCop.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 287ef7cf5acfeab499ef96e1ea51c971
+timeCreated: 1478323475
+licenseType: Free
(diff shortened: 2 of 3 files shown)


bjubes:
In visual studio im still getting the "blank line after comment" error, but no longer getting Hungarian notation error for `zLevel`.

In MonoDevelop(windows) there are no errors.

koosemose:
Yeah, that is the odd discrepancy I mentioned in my initial comment, despite now identical settings, those are still reported differently, and I can turn on absolutely any stylecop rule (that's vaguely related to spacing) I want in monodevelop, and it will never alert on that line.

bjubes:
Okay well im going to merge this to remove the Hungarian discrepancy. We can figure out the spacing one later and try to prevent people from modifying header spacing in their PRs

PR #1619, opened 2016-11-08
Update MouseOverFurnitureTypeText.cs class summary.

nikip3:
Minor change in the summary of the class.


zwrawr:
Looks good, i should really get a spell check plugin for VS

PR #1620, opened 2016-11-09
[Small] Update Unity 5.4 download link to stable

hcorion:
The download link goes to a unity beta download, which is for 5.5.0. Unity 5.4 is now stable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Please check the [CONTRIBUTING.md](CONTRIBUTING.md) file for contribution instru
 
 For further information, such as Roadmaps, explanations of systems and features, Standards and Conventions, and all your Git needs and troubleshooting see the [Wiki](https://github.com/TeamPorcupine/ProjectPorcupine/wiki)
 
-Make sure that you are using [Unity 5.4] (https://store.unity.com/download?ref=personal).
+Make sure that you are using Unity 5.4.2 [Windows] (https://unity3d.com/get-unity/download?thank-you=update&download_nid=43049&os=Win) | [Mac] (https://unity3d.com/get-unity/download?thank-you=update&download_nid=43049&os=Win)
 
 ## Vote on Proposed Features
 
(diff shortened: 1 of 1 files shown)


hcorion:
Also deleted fmod64.dll.meta from the old location.

koosemose:
Is there a link we can use that would directly lead to the project required version, as the store link seems to only be a slight improvement, since it will change as "current version" changes, and we may not wish to do so, depending on changes made.

hcorion:
Is this something we want though? being that it took this long to change the download link from a generic beta version of unity to a generic stable version. I think it would be more maintainable if we just pointed them to the latest stable version. 
Regarding your question, the closest thing I could find was the Unity download archive, but as far as I can see I can't find a way to point to a specific version without alienating users on other OS (i.e., pointing to a windows .exe) https://unity3d.com/get-unity/download/archive

BraedonWooding:
You could always do something like:Download current unity [5.4]:Windows: www.windowsDownload.whateverMac: www.macDownload.whateverLinux (no linux build yet, except for an experimental).

zwrawr:
I like @BraedonWooding's suggestion. Here it is with the correct links

Download **Unity 5.4.2** [Windows](https://unity3d.com/get-unity/download?thank-you=update&download_nid=43049&os=Win) | [Mac](https://unity3d.com/get-unity/download?thank-you=update&download_nid=43049&os=Win)

BraedonWooding:
Yeh, that looks really good!  A definite improvement on the current one.

koosemose:
Looks good to me

BraedonWooding:
Note: @zwrawr make sure you do another commit to make the changes said in your comment.

hcorion:
Well, if we're going to be providing separate downloads for each OS, why not provide a linux link as well?

hcorion:
Ok, I have changed it based on @zwrawr 's input. You can look at the new version here: https://github.com/hcorion/ProjectPorcupine/blob/change-download-link/README.md
I'm thinking a squash and merge would be the best for this PR.

PR #1580, opened 2016-10-25
Unreachable Job Fix

Geoffrotism:
Fixes #1536, #1545, #1602 
Fixes when characters cant reach jobs and go into infinite loops.
- [x]  Figure out why jobs arent readded to the queue when a path changing event occurs.
- [x]  Fix: Half of jobs being created in unreachable areas are "deleted".
- [x] Address characters going to pick up inventory before realizing there isnt a path to the job site.
- [x] Small bug where characters aren't picking up build jobs that had inventory requirements that were unreachable. 
- [x] Fix: drag and drop jobs getting deleted.
To recreate: add build wall job somewhere (reachable). then add a stack of steel (unreachable). queue job to deconstruct wall (to make steel reachable). Wall should be deconstructed but the current characters will not do the build job (new characters that spawn in will pick it up though).

PR #1622, opened 2016-11-10
[Small] Meta file cleanup

koosemose:
Fixes meta files with merge conflict results (HEAD) that were merged with #1580, along with removing the meta file for the old location of fmod64.dll which was recreated in #1614


koosemose:
Self Merging, as it only changes meta files.

PR #1626, opened 2016-11-14
[Small] Moved DialogBoxes in Resources/UI folder to Resources/UI/DialogBoxes

frankitox16:
Simply moved Dialog Boxes to a proper folder to keep things a bit cleaner.


koosemose:
Nothing *seems* to be broken by moving these. Merged.

PR #1632, opened 2016-11-17
[Feedback Requested] ReferenceManager, helper script added

crazyfox55:
There is lack of consistency with how classes should maintain references between each other (i.e Tile, Character, Furniture, and Room). The goal of these additions is to create a base for managing references.

The idea here is two classes who would like to grab information from each other must implement the IReferenceable interface. Then to make a reference between the classes a call to the static method MakeConnection in the ReferenceManager class should be made. Once the reference is setup the methods below, should be used to grab the information from the reference manager.
Get Methods:
int GetCountOfType()
IEnumerable GetEnumeratorOfType()
K GetFirstOfType()

Each class that implements the IReferenceable interface will have its own hashset of references. If a class instance is going to be destroyed a simple call to BreakAllConnections will clean up the references.

I will be working on implementing these changes in the rest of the code base. If you like where this is heading don't hesitate to ask questions. It would be great if someone else thought this was a beneficial addition.

Thank you,

David McDermott

### DIFF
--- a/Assets/Scripts/Models/ReferenceManager.meta
+++ b/Assets/Scripts/Models/ReferenceManager.meta
@@ -1,0 +1,9 @@
+fileFormatVersion: 2
+guid: 4afe36c34fffa6a468ccd07e2b474112
+folderAsset: yes
+timeCreated: 1479367921
+licenseType: Free
+DefaultImporter:
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/Assets/Scripts/Models/ReferenceManager/IReferenceManager.cs
+++ b/Assets/Scripts/Models/ReferenceManager/IReferenceManager.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+public interface IReferenceManager
+{
+    /*
+    bool MakeConnection<T1, T2>(T1 reference1, T2 reference2)
+        where T1 : IReferenceable
+        where T2 : IReferenceable;
+    bool BreakConnection<T1, T2>(T1 reference1, T2 reference2)
+        where T1 : IReferenceable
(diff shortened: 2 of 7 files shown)


BraedonWooding:
Okay, so as per the Contributing.md file (https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/CONTRIBUTING.md), "Before you start coding, open an issue so that the community can discuss your change to ensure it is in line with the goals of the project and not being worked on by someone else. This allows for discussion and fine tuning of your feature and results in a more succent and focused additions.".  This is under the idea of a significant issue so you really need an issue first.  Also having what is often referred to as a singleton manager, is often bad coding practice since it discourages abstraction, and discourages modular programming (seperate classes which are 'modular' rather than being a holistic program, which is really needed for this project).  Also often singletonManagers are very error prone (instances disappearing), and often classes should not require a reference to another class unless its an 'pawn' class (like actors and actor manager), its about building statics and different ways to alleviate rather than just a single reference manager.  I would suggest opening an issue and raising discussion.  Also stylecop errors exist as well.  Have a read of the contributing.md and the contributor portal.

crazyfox55:
After reading your comment I think your right and my method would only add to the problem. A good example is the List<Character> that every tile has. Looking at the references to that list you'll find it is only used to GetContextualActionProviderOnTile and BuildStuffInTile. It would be quite simple to ask the CharacterManager for a list of characters given a Tile rather than having 100x100 tiles with most lists being empty. Additionally every move of a character requires removing an item from a list and adding it to another for every character. Not to mention this complex set of data is only used when the player makes a Selection or ContextualAction. There should be a decent amount of performance in memory and cpu to be gained from making a change to this. I'll open up an issue discussion thread to continue this conversation. I'll look into stylecop and reread the Contributing.md file. 

Thanks,

PR #1636, opened 2016-11-19
Stylecopping in Travis

koosemose:
Currently Travis only checks that the PR builds and runs Unit Tests. This adds stylecopping to the Travis build process.

Additionally, Stylecop now explicitly ignores any files in StreamingAssets. This doesn't affect Stylecopping in Mono (and presumably Visual Studio) as C# files in StreamingAssets aren't included in UnityMono's assemblies. It just ensures that Travis isn't picking up Stylecop violations that aren't accessible to MonoDevelop normally.


BraedonWooding:
I say, push/merge it.  Then we can test it, and if anyone has any problems or it breaks then we can reverse the change.

bjubes:
I added in a inline bracket and [this is the result](https://travis-ci.org/bjubes/ProjectPorcupine/builds/177371945) Looks good to me, the error is caught and is shown clearly at the end. Maybe we should throw a different error code at the end just to make it easy to see at a glance what caused Travis to fail, compilation or style?

hcorion:
Ok, some questions: 
Why mono version 3.12.1 ? The mono version on my Mac is 4.6.1. Why not install something like homebrew and get the latest version of mono?
I was looking around and StyleCopCmd hasn't been updated in 2 years, however it links to https://github.com/DotNetAnalyzers/StyleCopAnalyzers which has reached version 1.0.0

koosemose:
I'm not sure the version of mono really matters as long as it can run StylecopCmd, and I use that because it is the version that StylecopCmd's sister repository (that has scripts to use for Travis) uses.

And from what I can tell from looking over the page, StyleCopAnalyzers, is an entirely different version of Stylecop, not just a command line app, and as far as I can tell is not available for Mono. So there is a possibility it doesn't enforce the rules exactly the same. It doesn't even use Stylecop.Settings and instead use stylecop.json. So we'd end up with a separate (potentially different) stylecop for Travis than for Mono (and presumably Visual Studio, assuming people would continue to use the version recommended). And the different settings file would mean we would have to maintain two separate rule sets and try to keep them identical.

@bjubes I know next to nothing about Travis, would throwing a different Error code show up anywhere other than 1 change anything other than where it says "Exited with Error code X"?

I'm also not entirely sure exactly how to work the logic of throwing different error codes with how I've done things (I could obviously do things in a different way, but at a guess, I'm thinking it would be more complex). I'd rather not do that if it doesn't do more than just changing the "Exited with Error code X".

Honestly what I'd really like for it to do is on github when you click the Details for a failed build for it to say if it was Stylecop or Tests (Or even compilation), but I have no idea how/if that could be done.

If you're more experienced than me when it comes to Travis (or shellscripting for that matter, when I linux I almost always use perl for things that would normally be shellscripted), feel free to make modifications, either in this PR or in a followup.

bjubes:
[Is this what your looking for?](https://travis-ci.org/travis-ci-examples/php/builds/63469639) It doesn't show directly on github ([its impossible](https://github.com/travis-ci/travis-ci/issues/1809)) but it does have a line for each test, so one for tests, one for style.

bjubes:
Okay so I figured out how to use Travis Matrix so that the two tests are separate and are even executed in parallel as to make them finish faster.
Here is a link to the[ working branch's Travis report](https://travis-ci.org/bjubes/ProjectPorcupine/builds/177548068) (I send you a PR of this one)
[Here is that same branch but with an added style error](https://travis-ci.org/bjubes/ProjectPorcupine/builds/177547646)
[Here is that same branch with a failing test](https://travis-ci.org/bjubes/ProjectPorcupine/builds/177547877)

hcorion:
Alright, would you be opposed to using something like homebrew to install mono? Maybe one of these days I'll make a PR for it.

koosemose:
@bjubes, Looks good.

@hcorion I am neither for it nor against it. I am so far out of my element dealing with this, never done anything with Travis before, and haven't touched a mac in a very long time. Feel free to make a PR, but you'll most likely need to explain how/why it's an improvement.

bjubes:
@hcorion I'm going to merge this in as is, if you want to use homebrew instead to install mono then you can make a PR if you see it as a benefit. Since this works fine, Id rather get it up and running as soon as possible

PR #1637, opened 2016-11-19
JSON support in components

dusho:
This adds support for JSON de/serialize for all current components. XML de/serialization is still there as well.
This can be used as example how to port XML to JSON and still have both serializations running at the same time.
Still not sure I like how JSON data look here (too much stretched vertically).
JSON example:
```json
{
  'Component':
    {
      'Type': 'Workshop',      
      'ProductionChain': [
        {
          'Name': 'Iron smelting',
          'ProcessingTime': 4.0,
          'Input': [
            {
              'ObjectType': 'Raw Iron',
              'Amount': 3,
              'SlotPosX': 0,
              'SlotPosY': 0,
              'HasHopper': false
            }
          ],
          'Output': [
            {
              'ObjectType': 'Steel Plate',
              'Amount': 3,
              'SlotPosX': 2,
              'SlotPosY': 0,
              'HasHopper': false
            }
          ]
        }
      ]
    }
}
```
vs. old XML look:
```xml
<Component type='Workshop' >
    <ProductionChain name = 'Power Cell Pressing' processingTime = '5' >
        <Input>
            <Item objectType = 'Uranium' amount = '3' slotPosX = '0' slotPosY = '0' />
        </Input>
        <Output>
            <Item objectType = 'Power Cell' amount = '1' slotPosX = '2' slotPosY = '1' />
        </Output>
    </ProductionChain>
</Component> 
```


zwrawr:
The Json doesn't contain a Item tag, would this Json work with more than one input/output item?

koosemose:
Note the square brackets after Production Chain, Input, and Output, that means each is an array, and should therefore support multiples.

bjubes:
usually when implementing a feature, there should be at least one specific use so it doesn't get forgotten and its easy to see if its broken in the future.
Obviously this will be used later, but In this case at least one component should use a json file in this PR to prove it works and give it a "reason" to exist in the project.

dusho:
not sure it's possible now.. as components are part of furniture definitions (which are still .xml)
Though there are unittests that prove it works and I'm sure those UTs would fail if something will be broken in here.
And migrating remaining .xml to .json will whole other beast PR..

bjubes:
Okay, since the unit tests are in place I think this is fine as is. I kinda have to take your word for it that it works with json since there is no way to test it in game other than unit tests, but its necessary for transitioning to json from xml, so merging.

PR #1643, opened 2016-11-22
Rename fmod dlls

koosemose:
When doing a standalone build 64 and 32-bit fmod dlls collide, so I've renamed them and fmod dll handling has been altered to compensate. 

This may break building on mac, need someone to test..


Randshot:
So, I just tested it on OS X and it runs just fine for me. There was an initial error stating that there was no config.xml but I don't think that this is related with your changes.

koosemose:
Alright since this apparently doesn't break osx, I'm going to ahead and merge this, so we can get proper builds again.

PR #1647, opened 2016-11-24
[WIP] Scaring Away The Dragons (Improve LocalizationDownloader.cs)

### LINKED ISSUE #1627: Improve LocalizationDownloader.cs
As some may know LocalizationDownloader.cs is a mess. (located `/Assets/Scripts/Localization/LocalizationDownloader.cs`)
This is an issue to discuss potential solutions.

Here are some solutions I can think of:
- Instead of downloading a .zip and moving it around, we just download config.xml, parse the languages in that, and then download the files individually.
- Move to a proper translation engine like weblate.
- Just download config.xml and en_US, then when the user changes their language, download that specific language.

I'm thinking number 3 is the cleanest solution. That way we won't need to package all of the locales, we can just have one. By downloading config.xml we can have the list of available languages, and by downloading en_US we can have a default language to work with. This will be slower the more translations we have to update, but I think it's a good tradeoff, being that (almost?) nobody is going to need ALL of the translations.

Pings: @QuiZr

hcorion:
~~**NOTE: This is extremely unstable and will most likely not work until I finish. At least it doesn't throw any errors.**~~
**NOTE: This should now work but needs testing. If you are rather fond of your current Localization files I wouldn't recommend trying this yet.**
This pretty much completely dumps (as in re-writes) everything LocalizationDowloader did/does. It reuses very little code. 
After this PR I'd like to move the config.xml for Localization to JSON or perhaps SKON.

Fixes #1627

### What this PR does (or is supposed to)

How the code works now (or is supposed to):

- Checks if config file exists
- If it doesn't, download it and every language file that is included in the config file.
- If it does, check if there are any changes since the last date, if there were, download only those files. (possibly even patching, further reducing download, we'll have to see)
- That's pretty much it.

### TODO

- [x] Actually implement checking what files need to be updated.
- [x] Implement patching the files using google's C# patch tool.
- [x] Implement a funny text pop up thing with a reference to scaring away dragons.
- [x] Clean up LocalizationDownloader.cs
- [x] Clean up and remove unnecessary functions in DiffMatchPatch.cs
- [x] StyleCop LocalizationDownloader.cs
- [x] StyleCop DiffMatchPatch.cs


Assets/Scripts/Localization/DiffMatchPatch.cs
```diff
@@ -0,0 +1,2308 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Diff Match and Patch
+ * http://code.google.com/p/google-diff-match-patch/
+ */
+
+ //Heavily modified and stripped by Zion Nimchuk (hcorion) for use in Project Porcupine
```
hcorion: Just so that everybody knows. I plan to strip away a lot of the functions and don't want to confuse people thinking that it is very close to the original file.

hcorion: Hmm, actually, according to [tldrlegal.com]( https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)) it says that to use it I must state State significant changes I made to the software.

hcorion: Yeah, okay. The official license says that

> You must cause any modified files to carry prominent notices stating that You changed the files;

Well look at that! I followed the license without actually knowing it!

BraedonWooding:
Note: you have like a hundred stylecop errors (sometimes it's easier to handle them if you run it as you go rather then just at the end).

Assets/Scripts/Localization/DiffMatchPatch.cs
bjubes: you also need to include the license in full according to that link. I guess for now make its own folder in /Assets/Scripts/Utilities/"somename"/ and place both the script and the full license text in there

hcorion:
@BraedonWooding Yes, that is Google's DiffMatchPath.cs, which was pretty much just a copy paste. I'll be removing a lot of the excess functions and I'll clean it up until StyleCop is happy.

dusho:
I haven't checked how this works, but currently having still issues with implementation in master - guess it sometimes returns null and then remaining of game code doesn't handle it well and messed up the game. Can you make sure that if some files are missing (en_US should be always there, right) that it will fallback nicely and doesn't mess up everything else.
Also will this downloading happen on every startup or only when in language menu?

hcorion:
@dusho What is returning null? I was actually pre-mature in saying this needs testing. I currently have it hard-coded to use an older version of localization, so it will screw up.

hcorion:
Alright, ready for testing and review, @dusho try again, it should work now.

Assets/Scripts/Controllers/GameController.cs
```diff
@@ -14,7 +14,7 @@ public class GameController : MonoBehaviour
 {
     // TODO: Should this be also saved with the world data?
     // If so - beginner task!
-    public static readonly string GameVersion = "Someone_will_come_up_with_a_proper_naming_scheme_later";
+    public const string GameVersion = "Someone_will_come_up_with_a_proper_naming_scheme_later";
```
BraedonWooding: You changed a static readonly to a constant; there are a few points for either case overall since it's in the 'main' assembly it's fine as a constant.  However, complications may occur if its in another different assembly as the main game (or any script); so basically if this (the whole localization set of scripts) has an possibility of becoming a separate assembly (like as a DLL) in the future it should stay as a static readonly else constant is better (slightly).

koosemose:
For future reference it's not strictly required to stylecop files from an outside source, we typically use `// <auto-generated/>` to have Stylecop ignore outside files.

Assets/Scripts/Localization/LocalizationDownloader.cs
```diff
@@ -6,403 +6,366 @@
...
-                    // Either the response was damaged or GitHub API returned an error.
-                    throw new Exception("Error at parsing JSON - hash closing tag not found.");
+                        case "added":
+                            Debug.ULogChannel("LocalizationDownloader", "Adding the file " + file.Filename + " to the localization.");
+                            
+                            DownloadLocalization(new string[] { file.Filename }, hash);
+                            break;
+                        
+                        case "renamed":
+                            Debug.ULogErrorChannel("LocalizationDownloader", "Error, we tried to rename the file " + file.Filename + " but that isn't supported yet!");
```
bjubes: Have this option default to completely reinstalling a fresh copy of localization until this is properly implemented. otherwise this could cause interesting bugs.

hcorion: Yeah, planning to implement this, but AFAIK there isn't PRs to the ProjectPorcupineLocalization that rename files, I'll implement this when I have some free time.

hcorion:
@koosemose now you tell me! 😝

koosemose:
Hah, yep, I didn't realize you were doing DiffPatch for PP, else I would've told you sooner. It can also be done with some ignores in the stylecop settings. And now that I think of it, I will probably see about having a certain directory set aside for outside code like this, that will be wholesale ignored, but that's at some point in the future.

bjubes:
You have some merge conflicts

hcorion:
@bjubes comments have been addressed and fixed.

Assets/Scripts/Localization/LocalizationDownloader.cs
hcorion: Alright, I implemented renamed in the latest commits.

bjubes:
For refrence, Im testing your PR using a branch I made of the localization repo. [here is a diff of it and the real one](https://github.com/QuiZr/ProjectPorcupineLocalization/compare/Someone_will_come_up_with_a_proper_naming_scheme_later...bjubes:test?expand=1).

The most obvious issue is that all the files are empty.

hcorion:
Thanks for pointing that out @bjubes it was just a silly mistake that I just fixed. Should work now. Thanks for putting together such a great test.

bjubes:
Okay so the files download now, and changing their contents works fine. 
Deleting a file on the repo leaves the original file with `404: Not Found` as the text in it. 
Renaming a file is completely ignored.
Creating a new file is also ignored.

Because of this problems I'm changing this to WIP. Change it back when they are addressed.

hcorion:
@bjubes Because you forgot to add the changed files to config.xml. It reads the localization from config.xml

bjubes:
derp.. Ill add the changes to the config and retest it

bjubes:
I incidentally found a small edge case. when config.xml is non-valid, the system give exceptions (as expected) but then does nothing. instead it should automatically download the latest config.xml

hcorion:
@bjubes Hmm, I'm having a hard time trying to solve this. I think my solution already works.
Here is what (I think) is happening and why I think it works:
1. It start parsing the xml on line [190](https://github.com/hcorion/ProjectPorcupine/blob/3a336aa83522f52fa3c4a0aad7ffd8387fdd8d98/Assets/Scripts/Localization/LocalizationDownloader.cs#L190), but it does it line by line, so it successfully gets the lastDate
2.  It still errors because it continues parsing the document and hits the malformed xml, and throws an error.
3. However, it continues on trying to update, if there was an update to the repo to fix the malformed xml, it would grab it, patch it (and if the patch failed) it would just download the file.
4. config.xml is now fixed, if it wasn't fixed, then the error will be thrown next time.

Here is some solutions: 
1. Leave it, it's fairly informative to the devs.
2. Add a `break;` after it finds lastDate, that way it will only throw an error if there is malformed xml before lastDate. This way the error is no longer thrown, because it doesn't parse past lastDate;
3. I attempted to just force it to download config.xml if it hits the xmlexception but Unity didn't think of allowing the ability to call `yield return somefunc();` in a catch, so I can only revert to ugly hacks like setting lastDate to null to force it to download all of the files or using a temporary bool.

bjubes:
After retesting on my test branch adding new files and updating existing ones works fine. Deleting an entire file and renaming still does not. Also the system always says it cannot get the date of the last update and never writes the date in config.xml (this is probably related to the lack of delete/rename functionality)

hcorion:
Ok, I'll work on this when I have some more free time after the Christmas season.

koosemose:
Marking as Abandoned due to inactivity for over a month, feel free to reopen if you return with time to work on it.

PR #1651, opened 2016-11-26
Optimize Travis Build Matrix

### LINKED ISSUE #1650: Travis inefficient setup.
### Expected behavior
Travis should only download mono to check for style errors, and only download unity to compile and run unit tests
### Actual behavior

Both style checking and unit tests download both, which makes both Travis builds unnecessarily time consuming.

This is an easy fix, so I'm self assigning and will push a PR soon.

bjubes:
### The issue this fixes
Fixes #1650 
Travis was downloading mono, stylecop, and unity for every build, when only a few are needed for each

### What this PR does

moves installation to the travis.sh script so that only mono and stylecop is installed for a stylecheck job and only unity is installed for a unit test job.
Doing so shaved about 30s off of the unit test job and almost 2 minutes off of the stylecop job.

Also moved installation scripts into `Scripts/Install/` for clarity


hcorion:
Here is what I would like to request, installing Mono and Unity via homebrew.
You can install it via this command: 
```/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"```

and install Unity via `brew cask install unity`
and can install mono via `brew install mono` or `brew cask install mono-mdk` for a more compatible setup.

I would also like to request the version stored in a variable for easier updating and cleaner code.

bjubes:
@hcorion does homebrew have a way to specify a version of unity to install? when the next version of unity comes out, we may not want to upgrade to it immediately.

hcorion:
Homebrew installs the latest stable version of Unity. IMO if we have a build script it should be building against the latest stable version of unity. The reason I wanted to install mono is because that way it would also be against the latest version of mono. 
The link you provided with a mono version hasn't been updated in two years.

hcorion:
Actually, just keeping the Project version in a separate variable would be awesome for now. I'll create a PR about using Homebrew since it is outside the scope of this PR.

bjubes:
Testing this Pr with broken tests/styles:
Style violation: https://travis-ci.org/bjubes/ProjectPorcupine/builds/179638067
Test violation: https://travis-ci.org/bjubes/ProjectPorcupine/builds/179638252

koosemose:
Could you add travis folding for mono installation as well?

bjubes:
Yep, when I get a chance

bjubes:
Travis is backed up again so I can't be certain the folds are 100% correct. I changed wget to curl for consistency between mono and unity installs and also so its easy to see the download speeds. Also curl has a much nicer "UI" IMO and wget logs a lot of unnecessary stuff like HTTP response codes and DNS lookups

koosemose:
Everything looks good. Merged.

PR #1652, opened 2016-11-30
Update PP for Unity 5.5.0

### LINKED ISSUE #1655: Project Porcupine doesn't compile on Unity 5.5
Hi all, 
in Assets/Scripts/UI/InGameUI/PerformanceHUD/MemoryPerformanceComponent.cs it doesn't compile on Unity 5.5.0. Unity changed the namespace of the Profiler class as mentioned on the release notes: 
> Profiler: Profiler class moved from UnityEngine namespace to UnityEngine.Profiling

The obvious solution would be to just add UnityEngine.Profiling, however this doesn't exist in versions before Unity 5.5.0, so it throws an error. 

Here are the solutions I can think of:

- We just stick with 5.4, and everybody who upgraded can either install 5.4 or just edit the file locally.
- We move to 5.5 and force everybody to upgrade.
- We add an if preprocessor check like 
```
#if UNITY_5_5_OR_NEWER
      using UnityEngine.Profiling;
#endif
```

hcorion:
You can see Unity's blog post here: https://blogs.unity3d.com/2016/11/29/unity-5-5-is-ready-for-you/
and you can view the full changelog here: https://unity3d.com/unity/whats-new/unity-5.5.0

fixes #1655


databyss:
You may want to update the install script as well: https://github.com/TeamPorcupine/ProjectPorcupine/blob/master/Scripts/install.sh#L5

hcorion:
@databyss Sure.

hcorion:
Actually, since @bjubes is currently changing some things in #1651 , I'll just ask him to do it in his PR once this PR is merged.

bjubes:
A few things that have to be addressed before we upgrade the project:

- foremost: the download is the usual 3+GB. We have to give everyone time to upgrade since not everyone can download 3GB on the spot.

- _5.5 depreciates OSX 10.8 Mountain Lion_. With macOS upgrades being free this shouldn't be an issue, but I think we should give anyone who needs it a little notice before we upgrade.

- _Object.Instantiate(Object, Transform) will now use the Object position as the local position instead of world position_. I don't think we have to worry about this but just in case im mentioning it.

- _Upgraded C# compiler to Mono 4.4_. Has the potential to cause a few hiccups with stylecop or other plugins like Fmod or Json. Needs to be tested.

Also for anyone fed up with MonoDevelop:
- _VSCode Support_
- _faster VisualStudio startup time,_ woo!

hcorion:
@bjubes True however, 
I see the Unity version featured in the README as more of a guideline for people new to PP. Having multiple people on different versions of Unity shouldn't really be a problem. 

Also, it doesn't quite compile on 5.5, shall I add the fixes to this PR?

hcorion:
Ok, fixes added to PR.

hcorion:
Ok, the changes in the PR now fixes #1652

README.md
```diff
@@ -1,5 +1,5 @@
 # ProjectPorcupine [![Build Status](https://travis-ci.org/TeamPorcupine/ProjectPorcupine.svg?branch=master)](https://travis-ci.org/TeamPorcupine/ProjectPorcupine)
-# Project Porcupine will migrate to Unity 5.5.0 on December 31st 2016, be sure to update your Unity installation before then.
+# Project Porcupine will migrate to Unity 5.5.0 on January 1st 2016, be sure to update your Unity installation before then.
```
JupiterJoe: should probably say January 1st 2017

hcorion: Ah, yes, good catch, thanks.

bjubes:
The changes in both .asset files are unity 5 specific, so merging this will cause these files to revert back for everyone on 5.4, creating a bunch of unwanted files in PRs.

hcorion:
Ah, right will do!

hcorion:
Ok @bjubes I have fixed the requested changes.

Issue #471, opened 2016-08-21
On high res display UI gets very small


bjubes:
I use an old analog monitor as a second display, so when I noticed how big the UI got I tried it the other way around. On a 5K iMac, the UI gets very, very tiny.

<img width="2672" alt="screen shot 2016-08-20 at 9 57 09 pm" src="https://cloud.githubusercontent.com/assets/7608114/17834855/714ffd30-6721-11e6-9563-d5ec4a17137f.png">

hopefully this can be easily fixed when the UI goes code-based.

dustinwloring1988:
I think this is because the canvas is set to scale - constant pixel. Quill18 talked about this in one of his videos. Easy fix though

bjubes:
thats good to hear. I was afraid this would require a hacky workaround

Fogsight:
Setting UI Scale Mode to Constant Physical Size will fix it, as well as fix any text blurriness, as a side effect. If we want to keep Constant Pixel Size setting, UI would have to be resized manually and fonts would have to be bumped up to double the size to fix the blur, (while scaling down text objects).

So I would vote for Constant Physical Size change as a simplest solution to both.

ComputerOverlord:
It was made CPixS to prevnt text bluriness. If we did that we'd get bluriness.

Fogsight:
It is not 100% sharp, but it is not blurry. It is a good trade off, because when you are using resized text to fit it on buttons, it becomes pretty blurry indeed, with Constant Pixel Size.

I think CPS should only be used when catering to specific device with set resolution. There's no other fix for variable resolutions. And it can be made crisp later by using custom 3D font.

mikejbrown:
Is this still a bug?

bjubes:
i think so, unless a PR for CPS snuck by

vogonistic:
Yeah, it's still a problem. All UI is super tiny on my retina display.

bjubes:
so can we just change it to CPS or will that screw up the entire UI in the process?

WizzardMaker:
Its a bit old, but I have a Script that modifies the cameras view rectangle to keep a specific aspect ratio (like 16:9 4:3) which makes the camera always see the same on different resolutions. It scales the canvas to the right size. (the gif is from my game, so dont worry about the game in the background)
![screen](https://cloud.githubusercontent.com/assets/7768485/19093384/34ebd50c-8a8b-11e6-9912-4556157e77d7.gif)

[...]

PR #1656, opened 2016-12-03
Settings Menu overhaul

### LINKED ISSUE #1615: Settings UI Redesign
WIP (started as of 25/11/2016):
Since this has two sections (an art/cosmetic side) and a practical side (which has been already outlined).  I'm going to build something that focuses on the practical with a **Proposed** art theme.  Then people can ask for adjustments, or overhauls to the theme.  Should follow a similar coding practice to the DevConsole (inheritance is key, but flattened significantly).

We need a settings UI overhaul, based upon previous issues and discussion (I've included all the ones I could find the rest are hidden or are discord chats), this is what I think would benefit us the most.

image
Excluding the colour scheme and look (just more focusing on the positioning of the main buttons), I think this would benefit us the most, the left row of buttons that we would have would be something like: Graphics, Sound, Controls, Gameplay, Developer, --Mods--, Hotkeys.  --Mods-- would just be all the various mod settings that the mod author can create (a standard in most games that we should 100% include) and each mod would get their own button most likely (we could either have a scrollview or just squish the buttons, or just if it gets too big have a 'more' button to open a context like menu).  Sound settings wise we agreed on Master, Music, Game sounds, Alerts/Events, UI.   Graphics would just contain our current graphics settings (resolution, graphics options, possible shadows/other rendering things in future, fullscreen toggle - could be a dropdown of fullscreen, windowed, borderless windowed, UI Skin, V-Sync), Controls is different the hotkeys since I would presume that controls is more like mouse speed, and UI sensitivity?  Hotkeys on the other hand would just be all the keybindings and would need to feature at least one alternate and main key (with two alternates and one main being a preferred choice).  Gameplay would be stuff like, Language, Autosave stuff...  Developer would be the developer settings that my dev console PR adds (#1569) - shown below.  As well as other developer features that may be introduced in the future.
image

### Relevant Issues/PRs
#1562, #1569 #1581

BraedonWooding:
Closes #1615

# What does this do?
Just implements a more clean and a more procedural settings / options menu.  This doesn't touch AT ALL the settings.json file (except to change naming so it's a little more descriptive).  It can be perceived as two individual parts, one the Settings Template this is a new xml file that all it does is dictate the 'template' or the elements of the settings menu (I'll put a little more detail of this below).

image

# What DOESN'T this do
This isn't to set what options should be available (however, I will remove/add any that people want), later on another pull request will handle things like key bindings, and some more graphic options, this was purely to port over our current settings menu to a nicer thing.

## A little more info on the Template
Essentially split up like this:
image

Each category is a button on the left (like a section i.e. Video, General, Sound); Each Heading is simply just a heading (the second partition i.e. Developer Console, Utilities, God Settings); Each Option represents an element (with name representing the display name, key representing its settings key, default value representing a default value, and class name representing the class implementation script).  

The class name is a little more complex than just that; classes are just C# (we only support C#, due to the speed needed and the fact it involves Unity.UI a lot so it's not really gonna be nice with LUA; and its class subclassing so again not really a lua thing).  You can either use the presets (GenericSlider/GenericToggle/GenericInputField) or you can subclass to get some more tweaking ability (you would in most cases subclass for dropdowns [need to, to state options] and sliders - to state ranges).  

Each class has to implement the following: 
GameObject InitializeElement() -> A function that should return the gameobject created, for display.
void SaveElement() -> Should do a Settings.SetSetting().  This is not the place to do a Settings.SAVE!! since this is sometimes called for speed, so you should not save settings and the settings menu will save it when the user wishes.  As well as a getValue() - This isn't required but is suggested (just do a settings getsetting or whatever).

## UI Building
You have the full power of Unity 5.x UI (yahhh....), this means you can create objects (new GameObject()) and add components at your will (including adding your own components that you created within the Settings Menu Functions script coolies right).  HOWEVER, if you're like me and hate having to organise things like layouts (and for example hate having to create like a billion gameobjects and adding a trillion components to get a single slider) then you can use presets (*Intense Clapping*).  What are presets you say??  Well they are just singular and modular elements that you can mix and match to make your own, let's say you want a toggle element with text on the left and a checkbox on the right.  Well you just create a base element (either you can either choose to create your own or choose from 5: fluid and non-fluid (essentially auto-expanding and constricted cell sizes) horizontal and vertical layouts, or a grid based element), then you call CreateText([Text you want], [Boolean indicating to best fit or not]), then just add its parent as the base element then just called CreateToggle() and add its parent as the base element and you're done.  There you have it you have your element in all its glory (with little code required). 

### A quick UI by-warning
Due to the nature of Unity Layouts, sometimes you have to do creative solutions lol.  For example for the toggle I wanted the text to auto-expand but the toggle to stay at an aspect ratio, to accomplish such a task all I did was just create a cell based horizontal element (80x40 : WxH), then make the toggle ignore the layout (through adding a layout element and setting its 'ignoreLayout' variable to true), then just grab its rect transform and do some trickery (not really trickery, just sets its deltaSize to 30, 30, and sets its anchor to 0.5, 0.5 min and max this way it was absolute 30), then set its local position to be 110, 0, 0 (since a cell is 80 across, so I wanted it to just be beyond the element by its width [to vertically align]).

Also for sliders (since the best fit will cause it to jump around erratically), I just got the canvas then just do a scale factor thing to get what the best fit would be and set it at the start (if you changed resolution or you would have to do is re-click the category thing which the system auto does anyway).

# TODO
- [x] Add the final set of buttons (for the bottom row)
- [x] Add ComboBox/Dropdown UI
- [x] Add Inputfield UI
- [x] Make sure colours match
- [x] Get Apply / Save working properly (haven't 're-implemented' it yet
- [x] Implement a generic combobox
- [x] Implement the new UI background
- [x] Implement all previous options (the new options will be a new PR)
- [x] Implement 'applying' scripts (to apply changes).
- [x] Implement a change follower script (to optimise saving)
- [x] Remove old implementation

## After Note:
I had to make a few changes to MainMenu/MainMenuController because Settings (and actually DevConsole) wouldn't work otherwise:
- I had to instantiate a new functions manager (no biggie), and had to make it load in the xmls and scripts for DevConsole and Settings (I did both, only because of weird conflicts that would arise by doing only one for this PR and another in another PR).  Again no biggie.
- I had to assign a variable of DialogBoxManager to MainMenuController (I just removed the one under MainMenu.cs and moved it to the controller variant that way not only does it make more sense it can be access through the 'instance' variable).  Again no issue :D.
These shouldn't effect anything significant.

## What does the changes just implemented do?
You can now 'request' to allocate some space for your settings element.  This is just a simple call whenever you are creating your base object (else you have another call called AllocateSpaceForGameObject that you can do if you are creating your own from complete scratch).  This allows you to have large height elements (which won't be used in this version but rather when I'm implementing all the options).  Everything still fits on screen but there are scrollbars (both vertical and horizontal) if they go over.  I also did a little bit of a UI overwork in respect to colours (making it more similar) and I changed the text on buttons to be a little smaller.  Oh and also it doesn't take up the whole screen now it still takes up a fair amount of room but is considerably smaller and easy to navigate.


bjubes:
The UI building sounds great! Hopefully we can use that same code (or at least something similar in idea) to eventually build the entire UI

BraedonWooding:
@bjubes It actually could be copy + pasted anywhere (like use the same base element and its extensions :D).  Though it should only really be used in areas where it is designated a size (like a cell size or whatever) since that actually fits nicer for Unity :D, I'm already planning to implement it later for the performance HUD (the fps stuff), since that used an earlier version of what I currently just implemented (no modding opportunity).  It would work amazing for like the bottom buttons and for like the 'dialog menus' and so on :D.

Assets/Scripts/Models/InputOutput/ModsManager.cs
```diff
@@ -81,6 +81,9 @@ private void LoadMainSceneFiles()
         LoadFunctions("FurnitureFunctions.cs", "Furniture");
         LoadFunctions("CommandFunctions.cs", "DevConsole");
 
+        // We only allowing settings menu to be C# due to class requirements
```
hcorion: Should be proper english, maybe
 ```// We only allow the settings menu to be written C# due to class requirements.```

Assets/Scripts/UI/DialogBox/Settings/New Settings/BaseSettingsElement.cs
```diff
@@ -0,0 +1,153 @@
...
+using System.Collections;
+
+// Base Class, will be a UI Element
+[MoonSharp.Interpreter.MoonSharpUserData]
+public abstract class BaseSettingsElement
+{
+    public SettingsOption option;
+
+    /// <summary>
+    /// Save to settings your value.
```
hcorion: Should be proper English, perhaps
```/// Save your value to settings.```

File: Assets/Scripts/UI/DialogBox/Settings/New Settings/SettingsCategory.cs [hunk 3]
```diff
@@ -0,0 +1,114 @@
...
+// and you are welcome to redistribute it under certain conditions; See 
+// file LICENSE, which is part of this source code package, for details.
+// ====================================================
+#endregion
+using System.Xml;
+using System.Collections.Generic;
+
+/// <summary>
+/// Holds the category and its options (and name).
+/// No need to write, only will ever read.
```
hcorion: Should be proper English, perhaps:
```/// There is no need to write because you will only need to read the values.```

Comment (BraedonWooding) [inline Assets/Scripts/UI/DialogBox/Settings/New Settings/SettingsCategory.cs]: Yeh, I normally don't focus on writing proper english till later (when I do cleanups), but I'll make the changes regardless :D.

**Comment by koosemose** [comment]
One concern I have, not with the settings menu directly, but rather the code driven UI you're slowly building up in this, is that it may in some cases duplicate the work done in #1470. At some point as you progress through this, you may want to look at this and see if there's any places where things can be merged together, so we don't end up with two different UI systems.

**Comment by BraedonWooding** [comment]
I get what you're saying but tbh it's VERY different.  The particular pull request you got is for popup dialog boxes??  Which is completely different and not transferable (in any real beneficial way) to a settings menu?  I may use aspects of it (like an 'are you sure you want to cancel', or a 'save changes' popup dialog) but in most cases (especially with the overall UI design being focused on layout groups rather than absolute resulting in a much more scalable UI which can look great on any aspect ratio) it will be non applicable?  For example the headings are a unique element that don't exist in the custom dialog boxes?  And that's what they were made for dialog boxes not menus?  And especially things like sliders and stuff that AREN'T prefabs but rather user created elements (though they will mostly consist of prefabs but they aren't a prefab as a whole), it's a lot harder to do things and tbh user side it's easier to use layout groups then to specifically say this is at x, y, in 90% of cases; and the other cases I have already built examples and functions to help alleviate and make it easier.  Overall I understand your concern but it isn't two UI systems its a dialog UI system (for popup boxes) and a menu UI system (which will soon hopefully spread to every UI element that isn't a popup box).

**Comment by koosemose** [comment]
The only difference between a Dialog Box, and a "Menu" is a matter of scale.

**Comment by BraedonWooding** [comment]
I would disagree a dialog box has often a single element or two, whereas a menu has multiple elements and often has multiple different sections (i.e. video, general...), this is harder to achieve in a dialog box; tbh though my biggest issue is the modability being poor (in comparison), i would have to create a new preset for every possible option (i.e. toggle with text, toggle without, two toggles, three toggles, toggle with slider, slider with dropdown, and so on...) it would just be ugly to maintain (which is fine for something like a dialog box where you really only ever have buttons or input fields) but for a settings menu where you have a large array of different elements it becomes annoying.

**Comment by koosemose** [comment]
Maybe it's just me, but the coloration of the sliders seems backwards, feels like blue should be "filled" and therefore on the left side (unless they fill from the right, in which case *that* seems backwards).

**Comment by BraedonWooding** [comment]
Nope they fill from left, I see what your saying I'll flip it :D

File: Assets/Scripts/UI/InGameUI/GameMenu/GameMenuItem.cs [hunk 4]
```diff
@@ -14,7 +14,7 @@ public class GameMenuItem
 
     /// <summary>
     /// Initializes a new instance of the <see cref="MainMenuItem"/> class.
-    /// </summary>
+    /// </summary>s
```
bjubes: accidental change?

dusho:
would use less colors, less fonts and less font sizes.. this looks a bit messy, like it's controls from several different games

BraedonWooding:
Its procedural so it's hard (ni-impossible to control sizes).  Personally I like the amount of fonts and colours they follow a colour theme (with headings being polar to the standard cool blue triad), so they don't really popout.  Again this isn't the implementation of the controls just the general system that will be a separate PR, so things like different sizes will be (or might not due to procedural nature) alleviated in that specific PR.

hcorion:
LGTM 👍

Assets/Scripts/Controllers/MainMenuController.cs
```diff
@@ -62,10 +74,14 @@ public void Start()
 
         // Dev Console
         GameObject devConsole = (GameObject)Instantiate(Resources.Load("UI/Console/DevConsole"));
-        devConsole.name = "DevConsole";
-        devConsole.transform.SetParent(canvas.transform, false);
-        devConsole.transform.SetAsLastSibling();
-        devConsole.SetActive(false);
-        DeveloperConsole.DevConsole.Close();
+
+        if (devConsole != null)
```
hcorion: I'm a little bit confused, why would DevConsole be null? You just Instantiated it.

hcorion: You seem to have a bunch of these checks, just wondering why.

Assets/Scripts/Localization/LocalizationTable.cs
```diff
@@ -114,6 +114,7 @@ public static void SetLocalization(int lang)
         {
             string[] languages = GetLanguages();
             currentLanguage = languages[lang];
+            Debug.LogWarning(languages[lang]);
```
hcorion: Doesn't need to be a warning (just a general) and can probably be the new UnityDebugger rather than just Debug.Log

Assets/Scripts/Controllers/MainMenuController.cs
BraedonWooding: They are called safety or guard checks.  We essentially 100% know it should exist but if it doesn't then who cares no need to do stuff with it.  One way it could be null is if someone wanted to delete the prefab of it, then this would catch it and make sure the game is still playable (for example we may wanted to delete the dev Console for 'commercial releases' though unlikely).  There are a ton of reasons but really in the end it doesn't effect gameplay nor speed and it just adds a little bit of safety (to make us all feel a little better :P)

bjubes:
When hitting cancel, the confirmation to not save settings comes up after the settings dialog is set inactive and then pops back up when you hit "no"
tests

BraedonWooding:
Welcome to the dialog box manager...  This is the ONLY way I can use the dialog box manager (and the prompt info thing), since the layering of Unity UI means that if I put this below the dialog boxes it appears before all other UI elements and looks really bad.  So its either cope with this functionality (it will pop back if you press no), or implement a popup and not use dialog box manager?

koosemose:
Or you could put this in the Dialog Boxes game object (and probably move the Dialog Boxes Game object further down in the list.

Also, I don't think the settings dialog should just fill the entire screen, it's just too much empty space. And even though that would allow it to hold more things, either way there comes a point where there will be too many options for it to display at once and will need scrolling, and we should really try to make sure it looks good for the base game.

BraedonWooding:
I really want to avoid the changing of hierarchy's in the main scene, since that has unexpected issues.  However, I will give it a go but I am majorly cautious of it causing other UI problems.  I am more cautious also changing the size of the settings menu since right now it only supports around 7 elements per column and even shortening that by 10% on each side reduces that to around 5 and as per the options issue there are 7 options already for some categories.  I am planning to add scrolling (which I'll do sometime today) for both the categories buttons and the options in each category however, I really want to avoid it appearing (except for weird resolutions or when mods are used).

BraedonWooding:
You can now 'request' to allocate some space for your settings element.  This is just a simple call whenever you are creating your base object (else you have another call called AllocateSpaceForGameObject that you can do if you are creating your own from complete scratch).  This allows you to have large height elements (which won't be used in this version but rather when I'm implementing all the options).  Everything still fits on screen but there are scrollbars (both vertical and horizontal) if they go over.  I also did a little bit of a UI overwork in respect to colours (making it more similar, @dusho are they are little better now?) and I changed the text on buttons to be a little smaller.  In respect to the DialogBoxes thing I want to open another PR with the only job being to re-order the UI (which won't cause any conflicts since most PRs use instantiated prefabs anyway from scripts so it should be a relatively easy merge over).

BraedonWooding:
Okay so @koosemose I made the changes suggested (I didn't want to put it in the dialog boxes object, because it isn't 'technically' one) but I did move the dialog boxes down a decent amount (and moved other things up) regardless it works nicely now :D.  Also I added scroll bars (just vertical for categories, but vertical and horizontal for options), they will prefer to not be scrollable and often shrink things down a little if they can, (however it is also up to us to make sure they fit nicely).  It right now supports around 6 elements (give or take) vertically and 3 headings horizontally without any scrolling (so 18 options per category before scrolling is needed).  Other than that I just generally made things smoother.  I think it's done for now???  Again I'll 're-iterate' that this WASN'T to implement all the available options rather just the menu itself.  Another PR will handle the options.

koosemose:
While I disagree about it being a dialog box, that's a minor point and ultimately one of terminology. I'm also not absolutely sure I agree with all of the visual design, but that's rather secondary, and past a certain point, all of our UI will at some point need to be redone in such a way that it is all consistent, so that's a very minor point.

One issue I do see, and I'm not sure if this a generalized problem with the controls, a problem in the specific implementation, or a side effect of the specific backing settings not being fully implemented. But the Volumes, that start on 100, say `01` rather than `100`. Since there's no other controls that go to 100 and show the value, I can't be sure if it's specific to them or a general problem, but I have tried saving to the settings with them properly set to 100 and reopening the dialog to see if that had an effect, but it didn't.

BraedonWooding:
I will just include that I'm not artist, I can throw together a semi-decent/decent UI but I'm not really skilled enough to do a full on UI.  So I just kind of built one based off multiple suggestions and our current UI elements (using similar/exact same colour schemes and shapes).  And yes I agree it was never a this will stay here forever it was more of a this looks good enough to pass for a settings menu for a decent period till someone who is more skilled can build a UI (since that requires a very high skill in UX design as well as UI design to build a full interface).

The other issue is purely the fact that this was the implementation of the settings menu not the options.  That bug exists purely due to the fact that the options were literally 'thrown' together as like a demo.  When I properly build them those kind of things should disappear, however since you did note that specific one I just opened it up and fixed it :D.  It was just that me forgetting to set it upon load (and therefore it used the base text formatting of 0 - 1).  Everything that is option based (like text not fitting, or options not saving) is purely the specific backing settings not being fully implemented since the system just holds a bunch of gameobjects and doesn't touch them really and expects them to handle UI and stuff (it does some layout work and so on, and has nice layout trees though).

koosemose:
Yeah, like I said, I wasn't absolutely sure what side of things it was on, so figured I'd let you know so you could find out and let me know. 

Things look good, this being a fairly large PR, I'm going to wait on @bjubes to give his thoughts on this (presumably) final version, prior to merging (or he can just merge if he approves, obviously.)

bjubes:
also keep in mind that the settings window may seem empty on your screen, but on a lower resolution screen it may need to be that big. When I don't maximize on play on a low res monitor, its actually not big enough and text spills over a bit, but thats the consequence of #471, not this PR

BraedonWooding:
I will say though @bjubes that for me when I tried to make it real small (Like REALLY small), it scaled 'pretty nicely' (with most of it being semi readable and the scroll bars help): 
image

koosemose:
Everything seems functional, and looks acceptable. Merged.

PR #1657, opened 2016-12-04
[WIP] Utilities connect to any part of furniture

dusho:
Solves problem where utilities (power cable) had to be placed exactly at origin position of multi-tile furniture - only then furniture was powered.
Now any tile of furniture will do
image
Also removed the hack implementation for more cleaner usage.
Please test scenarios with connecting cables to furniture. Placing cable first, then furniture and other way around.
Issues:
- [ ] problem when reconnecting broken cable (furniture not powered unless more connections to furniture are made)

### DIFF
--- a/Assets/Editor/UnitTests/Models/Power/PowerGridTest.cs
+++ b/Assets/Editor/UnitTests/Models/Power/PowerGridTest.cs
@@ -295,8 +295,6 @@ public class PowerGridTest
 
     private class MockConnection : IPlugable
     {
-        public event Action Reconnecting;
-
         public float AccumulatedAmount { get; set; }
 
         public float AccumulatorCapacity { get; set; }
@@ -329,6 +327,14 @@ public class PowerGridTest
         }
 
         public float OutputRate { get; set; }
+
+        public string Medium
+        {
+            get
+            {
+                throw new NotImplementedException();
+            }
+        }
 
         public void Reconnect()
         {
--- a/Assets/Editor/UnitTests/Models/Power/PowerGridTest.cs
+++ b/Assets/Editor/UnitTests/Models/Power/PowerGridTest.cs
@@ -295,6 +295,8 @@ public class PowerGridTest
 
     private class MockConnection : IPlugable
     {
+        public event Action Reconnecting;
+
         public float AccumulatedAmount { get; set; }
 
         public float AccumulatorCapacity { get; set; }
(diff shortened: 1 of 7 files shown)


koosemose:
You've broken some scenarios of network splitting/joining. One case I was able to get repeatable is the following:
image
To repeat, build this setup, remove cable at marked point, then replace it. The o2 Generator on the left will no longer receive power.

Also, unless it is integral to the rest of this PR functioning, could you remove the hack fix? In the branch I have working on the FluidNetwork, I already have a different fix in place (that takes into account some of the elements I have in place for fluids), and would prefer not to have to sort out the conflict between them.

dusho:
@koosemose I reverted back as much as possible. New functionality is only when reconnecting multi-furniture tiles. There is still that weird problem with reconnecting you mentioned.. I'm not really sure where the problem is, it behaves weirdly.

koosemose:
There's certainly something strange going on, as an additional odd case, presumably related, if you build a third O2 gen (I built it in the upper separate "wing" area), then disconnect it (I did it at a spot one away from where it connects) it will also cause the 2nd built (the one added in the previous case) to disconnect and stay disconnected (I've only been able to get it to reconnect by placing a cable in another of its tile), But it's unreliable (it did it the first time, but I can't recreate it yet)

koosemose:
Oddly enough, looking at my own implementation (regarding the hack fix) and yours, mine is almost identical to yours, and will have to become more so (I originally used an enum instead) since using a string will allow a much cleaner conversion from xml to utility. If you don't mind I may borrow the applicable pieces of your old code, since otherwise I'll just be typing nearly the exact same thing by hand.

dusho:
yes, sure - you mean that usage if GetPlugable(HashSet), right. Method is still left in Furnature.cs to be used (I removed the usage because you mentioned that problem with merge) as it should be generic enough to cover all we need.
I suspect some problem with merging grids.. I though I don't have to dig into that code, but I guess I will have to unless you want to check it - you did the original implementation, right?

koosemose:
It's not the grid merging that's the issue (all the cables still have the same grid (it's the unlabeled number that I forgot to remove from my implementation), somehow the message is never getting to the furniture to reconnect.

dusho:
getting rid of this as #1673 uses pieces of this code now
Will probably look into this issue and other issues related to grids in different PR

PR #1659, opened 2016-12-06
Fix crash on world gen

koosemose:
World gen was crashing due to trying it trying remove the floor under a wall, because IsStartArea had an off by one error in the Y direction. This adjusts the Y bounds. 

Additionally the World Seed in now logged to the console to aid in future debugging of anything that is inconsistent based on world generation.


BraedonWooding:
I've created about 15 worlds and it hasn't yet done the bug, so I would say it's fixed :D.

koosemose:
Goodbye Bug. Merged

Issue #294, opened 2016-08-19
[Bug] Wall edge conditions


kd7uiy:
The wall does weird stuff when you build along the edge of the map.

![capture](https://cloud.githubusercontent.com/assets/5951269/17817291/c07e4a90-660b-11e6-8274-13b381ecc088.PNG)

vogonistic:
Probably shouldn't allow building on the last tiles. That would be a good beginner bug fix. I'll write it up.

koosemose:
For others looking at this, as it's not completely clear what's going on from the picture alone, it appears that walls built along the edge don't show a sprite (they still show up in mouseover info), what can be seen in the picture is the walls connecting to the walls with no sprites.

I suspect it has to do with trying to check outside of the map for neighbors.

vogonistic:
Edges of maps are always problematic, best not to let the users build on the very last tile along the edge.

@kd7uiy Would you rather have a go at fixing it before I put it out as a beginner bug?

kd7uiy:
I'm not planning on fixing it.

The edge tile is a wall, but the sprite doesn't show. I would propose putting it as a sprite.

PR #1661, opened 2016-12-06
Remove MinEdgeDistance

### LINKED ISSUE #1660: Is MinEdgeDistance still needed?
Currently we have MinEdgeDistance of 5, which prevents you from building within 5 of the edge of the map, this was originally instituted to solve a bug related to building right along the border. However, preliminary tests suggest that even with MinEdgeDistance set to 0, no errors are caused by building near the edge of the map (there is an error if you try to build tiles off of the map, but that isn't related to setting MinEdgeDistance to 0 and is present in master).

Note: This change would help to support designing the start area in-game as proposed in #1662.

Is there any reason to lock off access to the edge of the map if there are no longer technical reasons?

I would like assistance in testing if there are any issues with removing MinEdgeDistance, and to support that have created PR 1661, please let me know if you find any issues caused by this, or if you test with no issues.

koosemose:
Closes #1660

This removes MinEdgeDistance, which I don't believe is needed any more.

Verbatim copy of text in #1660 as to why I believe MinEdgeDistance is no longer needed:
Currently we have MinEdgeDistance of 5, which prevents you from building within 5 of the edge of the map, this was originally instituted to solve a bug related to building right along the border. However, preliminary tests suggest that even with MinEdgeDistance set to 0, no errors are caused by building near the edge of the map (there is an error if you try to build tiles off of the map, but that isn't related to setting MinEdgeDistance to 0 and is present in master).

Update: Since this has not seemed to create any unforeseen bugs, I've now fully removed MinEdgeDistance, and consider this viable for merging.

### DIFF
--- a/Assets/Scripts/Models/Area/WorldGenerator.cs
+++ b/Assets/Scripts/Models/Area/WorldGenerator.cs
@@ -57,7 +57,7 @@ public class WorldGenerator
         int offsetX = Random.Range(0, 10000);
         int offsetY = Random.Range(0, 10000);
 
-        int minEdgeDistance = 5;
+        int minEdgeDistance = 0;
         
         int sumOfAllWeightedChances = asteroidInfo.Resources.Select(x => x.WeightedChance).Sum();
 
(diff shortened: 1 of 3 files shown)


bjubes:
Do you know what bug was/the commit that created minEdgeDistance?

hcorion:
was there from the start: https://github.com/koosemose/ProjectPorcupine/commit/b379d0f6e33cb47b5c8a083300421a94d14df557

koosemose:
#294

bjubes:
Well id that's the only reasoning for it then removing it shouldn't cause any problems

bjubes:
looks good. If there are any remaining bugs after this is merged we can readd it then or find a better solution

PR #1663, opened 2016-12-06
[Fix] Prevent Building Tiles off the map

koosemose:
Currently there is a bug where you can attempt to build tiles off of the map, this causes an error due to the null tile (only BuildMode.Furniture and BuildMode.Utility check for null). This fixes that bug by moving the null check outside of the code for BuildMode.Furniture and BuildMode.Utility, instead skipping the entire tile if it is null, regardless of the BuildMode.


BraedonWooding:
Works for me I can't build outside the map

PR #1665, opened 2016-12-07
Add z level to mouse cursor text

bjubes:
### The issue this fixes
There is no real indication of what the z index of the tile under the mouse is. This makes it hard to test other PRs and use dev console commands.

### What this PR does
Adds the Z level of the tile that the mouses is over. (toggled by Ctrl + M as before)
I also re implemented the toggling of this behavior to use the Keyboard Manager as opposed to unity's default input system.


koosemose:
Nice and Simple. Merged.

PR #1669, opened 2016-12-08
Start area from .sav

### LINKED ISSUE #1662: Load Starting Area from a save file
Currently the starting area is generated directly from the data in WorldGenerator.xml, however it is formatted oddly, meaning special handling needs to be done if any new features are implemented that would need to be present on world generation. Additionally it flips the Y coordinate, from what it is in game, which can be somewhat unintuitive.

I propose, rather than having the starting area defined in the WorldGenerator.xml, it instead direct to a file (stored in streaming assets, rather than the standard save location) which uses the same format as a save file (most likely with some information stripped out, such as characters for example), this would allow us to use the same code as is used to load worlds from save (possibly with special handling somewhere to handle it being offset). 

This would accomplish several things, first it would make it unnecessary to implement loading the starting area separately, making it easier to modify the starting area. Second it would allow the start area to be designed in game (though the save file would have to be manually moved to streamingAssets), and make it quite a bit easier for modders to create modded worldGenerators (possibly allowing worldGenerators to be part of scenarios). Note: designing the starting area in game relies on removing MinEdgeDistance, as proposed in #1660.

As a preliminary step, I have already added the ability to set the size of a new world from a dialog (a very badly put together dialog for the time being), and also to turn off asteroid generation, this is to allow simple conversion of the current starting area to a .sav file to be used as the starting area after changing it over. The variable size world will presumably stay in (hopefully with a better UI), The asteroid generation may or may not stay in, possibly moving to a setting in the world generator.

Future steps that I intend (though for a later PR, and almost certainly not the PR that will come from this assuming it is met with general approval) would be the ability to choose from multiple worldGenerator data files. With the data currently in worldGenerator.xml, this would allow worlds with different density of asteroids (or at least generating them on a different scale), more or less resources (possibly leaving some resources out), a different starting area, and different starting money.

koosemose:
Currently the starting area is loaded directly from worldGenerator.xml, in a semicustom format (y is flipped, and tiles are loaded as a number for the tiletype), this makes it a pain to edit the StartArea, and causes a need to code in extra handling for any new features to be available from the start area (as an example, despite Utilities loading from a sav file, they wouldn't be loaded from worldGenerator.xml).

This PR changes to using a .sav file (presently located in StreamingAssets\WorldGen) to load the starting area, so any feature that can be loaded from a save file will automatically be available for the starting area (with the caveat that some things may need manually reset, particularly anything that uses an array based off the size of the world, as an example temperature has to be reset, because it initially generates when loading the start area, sized to the size of the starting area only).

For development purposes, if the game is started from the World Scene, it will default to the old starting world size of 100x100x5.

At present, the .sav file is hardcoded, but will be changed to load from the worldGenerator file, and the worldGenerator.xml will likely be moved to streamingAssets/worldGen, to facilitate having multiple world generators available.

Also included, as it was instrumental in creating the startArea.sav file quickly and easily and also for testing proper generation (and is also a useful feature for the game in general), on selecting New game from either MainMenu or in-game, a dialog will appear allowing you to set the size of the world to generate, and determine if asteroids should be generated (this second option should perhaps be a setting in generator file, but was just a quick implementation for startArea.sav creation that may be useful in the future). World creation will fail if the world size is set smaller than the starting area (it should, at a minimum just not shrink dimensions that are smaller than the starting area, and possibly disallow you from setting a smaller area, or maybe even have a minimum size set apart from the starting area size, but I'm not sure of a good way to implementing the second two on the UI of the New Game settings).

Closes #1662

As an additional note, this disables TestRoomGraphGeneration since it is particularly tuned to the default world and won't work properly with other worlds (and Reading the start area from save file causes things to be loaded in a slightly different order, which causes this to fail, despite the room graph being generated properly). I'm not entirely sure TestRoomGraphGeneration is really needed, or at least not implemented in this fashion. It either needs to be built into a unit test (with a set world to test and retuned to the different loading order of things) or would need some form of error checking data in the world generator file (or perhaps the .sav) to tell it what should be connected where.

Update: 
You can now select from multiple World Generator files (currently there is Default, which is what was used before, and Empty, which has no base) on new game.
image


BraedonWooding:
_World 1 errors out on playmode.  It should probably tell you the minimum width and height, so I think that having a slider is probably better than having an input field (since you can enter negative numbers causing overflows and weird things).  Other than that everything works great.

koosemose:
For the record, to repeat my responses from Discord. World1 is solely a temporary thing for creation of UI which will be removed when no longer WIP.

Slider won't work (in my opinion), due to the need to set a maximum, and finding a maximum that is neither too low nor too high would be extremely difficult (if not impossible). I would prefer to do Input Validation. However, since the actual new world dialog is only a side effect feature of this PR (created to support other work, but left in because it is useful and so as to not have to attempt to duplicate the same work in another PR), I would prefer to do any extensive tuning of the UI in a separate PR, to keep this one focused on loading the start area from the save.

BraedonWooding:
Yeh sorry, probably should have deleted comment, though others may have similar ideas so it's still good I guess.  I would still recommend that when world generation is done it has a second validation or uses unsigned integers for those kind of values (actually saving space too).

BraedonWooding:
*StyleCop* (Just incase if you missed it)

koosemose:
Yeah, it will continue to have stylecop errors until I decide what to do with the roomgraphtest, since as it stands, that's just commented out... which stylecop doesn't typically like.

koosemose:
I *believe* this is now ready for merging, pending testing.

Assets/Scripts/Models/Area/World.cs
```diff
@@ -338,6 +333,74 @@ public void ReadJson(string filename)
...
+        if (width == Width && height == Height && depth == Depth)
+        {
+            // No change, just bail
+            return;
+        }
+
+        int offsetX = (width - Width) / 2;
+        int offsetY = ((height - Height) / 2) + 1;
+
+        Tile[,,] oldTIles = (Tile[,,])tiles.Clone();
```
hcorion: What's with the funky variable name? Your scaring me here with both capital i and l together.

Assets/Scripts/Models/Area/World.cs
```diff
@@ -56,21 +56,12 @@ public World(int width, int height, int depth)
...
-            }
-        }
+        roomGraph = new Path_RoomGraph(this);
 
         // Make one character.
-        CharacterManager.Create(GetTileAt(Width / 2, Height / 2, 0));
+        CharacterManager.Create(GetTileAt((Width / 2) - 1, Height / 2, 0));
 
-        TestRoomGraphGeneration(this);
+        // TestRoomGraphGeneration(this);
```
hcorion: Could just remove this line altogether.

Assets/Scripts/Models/Area/WorldGenerator.cs
```diff
@@ -47,88 +40,66 @@ public static WorldGenerator Instance
...
                             {
-                                int currentweight = 0;
-                                int randomweight = Random.Range(0, sumOfAllWeightedChances);
-
-                                for (int i = 0; i < asteroidInfo.Resources.Count; i++)
+                                if (asteroidInfo.Resources.Count > 0)
                                 {
-                                    Resource inv = asteroidInfo.Resources[i];
+                                    int currentweight = 0;
+                                    int randomweight = Random.Range(0, sumOfAllWeightedChances);
```
hcorion: camelCase if you please (and the line above it).

koosemose: Didn't even notice that, leftovers from the previous version that got shifted around.

koosemose:
That should take care of formatting concerns.

koosemose:
No issues reported (barring an issue with a temporary scene from development that has since been removed.). Merging.

PR #1670, opened 2016-12-08
New Commands for Dev Console

BraedonWooding:
New Commands are as follows:
NewCharacter(Vector3 pos) // New Character at the position provided
NewCharacterCenter() // Creates a character at the center square
Exit() // Closes the dev console
Status() // Prints the current speed (or if it's paused) and if devMode is on / off
DevMode(bool isOn) // Enables/Disables DevMode.

You can ask for more below :D

### DIFF
--- a/Assets/Scripts/UI/InGameUI/DevConsole/DevConsole.cs
+++ b/Assets/Scripts/UI/InGameUI/DevConsole/DevConsole.cs
@@ -815,7 +815,8 @@ namespace DeveloperConsole
                 new Command("GetRoomIDs", CoreCommands.GetAllRoomIDs, "Logs all the room IDs"),
                 new Command<int, string, Vector3>("DoBuild", CoreCommands.DoBuild, CoreCommands.DoBuildHelp),
                 new Command("DirtyTileGraph", CoreCommands.InvalidateTileGraph, "Invalidates the tile graph"),
-                new Command("GetCharNames", CoreCommands.GetCharacterNames, "Logs all the character names"));
+                new Command("GetCharNames", CoreCommands.GetCharacterNames, "Logs all the character names"),
+                new Command<Vector3>("NewCharacter", CoreCommands.NewCharacter, "Creates a new character at position provided"));
 
             // Load Commands from XML (will be changed to JSON AFTER the current upgrade)
             // Covers both CSharp and LUA
--- a/Assets/Scripts/UI/InGameUI/DevConsole/DevConsole.cs
+++ b/Assets/Scripts/UI/InGameUI/DevConsole/DevConsole.cs
@@ -819,7 +819,8 @@ namespace DeveloperConsole
                 new Command<int, string, Vector3>("DoBuild", CoreCommands.DoBuild, CoreCommands.DoBuildHelp),
                 new Command("DirtyTileGraph", CoreCommands.InvalidateTileGraph, "Invalidates the tile graph"),
                 new Command("GetCharNames", CoreCommands.GetCharacterNames, "Logs all the character names"),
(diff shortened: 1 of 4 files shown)


bjubes:
exit - close the console
devmode(bool) - set dev mode without using the settings UI

also, does the help command automatically show all possible commands or do new ones have to be added to it manually?

BraedonWooding:
Sweet yeh I'll add those, and yeh its procedural ( it just grabs the name + "(" + the parameters + ")" )

koosemose:
Purely for laziness sakes, can there be a no parameter version of New Character that just spawns them at the center? On a related note, just out of curiousity, is it possible for a command to have a default? i.e. To do a no parameter version of NewCharacter, does it have to be an entirely separate command or can it have be one command that spawns at coords given, or if no coords given (and thereby a default value that is invalid perhaps) it spawns at center?

BraedonWooding:
I'm going to in another PR implement some things you can enter as parameters that will auto do.  Such as Ptr (for current mouse position) as well as things like a center and things like current time scale, or a togglable option (so if you wanted to toggle DevMode you could just type DevMode(toggle) - though that one will be harder so it may just be something like DevMode(!=DevModeSetting) to reverse the setting idk, though I'm still trying to make it simple and not complex.  So to recap and answer your questions:
Yes, I'll add a command called something like NewCharacterCenter() :D, but it probs will be removed later and rather you will just have to type "Center"/"Centre" as the parameter for new character, and I'm thinking about adding an option for a default (which wouldn't require any effort really but a new xml parameter, since I already have tools to convert values to the parameters it wants from a string I would have to just do that on the xml default string).

Assets/Scripts/UI/InGameUI/DevConsole/DevConsole.cs
```diff
@@ -815,7 +818,9 @@ private void LoadCommands()
                 new Command("GetRoomIDs", CoreCommands.GetAllRoomIDs, "Logs all the room IDs"),
                 new Command<int, string, Vector3>("DoBuild", CoreCommands.DoBuild, CoreCommands.DoBuildHelp),
                 new Command("DirtyTileGraph", CoreCommands.InvalidateTileGraph, "Invalidates the tile graph"),
-                new Command("GetCharNames", CoreCommands.GetCharacterNames, "Logs all the character names"));
+                new Command("GetCharNames", CoreCommands.GetCharacterNames, "Logs all the character names"),
+                new Command<Vector2>("NewCharacter", CoreCommands.NewCharacter, "Creates a new character at position provided"),
```
BraedonWooding: Firstly, yes they should however they are using some older code to get the current tile (which I haven't updated to be 3D), I'm going to update it in my DevConsole2.0 Revisions PR thingy so should we just leave it till then?  The next point is odd?? Neither of them have any code that would do that functionality, the center just calls the new character with a center position?  Also I'm looking to remove center position code to replace it with defaults (which are already in place in PR #1680).  And yeh same with the third.  I'm thinking that due to the small nature of this either I just close this PR and incorporate changes in my bigger revision PR since I'm moving the location of all commands in that too (to make them more available to modders), or I guess we can just merge it since while it may not have optimum it has pretty close (though personally I would rather closing it down and just doing the changes in my revision PR due to reasons stated)?

bjubes: Since these are so small, you can just add these changes in your other PR if its easier. if @koosemose really wants the NewCharacterCenter he can just do it himself :)

BraedonWooding:
Closed since these changes are just going to be incorporated in another PR

PR #1671, opened 2016-12-08
[Small] Claim inventory

koosemose:
Presently characters will all try to go after the same inventory, this PR simply stores a claim in the form of a DateTime object, and will prevent a character from finding an inventory with a claim more recent than 5 seconds, and resets the claim time on pickup so that if they don't pick up the full stack it is now available for another character. DateTime is used rather than a boolean or the character with a claim so that it autoexpires, without needing any event to be subscribed to or an update sent to inventory. Optimally we will, in the future, implement a form of in game time that can be used to track claims (along with many other uses of such a system), however, as is, using DateTime, a claim will expire if the game is paused right after a character claims an inventory, allowing another character to claim the same inventory, and there is little point in saving the claim time, as it will definitely be expired by the time the game is loaded again.

The expire time could perhaps use some tweaking, and could maybe even be made significantly longer, since a character *should* release the claim upon picking up the inventory (more accurately, the act of picking up an inventory releases the claim), with the expiration present only as a failsafe.

### DIFF
--- a/Assets/Scripts/Models/Inventory/Inventory.cs
+++ b/Assets/Scripts/Models/Inventory/Inventory.cs
@@ -17,7 +17,9 @@ using Newtonsoft.Json.Linq;
 [System.Diagnostics.DebuggerDisplay("Inventory {ObjectType} {StackSize}/{MaxStackSize}")]
 public class Inventory : ISelectable, IContextActionProvider
 {
+    
     private int stackSize = 1;
+    private DateTime claim;
 
     public Inventory()
     {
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Models/Inventory/Inventory.cs
```diff
@@ -81,6 +82,30 @@ public Inventory Clone()
...
+
+    public void ReleaseClaim()
+    {
+        claim = new DateTime();
+    }
+
+    public bool CanClaim()
+    {
+        DateTime requestTime = DateTime.Now;
+        return claim == null || (requestTime - claim).TotalSeconds > 5;
```
bjubes: Theres no need for the separate requestTime variable.

bjubes:
Personally I think a claim system using subscriptions like you described will be needed in the end, but this is a good short term workaround.

dusho:
will DateTime.Now deal with different game speeds? shouldn't it use some in game timer?

dusho:
I think you could have introduced that game timer here as well.. and make claim inventory use it right away.  Coupled in that way you can use it for testing timer and showing its usage.

koosemose:
I could have, but I prefer not to do multiple things in a PR, unless one is integral to supporting the primary purpose. 

Plus I haven't put much thought in the specifics of game time, how exactly it should be tracked, (right now the most reasonable thing would be to tie it to the slow update, i.e. 5 ticks per second) nor discussed said things with the community.

Mostly I was just annoyed at my crew all piling up on a single inventory, so I just whipped this up real quick.

koosemose:
Going to go ahead and merge this, since it does work, until it can I or someone else can implement some form of Game Time that it can use.

PR #1672, opened 2016-12-09
Time Scale Indicator

BraedonWooding:
Image below.  Just a real time indicator which will tell you what your current speed is :D.  I'm going to add 3 time control options (Speed up/down, pause).  The + increases, the - decreases and if you click the arrow (which will represent the current speed) it will toggle pause.

The options (though we have mostly reached a consensus):
image

OR

image

OR

image

OR (the chosen one *so far*)
animation 2

# Tasks left
- [x] Make it spawn from prefab
- [x] Cleanup a little


koosemose:
A few issues: First the time scale controls are partially obscured by the headlines (rather this would be considered an issue with this or an issue with the Headlines is questionable). The size of the '+' '-' and the fact that it seems you have to click directly on the graphic can make them a little difficult to click on. There are also times when it appears I'm clicking on them, but they don't respond, and other times when they visually respond, but the time scale doesn't change (the first may be an aspect of the previous, while the second may have to do with fps occasionally tanking.

On the visual aspect, I'm not sure how I feel about the vertically oriented +/-, a more horizontally oriented layout seems more natural to me (with speed up/speed down controls on either side of the current speed indicator and perhaps a slightly smaller current speed indicator)

BraedonWooding:
Okay with the headlines, I placed it at the top since I didn't really like it on the horizontal sides of the screen, and I felt that since in most cases you close the headline pretty quickly (after a quick read) it wouldn't affect anything significantly?  With the size of things I'll increase the button size and decrease the visual indicator size as well as place them horizontally and see how that goes.  I can't replicate your bug of it not responding at all.  It perfectly responds for me?  Also there is no way for it to visually respond and the time scale to not update since it gets the visuals from the current time scale (every frame it checks if its changed and if the time scale has changed it updates the visuals).  It will however unpause the time scale if you press the +/- since that's standard behaviour normally?

I.e. the code that updates the visuals is:
```
if (oldTimeScalePosition != TimeManager.Instance.TimeScalePosition) {
    oldTimeScalePosition != TimeManager.Instance.TimeScalePosition
    
    if (oldTimeScalePosition >= 0 && oldTimeScalePosition < imageTimeScaleArray.Length)
    {
        imageElement.sprite = imageTimeScaleArray[oldTimeScalePosition];
    }
}
```

koosemose:
By visually respond I mean the up/down speed button darkens, but neither speed, nor the speed indicator changes. The not responding at all seems an extension of the first issue, as though either the checking of the mouse position is a little off, or the hitbox for the control is smaller than the actual control.

They may not necessarily need to be bigger, perhaps just a bit of a background behind them, so you have the entirety of a round/square button to click on, rather than having to hit the +/- signs.

BraedonWooding:
This will happen if there is no where to go?  Like if you are at the max / min speed.  With the little background I just added a more generous hitbox and so it treats it as a square so you can click above/below the minus sign and on the corners of the cross.  I added an image of the horizontal layout and smaller symbols above too.

hcorion:
I like the way Cities Skylines does it, screenshot, below (I'm talking about the bottom right corner). It's really obvious with the three arrows that it speeds up the game, and the play/pause button is really intuitive as well. Some obvious changes from green to red would really help as well.

BraedonWooding:
Yeh its similar to rimworld, the problem is that our time goes back and forward... like u can have slower and faster time scales (.25, .5, 1, 2, 4...) which mucks that up and we have 6 different options, this mucks that whole arrow thing up completely (it looks really really odd to have some arrows going back and some forward).  Now personally I see no reason while someone may want to slow down the speed (like 1x is pretty slow tbh), but while we have time scales lower than 1x the visual thing doesn't really look so good. HOWEVER, thinking about it what i could do is have the four arrows (if time scale > 1 and 2 arrows < 1x) then just like highlight the arrows so u have 2/4 arrows lit for like 4x or whatever.  So I'll go and do that 😀, again this kind of stuff is very finicky so we'll see.

hcorion:
Wait a second, so we can rewind?

BraedonWooding:
Not rewind just go slower speed then 'normal'. So if we place the game at 1x speed we can go half that speed 0.5x and a quarter speed 0.25x.  And i can't think of a way to represent that then arrows that point backwards?  If you can than that would be greatlyyyyyy appreciated.

koosemose:
For development the slower times can be VERY useful (things like being able to see exactly how a crew is pathing, what changes when, and so on. Pretty much anything you want to be able to keep a very close eye one or when you need to see what the state of several things are in a short time period. But I'm not entirely sure if it's useful for actual gameplay

dusho:
maybe just buttons would do for now.. until some art person comes with decent UI
image

hcorion:
What about a slider? Like this 
projectporcupine-1

dusho:
sure.. looks good as well.. can you make handle (runner) a triangle, instead of bullet? and maybe background can be bigger than handle? not sure

hcorion:
Yeah, that would all be a possibility, It was just mostly the default slider setup.

hcorion:
I was kind of thinking of having a separate button beside it for pausing/unpausing though, with a play / pause button, so I don't know if a circle would be better.

BraedonWooding:
I actually have another idea :D.  What if we squash all our ideas together??  And create this:
animation 0

dusho:
that backwards arrows are confusing.. it's not like it's going back in time (unless we want such a feature?)
also that mixing slider and pause button.. eh.. not sure
we do need some UI art guy that can come up with something clean and nice

BraedonWooding:
Yeh okay how about I make it so it just switches between the 1x arrow and the pause icon?  So it's just a pause button.  That'll work till some genius comes along it's may look slightly ugly but it shouldn't be confusing so it looks like this:
animation 1

hcorion:
I like that.

koosemose:
Keep in mind FPS may not always be on, and that placement to take into account the size of the fps meter you're using may look a little funny with no fps meter, or another level of fps meter.

BraedonWooding:
Yeh true, but this is where we run into space issues, top is for notification banner bottom is for buttons like construction with the entire right hand side used up for room/obj info. And the bottom left and top left corner used for stuff.  So it was more of a we have no space?  Maybe we could place it bottom left above options left of the mouse information panel?

koosemose:
You could also put it back where you had it to start (I was just trying to be thorough in my report), and treat that more as an issue with the headlines... they do need a good seeing to, both on a functionality level and on a presentation level (such as only using a predefined amount of space, or at least width, so they don't consume quite so much space for something that's only there occasionally... perhaps a ticker sort of thing would be fitting)

BraedonWooding:
Okay sweet will do :D.  I may give headlines a little attention soon since they do need a good R&E.

BraedonWooding:
**Note** the _World scene file thing (should disappear when I do the WIP things)

koosemose:
One last (maybe) concern, how feasible is it to have this respond to the hotkeys? i.e. If I press `+` to speed up the game or `space` to pause, rather than the indicator staying as is (as it currently does), have it reflect the current speed/pause state.

BraedonWooding:
Oooh yeh that's easy ish, i forgot to reimplement that ill do it in the morning.

BraedonWooding:
Should be done now @koosemose :D.

bjubes:
When you hit pause and then change the time scale using the keyboard, the graphic doesnt update until you unpause.
eg. hit 1, then pause, then hit 3. The scale indicator doesn't move. Then unpause, and it jumps up to 4x speed.

BraedonWooding:
Well that's cause the keyboard controls don't unpause the game when you cycle through the speeds, I will however make the slider move :D.

BraedonWooding:
@bjubes I had to do a semi-hacky solution, because whenever slider.value changes my OnValueChanged function gets called (Unity's side and it has to exist to detect slider changes UI wise), which sets pause to false, which would mean that whenever you use the time manager keyboard keys it also would unpause the game (which if that is wanted should be implemented time manager side), it's just a boolean that i flip before changing the slider value then flip back (minimum cost really :D).

koosemose:
The slider is a little peculiar when a new game is started (either from direct scene load, from intro screen new game, or in game menu new game. 

image

As you can see, the slider lacks coloration, and more significantly, it's starting on .75 speed... I'm reasonably sure this is just misreporting by the indicator, rather than the game speed actually starting off at a strange speed, but haven't investigated deeply.

BraedonWooding:
For me it starts at 1 everytime?

koosemose:
In that case it may be something with 5.5, could someone else running that give this a try and see what the indicator starts as?

BraedonWooding:
For some reason it's happening inconsistently sometimes?  I'm pinning it down as an initial update weirdness, and just kind of making it do it another way (essentially it sets the timescale position to the timescale position which updates my thing), and that for me ensures it to work all the time.  Tell me if this fixes things :D.

koosemose:
Everything seems to be working well, Merged.

Issue #1465, opened 2016-09-28
Planning 0.2 Milestone


koosemose:
We're coming close to the 0.1 milestone deadline, and while we don't quite have everything in place yet, I would like to go ahead and get the ball rolling on what we want to aim for in 0.2.

That being the case, I'll start with Listing things that either currently have open issues and no (known to me) current implementation plans. Not all may be appropriate for 0.2, but it's somewhere to get started. Feel free to suggest other things that I may have missed. Also note, I would prefer to focus this issue on deciding what is actually appropriate for 0.2, and what should be pushed to later milestones, rather than discussion the specifics of each feature. We can nail down specifics further once we have a more fully formed preliminary list.
#0.2
## Small
- [ ] Save/Load Jobs #177 #1590
- [x] Localize our localization (displaying US English as opposed to en_US, for example) #358 
- [x] Save Load Logic for power system #602 
- I'm not sure exactly what needs saved, other than the stored power on Accumulators
- [x] Asynchronous Saving #1459
- Any system which prevents PP from being reported as not responding would be acceptable
- Any speed ups in saving aren't necessary, but would be useful.
## Large
- [x] Integrate Power System and Utilities
- Rather than being universal, a power network should be anything that is connected to a given set of Power Cables.
- This requires utilities to intelligently build a network with all connected utilities of the same type.
- Should properly handle two power networks joining, or a single power network being split (by adding or removing cables)
- [x] Gas/Liquids/Fluids System @koosemose
- While there are currently gases that are kept track of in a dictionary in each room, some form of "Gas Network" would be desirable, in preparation for having gases transported by Utility pipes.
- Liquids should have a generally similar implementation.
- Liquid and Gas networks may be able to be implemented as a single fluid network due to the many similarities in how they would function within a network.
- Should roughly mimic how power is implemented, to ease future integration with utilities (Some of the parts of the power system may need altered to be more generic to allow them to be also used by fluids, most notably Grid and PowerConnection)
- Integration into Utilities is not required at this time, but is considered desirable if achievable at this stage.
- [ ] Metastockpiles 
- Allow Stockpiles to be grouped together, and request items and handle where they go as a whole rather than each stockpile individually
- May require Job system having distinct haul jobs, rather than hauling stuff to a stockpile implemented as a job that requires material.
- [x] Main Menu (Initial PR in #1461 )
- Should have New Game, Load Game, Settings, and Exit at a minimum.
- [ ] Core Game as Mod #772 
[...]

PR #1673, opened 2016-12-09
Universal fluid network

koosemose:
Per one of the goals listed in #1465

Currently implemented as a Universal system (similar to how Power was before it was integrated with Utilities. Leaving off integration with utilities because there are some broader considerations before that is done (Two notable ones, being rather a single fluid pipe should be treated as multiple pipes, and essentially allow any fluids to travel through it. And if not, should there be just a single fluid pipe that simply handles any fluid type but be typed to whatever fluid is put through it first, or should there be individual pipe objects for every possible fluid (perhaps not insanely complex, as presumably we won't have too many different fluids), and each of those has multiple possible repercussions and decisions to be made).

On the code side of things, the terminology used in FluidConnection FluidNetwork still reflects its origin as a copy of PowerConnection, variables and such will need to be renamed, and grid, which has now been repurposed for general usage by any utility network has not yet had names changed to reflect that.

I however do consider the actual implementation of Fluids to be complete for this stage of work. One aspect I would like to add, is for fluid providers to not produce fluid when there's nowhere for the fluid to go, but right now that would be mostly meaningless as the example water generator doesn't consume any resources to produce water, and I'm not sure how it would be made to do so in a meaningful way (only consume resource when it is producing. Additionally I would like the fluid system, rather than being somewhat all or nothing, as the power system is, to be able to provide fluid at a reduced rate when there's not enough things using it, and when there's too much draw on the system, to provide a lesser amount of flow to everything demanding its fluid (such that if there is a furniture requiring 2 water units per second, and the generator normally produces 4, it will only produce 2, and consume its resource at half speed. Or if there are furnitures drawing a total of 8 fluid (2 each) from the same generator, it will provide each with half the fluid it needs (again percentage based, so if one furniture has a higher draw, it will get a higher portion but still only a given percentage of its total requirement))

Note: Presently there are 3 furnitures making use of the Fluid system, the Water Generator (which presently doesn't require power), the tank (which can store any fluid, but right now there's nothing but water, and storing multiple liquids doesn't work so great with the Universal Fluid system, since there's no easy way to designate the fluid to be stored), and the Oxygen Generator that requires water (and power).

There's one little bug, that I'm not entirely sure of the cause, until the O2 generator has at least once had water and power, it will show the power missing icon even if it has power but has power, but once it has had power and water, it will properly only show the power missing icon (so if there's power and no water, it won't show the power missing icon).


BraedonWooding:
Works for me, after plugging in power and getting a water generator going the oxygen generator was running and creating oxygen.  Without water it stated it didn't have water (though that was a little weirdly worded and probably a localization thing), and the tank worked too.

koosemose:
YEah, the localization hasn't been set, and in some cases may still be using the power systems localization.

Assets/Scripts/Models/Buildable/Components/FluidConnection.cs
```diff
@@ -0,0 +1,305 @@
...
+            public List<ParameterCondition> ParamConditions { get; set; }            
+        }
+
+        [Serializable]
+        [JsonObject(MemberSerialization.OptOut)]
+        public class FluidConnectionParameterDefinitions
+        {
+            // constants for parameters
+            public const string CurAccumulatorChargeParamName = "pow_accumulator_charge";
+            public const string CurAccumulatorIndexParamName = "pow_accumulator_index";
```
koosemose: I realize this, as I mentioned in the original post, I haven't yet changed all the assorted names left over from its origins as PowerConnection, there's probably quite a few more beyond those two, and intend to change them all once I've hammered out the functionality, and gotten further input on said functionality.

BraedonWooding:
Tbh the power indicator is really buggy sometimes, it sometimes shows even with power and sometimes requires a rebuild to work, so it may be more related to that then your stuff?

Assets/Scripts/Models/Buildable/Components/FluidConnection.cs
dusho: the rest looks actually ok

dusho:
@BraedonWooding can you write steps to reproduce this? Some weird behavior may be related to wires requirement to be placed at 0,0 (bottom left corner) of multitile furniture

BraedonWooding:
#1677

koosemose:
@dusho, I believe I've taken care of all the inaccurate names in FluidConnection.

I believe this is ready for merge, pending testing.

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -380,6 +390,89 @@
...
+        </Component>
+    -->    
+        <Component type="FluidConnection">
+			<FluidType>water</FluidType>
+            <Provides rate="5"/>            
+			<!--
+            <Requires>
+                <Param name="cur_processed_inv" condition="IsGreaterThanZero" />
+            </Requires>            
+			-->
```
dusho: you would need to uncomment this as well, if you want it to use that 'ice' (previous comment)

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -380,6 +390,89 @@
...
+                <CurrentProductionChainName name="cur_production_chain" />
+            </ParameterDefinitions>
+            <ProductionChain name="Power generation" processingTime="60">
+                <Input>
+                    <Item objectType="Power Cell" amount="1" slotPosX="0" slotPosY="0" hasHopper="false" />
+                </Input>
+            </ProductionChain>
+        </Component>
+    -->    
+        <Component type="FluidConnection">
```
koosemose: At present I don't want to do that, because I feel it requires other things to be "proper". i.e. It should only actually produce water when needed, and only consume ice when it is producing water (and preferably at a rate matching it's water output), and I'm not entirely sure how to do that with the system as it stands.

dusho: this can be implemented in workshop component, but would require some additional flags.. and maybe some changes on how requirements (power, fluids, ... ) are handled (updated)

koosemose: I imagine being able to have the workshop stopped mid production (though I *think* that aspect can be done now, but just not externally), and producing slower would also allow some additional interactions, such as, rather than all or nothing power usage, power if you have half the power required, things work at half speed, similarly with fluid available, though of course that would require additional changes to power and fluids.

Assets/StreamingAssets/Data/Furniture.xml
```diff
@@ -380,6 +390,89 @@
...
+            <Requires rate="5" />
+            <ParameterDefinitions>
+                <CurrentAcumulatorCharge name="fluid_accumulator_charge" />
+                <CurrentAcumulatorChargeIndex name="fluid_accumulator_index" />
+                <IsRunning name="water_is_running" />
+            </ParameterDefinitions>
+        </Component>
+
+        <LocalizationCode>furn_accumulator</LocalizationCode>
+        <UnlocalizedDescription>furn_accumulator_desc</UnlocalizedDescription>
```
dusho: old accumulator name - should be something new here

dusho:
Hm.. there seems to be some problem with description of furniture - it's out of sync. But iy may have been introduced with cables already
image
power and fluid utilities are broken, yet furniture reports green

koosemose:
Well the fluid utilities don't matter. As I mentioned in the Original Post, the system is at present universal. And the issue with the power cables is present in master.

PR #1676, opened 2016-12-10
Atmosphere control dev commands

koosemose:
Adds dev console commands to control atmosphere, either of individual rooms (by roomID) or all rooms, also by individual gas or standard human breathable atmosphere mix, and finally commands to empty the atmosphere of rooms. Includes SetGas command in Room.cs (used in the DevConsole commands), to directly set the gas mixture in a room, rather than changing the already present gasses.

### DIFF
--- a/Assets/Scripts/UI/InGameUI/DevConsole/DevConsoleCoreCommands.cs
+++ b/Assets/Scripts/UI/InGameUI/DevConsole/DevConsoleCoreCommands.cs
@@ -280,7 +280,7 @@ namespace DeveloperConsole
         public static void EmptyRoom(int roomId)
         {
             Room room = World.Current.RoomManager[roomId];
-            foreach (string gas in room.GetGasNames)
+            foreach (string gas in room.GetGasNames())
             {
                 room.SetGas(gas, 0);
             }
@@ -292,7 +292,7 @@ namespace DeveloperConsole
             {
                 if (room.ID > 0)
                 {
-                    foreach (string gas in room.GetGasNames)
+                    foreach (string gas in room.GetGasNames())
                     {
                         room.SetGas(gas, 0);
                     }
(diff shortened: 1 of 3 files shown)


Assets/Scripts/UI/InGameUI/DevConsole/DevConsoleCoreCommands.cs
```diff
@@ -236,5 +236,68 @@ public static Tile GetTileAt(Vector3 pos)
...
+
+        public static void FillRoomWithAir(int roomID)
+        {
+            // Adding air to room
+            Room room = World.Current.RoomManager[roomID];
+            room.SetGas("O2", 0.2f * room.TileCount);
+            room.SetGas("N2", 0.8f * room.TileCount);
+        }
+
+        public static void FillAllRoomsWithAir()
```
bjubes: Im not too fammilar with the gas system, but does this function have a possiblity of over-pressurizing a room? like if a room was 100 percent hydrogen at the normal pressure, will it now be 50% H2, 10% O2 40%N2, making it not quite breathable and overpressured?

koosemose: Hmm, yeah, hadn't considered rooms with other gasses. I should probably 0 out any other gasses.

koosemose:
Semi-related thought: Perhaps 0 level gasses shouldn't be displayed (either by deleting the entry in the gasses dictionary or by just not displaying the gas when it's zero), since if we do start having more than the standard gas mix (such as a base on a planet with a toxic atmosphere) we could end up with having rooms display lots of different gasses that they no longer have, and aren't likely to have any more

Assets/Scripts/UI/InGameUI/DevConsole/DevConsoleCoreCommands.cs
koosemose: The latest commit takes care of this. The Gas commands can still overfill, but that's sort of the intent, to allow the user to bit by bit, build up the atmosphere of a given room as they please.

PR #1678, opened 2016-12-10
[WIP] DevConsole Revisions

BraedonWooding:
Obviously there will be a constant stream of new commands appearing as pull requests, but here I wanted to address some small/larger revisions that have been suggested by multiple people.  

The first set of changes is around the general feel of the DevConsole.  That is nicer errors, nicer text colours and sizes and just a general improvement to the 'feel'.  Now when you enter a command that doesn't exist if you enter "SetTimeStamdj" for example it will suggest that you meant "SetTimeStamp" via some really (and by that I mean REALLY) simple string comparison stuff.  Other than that it's also focusing on nicer error messages indicating paths to resolve what you wish to accomplish rather than just saying idk why it broke go fix it :P.

The second is default values (@koosemose I got overloads working too), this can be done in two ways, the first is function default values (only tested on C# but suspected to work on LUA), for example the Help function now implements this, you implement it in the function as normal and don't need to tell the dev console anything it's 'smart enough' to figure it out.  The second way to do it is through an extra (optional) parameter called DefaultValue, this should be respected as if you were creating a variable (therefore it should fit the syntax i.e. [] for objects, "" for strings, and just like normal stuff for everything else), and will fit multiple parameters using the commas.  A.k.a. the default value text option is just going to copy and paste that into the parameter list if it doesn't exist :D.  What this means is it's all or nothing, you can't have partial parameters (even though C# functions support them they are 'officially' unsupported due to just them not really being needed).

The third is a cleanup of the class structures.  I have removed the ability to just state the function and for it to gain the function name from that, also I require you always to have help text in the constructor.  These are not only to reflect the XML but also since it's more maintainable this way.  I will also be removing most of the commands and replacing it with a new command type called Internal.  In the end only the 'essential' commands (right now Help and Clear) will be implemented through the main assembly the rest will be mods (most likely all C# just due to the speed).

The fourth, is to implement tags.  Again another xml/object variable, just an array that modders/us can set on an object (purely optional).  Right now only plans for Help, though you can obtain a list via the normal means of DevConsole.CommandIterator(tag), or DevConsole.CommandArray(tag), if tag is "" (nothing) then it'll just count it as all objects.  This is just so you can say give me all with tag Room and so on cleaning up the Help command.  It will most likely have other uses later on (like maybe when you type a command we try to find a tag that's close and so on.

Why you may ask all these changes??? This 'DevConsole 2.0' is just about trimming off the fat, it's like saying this was a good idea and worked well but well we don't really use it, because of this (an example would be that I was thinking C# Modding would be .DLL so implemented commands with that in mind, when it turned out it was invoke only I removed the LUA one and made an invoke one which left 90% of the classes unused and while it was a nice reconciliation it wasn't really made for invoke commands).  And it's kind of improving generic things like errors messages and doing cleanups.  Also the addition of tags and default values were just thought up after using the DevConsole for longer periods of times (going hey we have a lot of commands, or well this kind of should have a default since you type this value 99% of the time).

I also just naturally did some (and will do some more) cleaning up of code, it's very minor and is like this comment isn't descriptive enough or whatever.  During this I do have plans to remodel some of the structures to make it more invoke friendly, but that will just be chipping away at speed improvements.

# Progress:
- [x] Implement Nicer Text
- [ ] Implement nicer error messages
- [x] Implement default values
- [x] Implement Cleanup of classes
- [x] Implement Tags
- [ ] Implement the moving of Help/Clear
- [ ] Implement the removal of most commands and addition of InternalCommand

Note: This should be merged after other DevConsole function additions, just for consistency (and using the new InternalCommand will have really no impact other than a single line).

### DIFF
--- a/Assets/Scripts/UI/InGameUI/DevConsole/CommandPrototype.cs
+++ b/Assets/Scripts/UI/InGameUI/DevConsole/CommandPrototype.cs
@@ -36,9 +36,9 @@ namespace DeveloperConsole
         /// <param name="description"> The description of the command.</param>
         /// <param name="helpFunctionName"> The name of the function to call to show help.</param>
         /// <param name="parameters"> The parameters that this class requires (a string in C# type formats and comma between them).</param>
-        public CommandPrototype(string title, string methodFunctionName, string description, string helpFunctionName, string parameters) : this()
+        public CommandPrototype(string title, string methodFunctionName, string description, string helpFunctionName, string parameters, string tags, string defaultValue) : this()
         {
-            ConsoleCommand = new InvokeCommand(title, methodFunctionName, description, helpFunctionName, parameters);
+            ConsoleCommand = new InvokeCommand(title, methodFunctionName, description, helpFunctionName, parameters, tags.Split(','), defaultValue);
         }
 
         /// <summary>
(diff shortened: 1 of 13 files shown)


BraedonWooding:
The branch name was bad so I renamed the branch causing me to have to delete the old one and move over the changes to a new one.

PR #1679, opened 2016-12-10
Reload debug channel controller via context menu

bjubes:
### The issue this fixes

Sometimes the debugger channel control window can have wierd bugs after changing to and from PR branches: 
capture

This can be fixed by deleting ChannelSettings.asset and then reloading the window, but requires a manual restart of unity.

### What this PR does
adds a nifty option in the right click menu of the DebuggerChannelWindow tab that allows you to reload the window without quiting unity.


koosemose:
Looks good. Merged.

PR #1695, opened 2016-12-16
[Testing] Chunking the map for furniture showing/hiding

BraedonWooding:
Note: I got all the code done in a single sitting, so its basically done but there may be bugs or other stuff :D, should also be fully stylecopped.
Here is a video of it playing.  The FPS drop at the start is _only_ due to the fact I alt-tabbed to start recording the gif :D.

animation 3

All this does is just assign a bunch of chunks through some pretty easy math, I do use Vector2 as ranges (there may be a case for using a Range Struct, but in all honesty it doesn't really matter).  Then it just abstracts EVERYTHING :P.  I.e. to remove a furniture from it you just call the ChunkManager (which handles the abstraction) and just say chunkManager.RemoveFurniture(myFurniture). It will find where that furniture exists (through some VERY fast optimisation, of using some matrix math to locate the location in the chunk array but it does have a for loop backup, but I have yet to make it have to resort to that :P).

Due to the math you may say it will 'over estimate' and make too many chunks since it does make square ones, and it may have squares that are never populated (due to it rounding everything UP to an Int), but the system handles this by capping chunks size (so actually every chunk won't be exactly the chunk size but that'll only be the end/bottom row if you have a width/height < 100).  It also will not include chunks that are FULLY out of the map range (so you won't have useless chunks), again some simple maths.  I haven't included any diagrams of the math since it's all heavily commented and it's like just general vector/matrix math :D.  Regardless with this I never dropped frames while moving the camera (except if I alt-tabbed then I would have slightly choppy frames for half a second but that's alt-tab fault not this).

This doesn't really change the furniture manager class really at all (and doesn't touch the furniture class), it just incorporates ChunkManager in.  Also it doesn't need to have the delegates for the everyFrame and fixedFreq. in furnitureManager it just assigns them to ChunkManager.  Also ChunkManager 'copies+pastes' the logic from FurnitureManager (that is on EveryFrame only do visible furniture which is now visible chunks and on FixedFreq. do EveryFrame calls on Furniture that isn't visible, and do FixedFreq. on visible furniture).  Anyways cool stuff eh?

Currently it's set to make chunks at a 10% frequency (so 10% of width and 10% of height) so that means for a 100x100 map we will have 100 chunks (100*10% * 100*10% or 10*10), so that means every chunk is a 10x10 area, which is pretty freak'n small if you ask me (like I doubt we'll ever really want smaller chunks... but hey it's a good test), so we may want to make it 20% frequency (so every chunk is a 20x20 area) but idk 10x10 is a nice area.

Anyways have a go and test it :D, experience the full frame rate and the POWER of chunking.  lol I'm a little overexcited.


koosemose:
Not sure if it's just an error in your description, or if the error is reflected in the code, but you seem to imply it only does the FixedFreq calls of visible furniture. Is that reflected in the code?

BraedonWooding:
It's probably me typing a confusing description, it does the exact same effects as the old system.  There are two functions, one called EveryFrame, in the old system it only called that on visible furniture, in the new system it only calls it on visible chunks.  The other function FixedFreq. in the old system used to call FixedFreq. on visible furniture (and non-visible) and everyFrame on invisibleFurniture, the new system calls FixedFreq. on all chunks and everyFrame on non-visible chunks.  So no functionality change whatsoever in this new code.

dusho:
So you created something similar to quadtree optimization. Chunk and ChunkManager names sound a bit off.
Is there a reason why you use CopyTo for those furniture in chunk updates?

BraedonWooding:
@dusho Kind of yes, Chunk and ChunkManager I think are right on spot since they represent Chunks of the map kind of similar to Minecraft Chunks (which is where I got the name) though I would be open to any name that you can think represents it better :D.  Yes there is a reason, I'm converting the hashset to an array (faster to iterate through, and more safe, since I want indexes in some of those cases) and toArray() isn't standard for HashSet (it's included in Linq) and due to those reasons I preferred to use the more standard version for HashSet (since not only its faster it basically doesn't require linq, and the ToArray method just does some weird iterate thing anyway).

dusho:
Doesn't Utility need something similar? I think need to update IBuildable to cover all these needs to prevent copying of same code for Furniture and Utility classes.
Also, regarding that CopyTo().. maybe you can have in Chunk class array with furniture as additional member variable (together with that HashSet) and create it with that CopyTo() only when add/remove is called ? It would be less heavy on copying (no need to copy every frame) and would keep the performance...

BraedonWooding:
I have no problem making it IBuildable though I doubt if it'll really impact so heavily since the main use of this is multiple update functions (legit really the only use) and Furniture is really the only IBuildable with multiple update functions (and if I make it IBuildable the thing could actually become slower since I'll have to do a check if something is furniture to do the multiple updates).  Also (even though you haven't suggested it I want to point this out) I don't really want to make it generic either due to the fact it would require less abstraction and 'is' calls.  On the second point about CopyTo; personally it really doesn't matter, like we are talking small differences in speed, but sure I'll do that and it would reduce load on GC/Memory (slightly, like real slightly... like I doubt you would notice it slightly).  @koosemose you know quite a bit of utilities from your liquid/gas work do you feel it needs something like this?  Or is it better to leave it till it REALLY does (since it really won't ever be near the count of Furniture, and to the best of my knowledge it doesn't currently have the visible/non-visible functionality)?

koosemose:
Utilities does almost nothing, pretty much all it does is contain a grid, and network similar utilities so that furnitures can communicate to the same grid, and networking only happens when a utility is added/replaced, which is only going to be at max once per frame in regular gameplay (the only time it won't be is in dev mode or on save/load, it has no need of anything relating to updates... I think I may have left the update functions in there, because when I was designing it, I wasn't entirely sure rather or not it would be needed, but as I develop the surrounding systems the more sure I become that they're not needed.

BraedonWooding:
Okay awesome, I didn't look just was asking anyway since I felt they may have been.  So from that it doesn't need chunking and it would just unnecessarily complicate things (and possibly lead to more bugs).

dusho:
ok. Then is sounds Utility really doesn't need it.. I just remember some copy-paste code from Furniture in Utility and was wondering.

koosemose:
Yeah, there was a lot of stuff I left in Utility because I wasn't yet certain exactly how things were going to work, I've been working on trimming the fat out as I revisit it with a better idea of how it's used.

BraedonWooding:
Okay so what do we think we should do with this?  I'm happy to resolve conflicts and apply wanted changes (just remove the creation of the array every frame that was mimicked from old behaviour), but if we moving to slowing down updates then it may just be better to close this?

dusho:
I would still prefer this optimization (minus that CopyTo) over spread update in current state. Spread update can be integrated into this later on.

BraedonWooding:
Yeh that's what I meant by apply wanted changes the minus copyTo.  @koosemose @NogginBops what do you think?  Want to get a decent consensus.

koosemose:
The main problem with this is that it is a fix to a fix, making the vis/invis tracking more efficient, which we needed to do because our update cycle was inefficient, wasting a ton of time calling updates that didn't actually do anything. The spread updates performs better than this on camera move (obviously since it removes the vis/invis tracking), and performs drastically better with things on or off screen because only what needs to be updated is being updated, and the slow updates are no longer all being performed on a single update.

I'm not sure how one could merge this with spread updates without making things worse by clumping updates as chunks with more furniture are updated all at once (and functional furniture will almost certainly tend to be clumped into one or two areas in actual play). And there's also the issue that moving things between update types will cause things to miss or gain more updates as things move around.

BraedonWooding:
Okay fair point, yeh it was just a fix to a broken system and it won't merge so nicely as you were saying.  So what i'll do is just close it for now :D.

PR #1696, opened 2016-12-16
Ignore ProjectSettings.asset

BraedonWooding:
As the title says, update the .gitignore to ignore project settings.  Won't affect current PRs, but future.


hcorion:
While your at it you could ignore .DS_Store

BraedonWooding:
I can't find any .DS_Store (I know it exists but search is revealing nothing), can you give me an file address for it?

hcorion:
Just add `*.DS_Store` it is a Mac only file that is created whenever you open up a folder in the file manager.

BraedonWooding:
Ohhh got it !

koosemose:
We cannot ignore ProjectSettings. If we do any new developers (who don't already have the current ProjectSettings, will be unable to compile.

BraedonWooding:
Oh okay fair point

PR #1697, opened 2016-12-17
Removal of the Character state list for a hashset

BraedonWooding:
Just what the title says removes a list in the Character class for a hashset, it doesn't have grand results but I do notice about a 2-3 frame difference on dips (like every 5-12 seconds it dips to ~35 now it dips to around ~37+.  It also just doesn't need to be a list since there isn't any need for order :D.

### DIFF
--- a/Assets/Scripts/Models/Character/Character.cs
+++ b/Assets/Scripts/Models/Character/Character.cs
@@ -81,7 +81,7 @@ public class Character : ISelectable, IContextActionProvider
     private State state;
 
     // List of global states that always run
-    private List<State> globalStates;
+    private HashSet<State> globalStates;
 
     // Queue of states that aren't important enough to interrupt, but should run soon
     private Queue<State> stateQueue;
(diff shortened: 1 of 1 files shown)


koosemose:
I'm not sure if you're actually seeing the results you think you are seeing, this is a small enough container that either the difference shouldn't matter or list should be slightly better. 

An important step in comparing changes that I expect you didn't do in comparing, is setting your seed (has to be done in code), because different asteroid formations (i.e. more or less furnitures generated for asteroids), can cause a more significant change in ms per frame (and in turn fps) than the 2 frame difference you've seen.

To set the seed in code, in World.cs, insert after line 53 `seed = <someNumber>` (or just change line 53 to use that number rather than random, I use the insert method so I can quickly delete it when I'm done), that way, when you're comparing performance, you can reduce outside factors such as the amount of furnitures generated for asteroids.

BraedonWooding:
Sure, okay I'll try that.  I should also add that a hashset (though maybe it should be a set, since hash sets aren't so good for small sets of data), is good since the global state list should be distinct?  And having sets just are generally nice for this reason.

koosemose:
I think with the small size of globalStates we would be better served leaving it as a List, which would allow us to switch over to using a for loop rather than a foreach loop, (and accessing the state directly as globalState[i] rather than creating a new object), which should be a painless way to less our garbage collection problem.

BraedonWooding:
I agree, I was going to delete this tbh but it kind of got hidden :P

PR #1698, opened 2016-12-17
Changed the double queues to a single array in CharacterNameManager

BraedonWooding:
I was just looking for lists to change to hashsets, and came across the name manager which utilised two queues to maintain a cycle of names.  This just replaces that with a single array and an int (which acts like a pointer), and it just cycles that pointer till the end then cycles it over again, overall this is faster and requires less space and is faster, it doesn't have to create a new array (when adding new names) it just sets the array to a concencated one so no weird heap stuff there either.  I know this is probably quite subjective and probably quite a few people think that it's not really necessary, and kind of random but I just came across it randomly and thought hey I'll propose a different structure that's a little more typed :D (and it actually requires less GC since it actually doesn't have to remove/add elements so that's kinda significant).

### DIFF
--- a/Assets/Scripts/Models/Character/CharacterNameManager.cs
+++ b/Assets/Scripts/Models/Character/CharacterNameManager.cs
@@ -49,7 +49,7 @@ public class CharacterNameManager
         }
 
         // Re-Loop Pointer
-        if (ptr > 0 && ptr >= characterNames.Length)
+        if (ptr >= characterNames.Length)
         {
             ptr = 0;
         }
--- a/Assets/Scripts/Models/Character/CharacterNameManager.cs
+++ b/Assets/Scripts/Models/Character/CharacterNameManager.cs
@@ -6,7 +6,6 @@
 // file LICENSE, which is part of this source code package, for details.
 // ====================================================
 #endregion
-using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


BraedonWooding:
The multiple commits were because I removed the scheduler changes wrong and *facepalm* that caused problems.  Regardless it should say only one file changed now *wipes brow*

Assets/Scripts/Models/Character/CharacterNameManager.cs
```diff
@@ -49,16 +42,20 @@ public static void LoadNames(string[] nameStrings)
...
+        // If character names doesn't exist then just return null
+        if (characterNames.Length == 0)
+        {
+            return null;
+        }
 
-        // We run out of used names, lets start using the used ones
-        if (unusedNames.Count == 0)
+        // Re-Loop Pointer
+        if (ptr > 0 && ptr >= characterNames.Length)
```
bjubes: when would ptr be less than 0? doesnt it start at zero and keep increasing?

BraedonWooding: Its > 0 so it won't set 0 to 0, but honestly doesn't need to be there (since the array can't be length 0 so it'll never trigger as 0), so I'll remove it :D

PR #1699, opened 2016-12-17
Scheduler hash set

BraedonWooding:
Just changes the scheduler to use hashsets instead of lists, also adds a purge counter to delay GC since as @dusho has been talking about GC is too high and causing problems.  It also just cleans up small parts of the scheduler in general :D.


bjubes:
Since you changed the scheduler code, make sure you update the related unit tests. You can run them locally in `window > Editor Tests Runner`

BraedonWooding:
Ooops okay will do

BraedonWooding:
Closing because we require multiple elements that are the same and hashsets are distinct.

PR #1701, opened 2016-12-18
Failsafes for pathfinding from null room

koosemose:
Currently, in master, any pathfinding that uses room pathfinding will fail if starting from a null room (such as a door. This first has it use GetNearestRoom, so that it will look for an adjacent room, failing that it will now also try to get the nearest room via pathfinding, and for the worst case scenario (room filled with doors, and not connected to any other areas, so there are no room objects accessible, it should just try to dijkstra pathfind to what it is looking for. I feel this worse case is going to be exceedingly rare, and further if it does happen will be unlikely to be a large room, so dijkstra shouldn't be a big performance hit, and if it is, this is a player trying to push the edge of what is reasonable and likely is doing it for no reason other than to break the system, this shouldn't break in that situation, it will just drop frames while it pathfinds in the most inefficient way possible (since it can't make use of room pathfinding due to there not being any rooms)

### DIFF
--- a/Assets/Scripts/Models/Buildable/Tile.cs
+++ b/Assets/Scripts/Models/Buildable/Tile.cs
@@ -6,12 +6,14 @@
 // file LICENSE, which is part of this source code package, for details.
 // ====================================================
 #endregion
+
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


koosemose:
Merging this now as it fixes a rather annoying bug in master.

PR #1702, opened 2016-12-18
Overlay for power and fluid networks

dusho:
Introduces new overlay for power and fluids. This should help debug problems with furniture being on/off grid.
image
- adds new Palette coloring with some basic colors
- new OverlayFunctions.cs script
- better fallback when compilation of C# scripts fails


koosemose:
Looks good, should be useful for debugging the appropriate systems. Merged.

PR #1703, opened 2016-12-19
Better claim system

koosemose:
This still needs a bit of work, but currently there is currently an issue in master with the claim system that can lead to issues building.

The issue is that, as implemented, the current claim system claims a stack of inventory in its entirety, so if there is one stack of 50 copper wire and you place down multiple power cables to be built, a crew will accept the job for one of the cables, claim the inventory, and go to grab it to proceed with the job, but if another crew goes to take another of the cable jobs, he can't access the inventory because it has been claimed, so the job is marked as uncompletable. This will repeat with some jobs being marked as uncompletable, and other's being taken, depending on where it falls relative to the inventory being claimed. 

There are two issues at play here, one being that the crew marks the entire inventory as claimed (even though they may only need 2 of them) since that is the only way it can be claimed, and that a released claim doesn't allow a job to be re-added to available jobs. This fixes the first issue by making claims a list on an inventory that contains a struct that has the time of the claim (so they will still autoexpire, with the list being periodically cleaned out as new claims are added), the character with the claim (so a character can release all their claims on an inventory, rather than having to store the InventoryClaim struct or something else to find the right claim), and the amount claimed so that, if there is 4 copper wires, two characters can claim 2 each, but no more can claim (and therefore won't be able to access them).

I intend on investigating how to make sure the job is added back to the queue when a claim is released (in addition to ensuring that claims are explicitly released as often as possible, leaving the autoexpire as only a fallback).

Update: I *believe* I now have inventory being properly reevaluated when a claim is released, but only when that claim brings the available inventory above 0 (as my understanding is that an in process job will take whatever inventory it can, even though it's not enough to be built, so it should only be marked as not having enough inventory when available inventory is 0), and also release a claim when the job is interrupted (more accurately when move state is interrupted, as that is the only time period a crew should have a claim, as it's claimed in HaulState, and immediately goes to MoveState, once they arrive, they will pick it up as they reenter HaulState)

Once it can be more certainly confirmed that this all works as intended, I will most like set the time for a claim to expire to be much longer.


dusho:
aaah.. so that explains dudes not taking jobs (we need better logging or something to see this right away)
Looks like it's getting more complex, and I'm not sure about how the API to outside looks like now. In most of the 'mods' (lua scripts) stacksizes of inventory were manipulated (which is kind of weird to begin with).
Will that be still ok?
Ideally there should be one common way to deal with inventories ( InventoryManager?) with set of basic functions that used and no other access (that may break claim, lock or stacksize) should be allowed..

koosemose:
I'm not sure why manipulating stack size is weird, if something is changing the amount that's in the stack, then that'd be pretty much the way to do it. 

The complexity from an outside perspective is nearly the same, it just requires a little more information to be passed into the function. Claiming really only matters for characters, they're the only thing that interacts with inventory from a distance/with a delay (they decide they want it and then have to walk to get it). The claim system in general is basically just a tool for the characters to communicate their want of an inventory, and all they have to do is check CanBePickedUp (which they already had to do) and then claim it if they want it.

And to the best of my understanding of C# access levels, there's no way we can really put Claiming in Inventory Manager and not have either a complicated structure for it to store the claims itself or control the claims in the individual inventory (without said inventories giving everything access to the claim mechanic through a public function since it would need to be public for Inventory Manager).

Regarding logging, now that we can log more freely without bogging things down, we can start having more verbose logs that will aid in things, but I doubt we can have completely comprehensive logging that will narrow down every problem without filling a channel with so much noise that it ceases to be helpful. Sometimes it will require targeted logging looking to narrow a particular problem down.

dusho:
by weird I mean LUA code I saw repeatedly where stacksize minus required amount was checked if it's >= 0  and then it was subtracted.. instead of having calls like CanTake(int amount), Take(int amount)
but if it's working well with existing scripts then it's ok, I guess

koosemose:
Ahh I see, can you recall what particular Lua code that is? The only Lua code I know of that interacted with inventory was machines using it, which was superseded by workshop. If there's any remaining I'll investigate it, but it should work fine, CanTake is, to the best of my knowledge, used only by characters and is designed as such. But if you can point out some Lua still interacting with inventory, I can look into it to make sure things still act properly, and if not figure out how to get things to play nice.

koosemose:
Everything seems to be working as intended, removing WIP, and claim duration increased to 2 minutes (in addition to making a constant field, rather than hard coded.)

bjubes:
I'm getting `Character could not find a path to the job site.` errors when placing inventory in dev mode while multiple characters are idle. maybe they are all trying to haul the same stack to a stockpile?

bjubes:
I'm not seeing any problems either. merging.

PR #1704, opened 2016-12-19
Furniture status indicators (power, fluid, production, ...)

dusho:
Introduces additional status indicators for furniture to easily identify what is missing for it to function.
Also reduces creation of gameobjects a bit as only those with requirements (components) will get gameobject with sprite for status
image


koosemose:
Does missing production mean it doesnt have a production chain set or that it can't produce anything?

dusho:
@koosemose production chain set

BraedonWooding:
It should have a no materials indicator too (if it's not already implemented).

dusho:
was thinking about that as well.. but at some point you want your machines to lie there without processing anything.. and then you probably don't want to be spammed with warning icons

I wanted to have those icons changing alphas (softly blending) instead of just having static picture.. would that be ok?

BraedonWooding:
Depends on performance impact I would presume.

koosemose:
In most cases I agree with dusho, that a machine without inventory to process just isn't doing anything because there's nothing to do, so shouldn't have a warning... However, it is also used for things that require inventory to perform another function (such as the fission reactor), and in those cases a warning would make sense. Would there be a way to make an inventory requirement only show in those circumstances (presumably through a setting in the data file), so that regular processing machine's aren't showing it, but something like a generator that requires input to work would show that they need supplied with something.

dusho:
hm.. ye.. that would make sense
trying to think about way to have it somehow generic -> if we would have attributes for inventory, we can hook indicator if inventory has some property, like mentioned fuel for fission reactor
That would even open possibilities to have input for workshop have just some category of inventory. Then input for fission reactor can be plutonium, radium, uranium.. with different fuel power (so similar how factorio has fuel power in MJ for wood, coal, solid fuel, rocket fuel... )
But that may be something for future maybe? if we go that way

koosemose:
I see that you've added a Requirement for gas, but no accompanying graphic, so it just shows a red square. I'm not sure what image to suggest for gas however. Perhaps a cloud-like shape or blowing wind sort of thing? Those are the closest I can think of to represent gas.

BraedonWooding:
Maybe like a tear-drop?  Sometimes that's used for gas?

dusho:
yes, cloud is a good idea.. here is what I managed to come up with for now
image

koosemose:
It'll do, though I think that one in particular is probably still on the list of images looking for something better. 

Note: you have a stylecop violation failing your checks.

koosemose:
Things seem to work as expected. Merged.

PR #1708, opened 2016-12-21
[Fix] Reconnect FilterField to FilterFunction

### LINKED ISSUE #1700: Construction Filter no longer working
### Expected behavior
typing  into the filter in the construction menu should search for furniture that matches the string and remove others from the list.

### Actual behavior
nothing happens. This feature was working at some point so if anyone has a guess of when it regressed or can find  the commit/PR that broke it by bisecting please post it.

koosemose:
Fixes #1700

Due to changes in ConstructionMenu the Filter was no longer working. This simply adds the Listener through code. It also adds the listener to the similar Filter in the order menu, which while not strictly needed at this time, due to the small size of the order menu, it does have the Filter field present and may be needed in the future.


BraedonWooding:
Title has a typo :P.

PR #1710, opened 2016-12-22
Corrected migration date

Dan54:
Fixes date of migration being January the 1st of 2016 from PR #1709

PR #1712, opened 2016-12-22
Migration correction

Dan54:
Changed 'are currently migrating' to 'may soon migrate'. Was going to put this in #1710 but it was committed before I could


bjubes:
It may be best just to remove the Unity 5 section all together until we actually plan to upgrade again

PR #1714, opened 2016-12-23
[Tiny] Mild Pathfinding Optimization

koosemose:
This is an old change I made while experimenting on ways to improve the pathfinder that never actually made it's way into any PRs. It simply changes the Priority queue used for Pathfinding to LIFO instead of FIFO for equal Priorities. It's a significant improvement when using Dijkstra (on the order of 200ms per frame [that has to pathfind] difference), though with the other changes we don't have to rely on Dijkstra as often anymore. It's of course a much smaller improvement on other pathfinding, but as pathfinding is such an integral part of PP, I feel it is still enough to be worthwhile.

It doesn't change any significant behavior, when two tiles that could lead to the goal are equivalent, it chooses the first, rather than the last, so will end up with a path that is slightly different from what would otherwise be chosen, but takes the same amount of time.


BraedonWooding:
I tested it out, and it all worked well with no problems (on a clean master + this).

PR #1715, opened 2016-12-23
Remove Utility's Update functions

koosemose:
When I created Utility.cs off the basis of Furniture.cs, I left in the update functions, because at the time I was uncertain exactly how various aspects of the actual networks would be implemented, and wanted to leave them in in case they were applicable, however, now that I've started integrating the appropriate systems, I can see they're not going to be used. This removes the update functions, and the various calls to them that did nothing but waste a bit of time.

### DIFF
--- a/Assets/Scripts/Models/Area/World.cs
+++ b/Assets/Scripts/Models/Area/World.cs
@@ -511,7 +511,6 @@ public class World
     {
         CharacterManager.Update(deltaTime);
         FurnitureManager.TickEveryFrame(deltaTime);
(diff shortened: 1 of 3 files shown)


BraedonWooding:
I would say since you created them, and the check passed just merge :D.

koosemose:
Doesn't matter that I added them, it doesn't mean that someone else doesn't have some reason they should stay or perhaps is working on something that hinges on them. So as usual, I'll give it a bit of time for review and input.

BraedonWooding:
Fair point, okay :D.

PR #1716, opened 2016-12-23
[Experimental] Spread Furniture Slow update across frames

koosemose:
This is an experiment to alleviate the spikes in performance, where 5 times every second, all slow updates and fast updates of all off-screen furnitures are ran. Roughly it divides the relevant collections into 10 parts and runs the updates for that 10th on the fast update, moving to the next 10th on the next frame, keeping a rolling count of the deltatime. This greatly smooths the performance (a very smooth 60fps, and smooth enough that it could potentially hit 120fps with no v-sync), but I'm unsure about some of the implementation details, most particularly, if a furniture is added in the middle of an update cycle (as in during 1 of the 10 frames rather than at the beginning), it could cause a furniture to be missed (since we're using hashsets), or possibly ran twice. Also I'm not fond of the slow update being implemented outside of the TImeManager, and no longer having an connection to the update time set in there, but I don't see how it can be done that way, since it needs to be called every frame (or nearly every frame) and has to be aware of the details of the furniture to properly dispatch calls.

I am looking for input on ways to improve this, and if any one has any ideas how this could be handled more in TimeManager, rather than FurnitureManager handling all the implementation details.

As this is Experimental and will need more work to be merge ready (and will likely have to be completely retooled when/if #1695 is merged), I will likely not be worrying about details like stylecop, or other style related things, and am just looking for input on how to improve the mechanics of the system. 

Whatever is submitted for final merging is likely to be an entirely new branch just using the concepts from this branch.

Paging: @BraedonWooding , @dusho

### DIFF
--- a/Assets/Scripts/Models/Buildable/FurnitureManager.cs
+++ b/Assets/Scripts/Models/Buildable/FurnitureManager.cs
@@ -172,6 +172,10 @@ public class FurnitureManager : IEnumerable<Furniture>
         return furnitures.Where(filterFunc).ToList();
     }
 
+    private int invisibleFurnitureProgress = 0;
+    private int furnitureProgress = 0;
+    private float[] accumulatedTime = new float[10];
+    private int timePos = 0;
     /// <summary>
     /// Calls the furnitures update function on every frame.
     /// The list needs to be copied temporarily in case furnitures are added or removed during the update.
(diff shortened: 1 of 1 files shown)


BraedonWooding:
I quite like this idea.  I suggest we add a new tick, so we have EveryFrame, Fixed Freq. (though this may be removed), and a new one (maybe Every10Frames?), and it just does the kind of code that is done in the tick every frame update?  Or as you said if chunking is approved and decided a good thing (which I'm pretty confident it will tbh), then this stuff could be done in the ChunkManager (which I would still say is about as good as being done in the furnitureManager, so I still like the idea of a new delegate/frame updater).  Now that is a relatively easy implementation but let's say you want to make it so it doesn't need to add a new delegate, well then could we add a new parameter to all the updates something like (float deltaTime, int frames), where the frames represents the current frame counter (it resets every second), now what you could do is just a simple is frames % 10 == 0, to check every ten frames or whatever, this offloads major amounts of the code to timeManager, which just could have an int counting up (and resetting every second, or maybe every tick freq. idk).  I quite like this idea since other systems could use an update so it's every second frame rather than every frame, or every third frame and so on.  So each system can have a tailored time system WITHOUT having to have multiple delegates/hooks :D.

dusho:
trying to think how would it behave if someone had less than 10fps for some time..
I believe I mentioned in chat something like array of timers with same or different update times.. (like 10 timers with frequency 10ms, 10 timers with 100ms, 10 timers with 200ms, ...), timers would have different start offsets and furniture (or anything that needs update with some frequency - buffs, effects) would just 'subscribe' to some timer. Some manager (time?) would balance it out what furniture will get what update (timer).

koosemose:
As for how it will behave under low fps, it will behave just the same, just the updates will be slower, exactly how updates are slower for fast updates when fps is low, just 1/10 that, and delta time will still be correct as the delta time for its slow update is just the accumulated delta time for the last 10 frames.

As to the large number of other timers, I don't think that would work out well (or possibly at all) it is extremely difficult (if not impossible) to get that much precision in your update times (particularly down to 10 ms), and if Time Manager did it, it would have to have a large amount of awareness of furnitures to dispatch them in any reasonable way.

koosemose:
I have a separate branch I am just starting experimentation with, around the basis of an IUpdatable interface, which has Managers that currently subscribe to TimeManager updates, and then dole out the updates to their constituents (FurnitureManager and CharacterManager), instead, on creation of the object register that object (as an IUpdatable) with TimeManager, allowing TimeManager itself to handle spreading out the slow updates between frames.

Currently it doesn't have visible/invisible furniture implemented (or IUpdatables I suppose), as I need to port that from the furniture class over to TimeManager, but even with it doing full speed updates even on off-screen stuff it has an appreciable effect in smoothing out the framerate. I believe that once that is reimplemented it may at least match this PR in performance, if not surpass it, along with putting the handling of updates back in the hands of TimeManager, along with allowing things like Characters to benefit from having slower updates even when offscreen.

I'm not certain if it will be compatible with chunking performance wise, as the improvements gained from both this and the alternative version I'm experimenting with rely on having a roughly even amount of objects being updated per frame, and doing updates on a chunk by chunk basis wouldn't give the effects wanted (as chunks will tend towards having either lots or few objects in them, such as an asteroid or the base being each mostly in a single chunk), and the process of creating a list of all furnitures in visible/invisible chunks *may* create more overhead than it gains. Though of course I would be happy for them to be able to work in concert, if they can't I believe this achieves a better performance.

koosemose:
Closing this as I believe the other branch functions better, and with a more reasonable structure.

PR #1721, opened 2016-12-26
Spread Slow updates across frames

koosemose:
Building on the concepts in #1716, this implements IUpdatable, a simple interface that makes sure the class has the proper hooks needed for an object to receive its updates directly from TimeManager. 

Basically, simple slow updates can still be handled as before (it's not necessary that they are, but they can be), but updates such as those currently done in FurnitureManager and CharacterManager can instead register their respective objects with RegisterSlowUpdate or RegisterFastUpdate (and unregister on object removal) instead of doling out the updates themselves. 

Time Manager now keeps an array of lists (set to match the number of frames to spread the updates across), and when a new slow updatable is registered (as fast updatables are updated every frame there is no need for this, but can be easily added by duplicating what is done for slow updates, if it's ever decided to have fast update somewhat slower than every frame), it will add it to whichever list within the array has the fewest elements, keeping each list approximately equal, and it cycles through the list updating them across the 10 frames, with a rolling array of the delta times, so each will have the proper delta time. Furnitures can now report if they need an update of either type (rather they have the matching Lua update EventAction, and querying each of their components if the need them, most components will, but a few will not (such as a Visual that only sets default sprite name), and furniture manager will only register updatables that need updates (saving a tremendous amount of time doing updates on non updating furniture, which will allow us to the amount and size of asteroids for example). 

With the current methods used, time is only lost/gained when an updatable is first added, and it would be less than 1 second, as long as fps is at least 10 (in which case we have much bigger worries than more lost time).

There are no longer separate lists for visible and invisible, as, even on a Stress Test ( https://gist.github.com/koosemose/0fc1cd920af39b3b9fb8a7a6ee666d35 ), the time taken maintaining the lists far outweighed the benefits (with only furniture that needs updates being callled, there is virtually no visible benefit, and in most cases the bulk of functional furniture will be on screen anyways. I am not opposed to it being readded, but I am unable to figure out a way to do it that doesn't greatly decrease the benefits of this, or cause an assortment of other timing issues.

As is, this improves on the previous work in #1716, both achieving better performance (consistent 120-150fps with vsync off), and is structured more appropriately (with handling more update related stuff in TimeManager).


koosemose:
Another thing this allows us to do is arbitarily have certain objects not get updated. For example, Furniture manager won't register either update if a furniture is a wall, and won't register a fast update if a furniture doesn't have either an OnFastUpdate eventAction or any components. These could use tuned, but do improve performance on a stress test (using https://github.com/koosemose/ProjectPorcupine/tree/Big-Asteroids ). I don't like hardcoding walls not to update, but without a better, more general idea of what to use for what not to update, I can't avoid updating them without  explicitly ignoring them. This could also possibly use some form of support to tell if a component actually needs regular updates (for example, a VIsual compoenent with only a defaultSpriteName probably shouldn't need regular updates).

koosemose:
Merging

PR #1722, opened 2016-12-27
Money counter

buggy213:
See #1075 

This PR adds a little textbox to the UI in order to display the amount of any given currency a player has.

ppmoney

### DIFF
--- a/Assets/Scripts/UI/InGameUI/MenuTop/CurrencyDisplay.cs
+++ b/Assets/Scripts/UI/InGameUI/MenuTop/CurrencyDisplay.cs
@@ -1,0 +1,31 @@
+﻿using UnityEngine;
+using UnityEngine.UI;
+using System.Collections;
+using System.Text;
+
+public class CurrencyDisplay : MonoBehaviour {
+
+    public Text text;
+
+	// Use this for initialization
+	void Start () {
+	    
+	}
+	
+	// Update is called once per frame
+	void Update () {
+        string[] currencies = World.Current.Wallet.GetCurrencyNames();
+        StringBuilder text = new StringBuilder();
+
+        // Populate the text box
+        foreach (string currency in currencies)
+        {
+            text.Append(currency + ":");
+            text.Append(World.Current.Wallet[currency].Balance);
+            text.AppendLine();
+        }
+
+        // Don't question it ok.
+        this.text.text = text.ToString();
+	}
+}
--- a/Assets/Scripts/UI/InGameUI/MenuTop/CurrencyDisplay.cs
+++ b/Assets/Scripts/UI/InGameUI/MenuTop/CurrencyDisplay.cs
@@ -13,22 +13,27 @@ using UnityEngine.UI;
 public class CurrencyDisplay : MonoBehaviour
 {
     public Text text;
-    
+    string[] currencies;
+
+    private void Start()
+    {
+        currencies = World.Current.Wallet.GetCurrencyNames();
+    }
+
     // Update is called once per frame
     private void Update()
-    {
-        string[] currencies = World.Current.Wallet.GetCurrencyNames();
-        StringBuilder text = new StringBuilder();
+    {    
+        StringBuilder content = new StringBuilder();
 
         // Populate the text box
         foreach (string currency in currencies)
         {
-            text.Append(currency + ":");
-            text.Append(World.Current.Wallet[currency].Balance);
-            text.AppendLine();
+            content.Append(currency + ":");
+            content.Append(World.Current.Wallet[currency].Balance);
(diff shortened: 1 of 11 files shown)


ProjectSettings/ProjectSettings.asset
```diff
@@ -29,6 +29,7 @@ PlayerSettings:
   m_StackTraceTypes: 010000000100000001000000010000000100000001000000
```
hcorion: Undo the changes to this file, we don't usually change these files.

buggy213: huh. good point. I'm not exactly sure why that happened...

Assets/Scripts/UI/InGameUI/MenuTop/CurrencyDisplay.cs
```diff
@@ -0,0 +1,34 @@
...
+using UnityEngine.UI;
+
+public class CurrencyDisplay : MonoBehaviour
+{
+    public Text text;
+    
+    // Update is called once per frame
+    private void Update()
+    {
+        string[] currencies = World.Current.Wallet.GetCurrencyNames();
```
hcorion: You can cache this variable outside of Update and get it during Start, we don't need to get this every Update().

Assets/Scripts/UI/InGameUI/MenuTop/CurrencyDisplay.cs
```diff
@@ -0,0 +1,34 @@
...
+        StringBuilder text = new StringBuilder();
```
hcorion: You declare a public variable with the same name at the start, rename one of them.

Assets/Scripts/UI/InGameUI/MenuTop/CurrencyDisplay.cs
```diff
@@ -0,0 +1,34 @@
...
+        // Populate the text box
+        foreach (string currency in currencies)
+        {
+            text.Append(currency + ":");
+            text.Append(World.Current.Wallet[currency].Balance);
+            text.AppendLine();
+        }
+
+        // Don't question it ok.
+        this.text.text = text.ToString();
```
hcorion: This is extremely confusing, instead use the public variable you declared. assign it at the start using GetComponent, then you don't have to use that fancy this.text.text weirdness.

hcorion:
Sorry, when I referenced that line in ProjectSettings/ProjectSettings.asset, I meant your changes to the entire file, not just that line.

buggy213:
roight

BraedonWooding:
Looks good, 😀.  Now I'm thinking maybe we should also have support for individual images instead of the text, but that would require specifying an image in the currencies xml/json file so I think a little out of scope?

hcorion:
It's a good jumping off point.

buggy213:
I think that's best left for a separate PR. This solution is functional and will work well enough right now, we can make it prettier later

BraedonWooding:
Yeh that's what I thought but felt still valid to bring up 😀.

buggy213:
It might be worth opening up an issue to request for make a request for artwork, although at this point it seems a little unnecessary.

koosemose:
I could swear this used to be displayed on screen, but disappeared at some point...

Anyways, how does this interact with the "quest" overlay? i.e. does it display over top of it or under it (the preferable option

buggy213:
Yeah, i made a PR for this a few months back but then i didn't have enough time to work on it and so it was closed. As of right now, the two currencies are entirely unobstructed by the quest overlay but even if we decide to add more, it does display under it.

BraedonWooding:
Wait @buggy213 just realised this now it seems that the performance HUD (the frames, memory, max/min and average) is weird?  Normally they should sit in a 2x2 area rather than a 3x2.  So maybe just see if you edited it and if not it may be for some bizarre reasons and you can just make the prefab a little longer (width wise) :D.

buggy213:
Huh. Alright, I'll do it tomorrow

BraedonWooding:
Note: no pressure if for some reason you can't figure it out then we'll just ignore it and I can fix it once its merged into master :D.

koosemose:
Let me know your progress on this, if you're going to be able to fix the issue @BraedonWooding pointed out or if you want to pass it on to him to fix.

BraedonWooding:
I think @koosemose just merge it and when I get time today I'll go and fix the issue since it seems to actually be a 'big monitor' thing, and I will go and fix some/all of the UI for big monitor (depending if we want to do it by segments; i.e. Performance HUD UI fix, Dialog Box UI Fix...).

bjubes:
@BraedonWooding  merging

PR #1727, opened 2016-12-29
Localize Settings Menu

bjubes:
part of #1513

Localizes most of the titles, headings, labels, and text elements in the new settings menu. 
I am still trying to find where the strings for sliders and buttons get set in code (if they are set in code rather than directly in a prefab). Any help finding them is appreciated.
After I have found where strings are set and passed them through the localization system, I will write key-values for each item in en_US.lang

### TODO 
**first feed into localization system:**
- [x]  labels (all but sliders)
- [x] subheadings
- [x] headings outside of scrollRect
- [x] sidebar buttons
- [x] sliders

**then:**
- [x] create keys for all the elements in en_us.lang
- [x]  revert changes in localizationTable for default and fallback enum (for testing purposes)

note: the bottom buttons are hardcoded, so they are not localized in this PR.


BraedonWooding:
Hey, sliders and all UI elements are stored in the settingsMenuFunctions.cs file in StreamingAssets/CSharp folder (path may be slightly off).  However don't worry about doing those since I'll do it when I add all the new options in the settings menu.  Also there is a text UI prefab that you can add the language localizer too which should automatically localize all elements (if the localizer works like I think it does). Also the entire system doesn't use prefabs (well like it doesnt revolve around them), however there are a few presets that you can mix and match to avoid having to write tedious code. Currently don't have access to a computer and won't till the 4th but that should be correct to the best of my memory.

bjubes:
By modifying all the text in the prefab and then seeing what is updated at runtime, I'm guessing that the bottom buttons on the menu (for saving, applying and such) as well as the static "Categories" heading is the only parts that are typed directly into a text field in the prefab. I have hooked up every other text element into the localization system apart from sliders.

BraedonWooding:
I'm a little fuzzy on whether or not they are set in the SettingsMenu prefab, or if they are set doing start.  I'm pretty sure they are set in the SettingsMenu prefab; can't check sorry no access to computer till the 4th.

BraedonWooding:
Also the sliders all use the text element so if you localize text you localize sliders :D.

bjubes:
when you say text element, are you refering to unity's Text object from UnityEngine.UI or a custom text element that then creates a Unity text object? I've been impelementing most of the localization in the classes in Assets/Scripts/UI/InGameUI/SettingsMenu. It seems like most of the classes in there create a Unity text object and then assign it via `newinstanceofText.text = "blah"` and I am changing it to `newinstanceofText.text = GetLocalization("blah")`

BraedonWooding:
Yep thats what I meant, that should localise it.

bjubes:
@BraedonWooding I have adapted the localization system so it can convert the names in settingstemplate.xml into localization keys (so "Developer Console" becomes "developer_console", thus allowing this pr to not mess with changing all the names into localizaiton keys) but even this isn't working because somewhere a ": " is being added to each label. Do you know where this happens? I haven't found it by using CTRL -F yet.

BraedonWooding:
@bjubes yep that's a good idea, it's added in the SettingsMenuFunctions.cs file under StreamingAssets/CSharp folder.  If you do a search for `": "` you'll find it, they are done on `CreateText(...)` calls, I'm quite happy for them to be removed if its going to mess with the localization?

bjubes:
Ill remove them in this PR, but add the colons directly into the localization so there is no difference to the player.

BraedonWooding:
Tbh I think that's even a better way to do it.

koosemose:
Everything looks good, merged

PR #1728, opened 2016-12-30
Optimize temperature

koosemose:
Temperature.cs has a few issues, this fixes some of them.

Temperature progression does not take delta time into account, so as best I can tell it functions as though 1 second has passed every update cycle (presently 1/5 second), this includes delta time in the calculations in the place that the comments imply it should be.

The two objects that update temperature both call the same function eventually, and furthermore I don't see a reason that things would do any different and heat things with some other calculations, this moves that function into C#, preventing a large number of heat sources from bogging the game down quite so bad.

The Temperature update event action for heat sources was being registered twice, once automatically via the xml, and again manually in the install event action. This removes that extra registration.

Finally, it adds a simple mouse over text to show the temperature in more commonly understood units (both fahrenheit and celsius)

Displaying the temperature in more understandable terms exposes some other issues preexisting with temperature, most particularly it heats tiles up insanely fast at the heat source but doesn't dissipate that heat to surrounding tiles nearly fast enough, so the heat source will end up at around 1000F, the next tile over at 256F and the next out will be around -256F. Even doing it on a faster timer (everyframe) it doesn't work in any vaguely realistic way (and bogs the game down horribly, as one might expect).

### DIFF
--- a/Assets/Scripts/Models/Area/Temperature.cs
+++ b/Assets/Scripts/Models/Area/Temperature.cs
@@ -44,9 +44,9 @@ public class Temperature
     /// <summary>
     /// Size of map.
     /// </summary>
-    private int sizeX = World.Current.Width;
-    private int sizeY = World.Current.Height;
-    private int sizeZ = World.Current.Depth;
+    private int sizeX;
+    private int sizeY;
+    private int sizeZ;
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


koosemose:
Merging this as the temperature issues and inefficiencies are causing issues in other areas of development.

PR #1730, opened 2016-12-30
[Fix] building utility network  -- reset updated this frame

koosemose:
Previously I removed updates frame Utilities on the premise that they did nothing. However, I overlooked on thing, and that is resetting the boolean that says rather or not they had updated their network that frame, to prevent them from getting stuck in an endless loop, crawling up and down the network of wires trying to build the network.

I still stand by removing the update, because it was way too much just for one little thing that is only used for one maintenance level thing and only when utilities are built. 

To reimplement the reset, I implemented a method in TimeManager, RunNextFrame, which simply takes an action with no parameters, and runs it on the next frame before anything other updates are processed.

### DIFF
--- a/Assets/Scripts/Models/Time/TimeManager.cs
+++ b/Assets/Scripts/Models/Time/TimeManager.cs
@@ -6,8 +6,10 @@
 // file LICENSE, which is part of this source code package, for details.
 // ====================================================
 #endregion
+
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using UnityEngine;
 
 public class TimeManager
@@ -15,6 +17,8 @@ public class TimeManager
     private static TimeManager instance;
 
     private float gameTickPerSecond = 5;
+
+    private List<Action> nextFrameActions = new List<Action>();
 
     // An array of possible time multipliers.
     private float[] possibleTimeScales = new float[6] { 0.1f, 0.5f, 1f, 2f, 4f, 8f };
@@ -130,6 +134,16 @@ public class TimeManager
         // Systems that update every frame.
         InvokeEvent(EveryFrame, time);
 
+        if (nextFrameActions.Count > 0)
+        {
+            for (int i = 0; i < nextFrameActions.Count; i++)
+            {
+                nextFrameActions[i].Invoke();
+            }
+
+            nextFrameActions.Clear();
+        }
+
         // Systems that update every frame not in Modal.
         if (GameController.Instance.IsModal == false)
         {
@@ -195,6 +209,11 @@ public class TimeManager
     public void Destroy()
     {
         instance = null;
+    }
+
+    public void RunNextFrame(Action action)
+    {
+        nextFrameActions.Add(action);
     }
 
     /// <summary>
--- a/Assets/Scripts/Models/Time/TimeManager.cs
+++ b/Assets/Scripts/Models/Time/TimeManager.cs
@@ -212,7 +212,7 @@ public class TimeManager
     }
 
     /// <summary>
-    /// Runs the next frame before any updates are ran.
+    /// Runs the action in the next frame before any updates are ran.
     /// </summary>
     /// <param name="action">Action with no parameters.</param>
     public void RunNextFrame(Action action)
--- a/Assets/Scripts/Models/Time/TimeManager.cs
+++ b/Assets/Scripts/Models/Time/TimeManager.cs
@@ -211,6 +211,10 @@ public class TimeManager
         instance = null;
     }
 
+    /// <summary>
+    /// Runs the next frame before any updates are ran.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Models/Time/TimeManager.cs
```diff
@@ -130,6 +134,16 @@ public void Update(float time)
         // Systems that update every frame.
         InvokeEvent(EveryFrame, time);
 
+        if (nextFrameActions.Count > 0)
+        {
+            for (int i = 0; i < nextFrameActions.Count; i++)
```
Dormanil: Personal preference or why do you use a iterative `for` if you could use a `foreach`?

Assets/Scripts/Models/Time/TimeManager.cs
```diff
@@ -198,6 +212,15 @@ public void Destroy()
     }
 
     /// <summary>
+    /// Runs the next frame before any updates are ran.
```
Dormanil: I _think_ it should be "are run".

Assets/Scripts/Models/Time/TimeManager.cs
koosemose: We've been having a lot of issues with garbage collection, and apparently the version of mono used by Unity creates a *lot* of garbage with foreach, so if I can use a for statement, rather than a foreach, I will (and in fact have been changing other areas of code to use for in preference of foreach when possible)

Assets/Scripts/Models/Time/TimeManager.cs
koosemose: .... I *think* you *may* be right... but they both sound just as *proper* when I say them, so I'm not entirely sure... 

And I'd also like to say I appreciate that even though I'm probably not going to change in this case (as I *think* both may be valid, and in either case am not certain which is correct), I appreciate that you've paid attention to the comments, since we've had issues with them being either inaccurate or just having terrible grammar/spelling.

Assets/Scripts/Models/Time/TimeManager.cs
Dormanil: Alright then, that explains that. I had much "fun" in Java in the last couple of months, where it's actually generally faster to do foreach-style `for` loops than iterative `for` loops. (Yes, they have the same keyword for both, I loathe it.)

Anyway, good to know that the old Mono seems to be bad in that regard. Maybe it will be better in Unity 5.5 with newer Mono version, but who knows.

Assets/Scripts/Models/Time/TimeManager.cs
Dormanil: Well, a quick Google search seems to suggest that "are ran" is never correct, but this is really not the worst thing in comments I have seen since I have returned, so I it's nothing I need to have fixed immediately. Both are understandable, but my bigger issue with that line would be more that I need an extra "in" to parse the sentence as you have it intended:

```csharp
/// <summary>
/// Runs in the next frame before any updates are run.
/// </summary>
``` 
Otherwise it could potentially be parsed as "runs _the next frame_" as in "we do all the updates of next frame _now_". Or maybe that's just my caffeinated, insomniac brain parsing English weirdly.

koosemose: Yeah that makes sense... I left that part as the autofilled summary tag had it... My brain was incorrectly applying it to the nextFrameActions field, rather than the function itself...

Assets/Scripts/Models/Time/TimeManager.cs
koosemose: Being newer to C# than Java, I prefer having the same keyword personally. And I used to use foreach at every opportunity, not fully believing that it could be *that* bad (and being lazy)... but then with my recent attempts at optimizing I've been trying everything I can think of, and seeing what has an effect... and was astounded at the difference changing a few of the larger foreach loops to for loops had in garbage collection

koosemose:
As this is primarily a bug fix, I'm going to go ahead and merge this now.

Assets/Scripts/Models/Time/TimeManager.cs
Dormanil: I personally simply like the verbosity and more natural feel of a `foreach (String param in params)` more than the a little more clunky `for (String param : params)`. And yes, I do realise that Java is shorter there. 

But enough of that tangent here.

bjubes:
@koosemose  the foreach loop extra GC has been fixed in Unity 5.5

> In versions of Unity prior to 5.5, a foreach loop iterating over anything other than an array generates garbage each time the loop terminates. This is due to boxing that happens behind the scenes. A System.Object is allocated on the heap when the loop begins and disposed of when the loop terminates. This problem was fixed in Unity 5.5.

https://unity3d.com/learn/tutorials/topics/performance-optimization/optimizing-garbage-collection-unity-games

PR #1731, opened 2016-12-30
[WIP] [Feedback Requested] Simpler updates with coroutines

NogginBops:
### Making updates run in a co-routine.

The PR moves update handling to a coroutine and reworks the update system to only update the furniture that needs updates.

This could be combined with #1721 to create a very efficient system for updates.

### What this PR does

- Removes the idea of visible and invisible furniture and updating them differently.
- Moves update handling to a co-routine.
  - This means that updates are not dispatched by frames but by time.

### TODO

- [ ] Stresstest the system to see if it holds up
- [ ] Test to see if any bugs has appeared
- [ ] Bring even more optimizations to the update system

Pings: @koosemose for potential merging with #1721

### DIFF
--- a/Assets/Scripts/Controllers/GameController.cs
+++ b/Assets/Scripts/Controllers/GameController.cs
@@ -9,6 +9,7 @@
 
 using ProjectPorcupine.Localization;
 using UnityEngine;
+using System.Collections;
 
 public class GameController : MonoBehaviour
 {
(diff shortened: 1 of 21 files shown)


dusho:
interesting solution.. something I had in mind when dealing with updates and time (except that coroutines I still don't quite understand and use in Unity)
though I do like chunking mechanism @BraedonWooding introduced in his PR ( #1695 ) when it comes to visibility and needed update frequency

NogginBops:
I will just say that updates being handled in a coroutine does not bring any performance in its current implementation, what really need to be done to get the most out of this system is to spread out updates. The fact that it's done in a coroutine just makes that easier. 

Currently all the performance gains seen in this PR is from the fact that we are not updating furniture that don't need updates.

koosemose:
Closing this as it's no longer applicable to the current codebase. I wouldn't mind seeing it applied to the current code to see what benefits it may bring if you get the time @NogginBops

PR #1736, opened 2017-01-03
Remove name and description from prototypes

### LINKED ISSUE #1608: Should item (furniture, utility, etc..) Name be removed?
I don't think the Name in the various Prototype classes actually serves a purpose, it basically functions as a US english translation of the type, and I suspect leads to confusion at times of which to use, type or name, and may also end up being used where a Localization should be used. Additionally there is the oddity and potential confusion of the Name property returning the Name set in xml (which may or may not match the localization for that item), and GetName() returning the Localization code.

I propose we fully remove Name from those prototype classes which have it and their xml, and anywhere in code that uses Name be changed to use Type.

koosemose:
Per discussion in #1608 this removes Name and Description from all objects (it's possible I missed something that has them, but this at least removes most), though Description wasn't discussed, it is similar to Name in that it should be localized, in that it duplicates what should be localized text, and only serves to create confusion and possibly conceal things that aren't localized properly.

Also, I'm reasonably sure I replaced all lua that referenced Name or Description with the appropriate use of either the appropriate localization code or type, as I felt appropriate to how it was being used, though again I may have missed some, but I *think* I got it all.

Closes #1608


Assets/Scripts/Models/Buildable/Components/Workshop.cs
```diff
@@ -364,7 +364,7 @@ private void HaulingJobForInputs(ProductionChain prodChain)
                                      false,
                                      false);
                         
-                        job.Description = string.Format("Hauling '{0}' to '{1}'", desiredInv, ParentFurniture.Name);
+                        job.Description = string.Format("Hauling '{0}' to '{1}'", desiredInv, ParentFurniture.GetName());
```
dusho: now this GetName() looks out of place while everywhere Type is used.. that method returns LocalizationCode (why is it a method, btw.)
from GetName() method I would expect it to return already localized name

koosemose: it's a method because that's what it was? I have no idea, this is honestly a case of something that needs ran through the localization system, and I'm not sure if it would be best for localization to just get the Unlocalized Name and it will automatically localize it, or if it needs to be passed in already Localized (with getName getting the Localization)

koosemose:
Merging to ensure no other uses of Name or Description slip into other PRs

PR #1737, opened 2017-01-04
Settings menu options

### LINKED ISSUE #1658: What settings options should we have?
This is to have a discussion on what options should be implemented, currently the settings menu supports multiple segmented 'categories' (like sound, video, general...) as well as 'headings' (like 'Advanced' / 'Basic' Video settings).  Below will be a table of current options (I'll update as people suggest).  These won't be implemented in my current PR however they will be implemented in another one (which may be made parallel or concurrently to the current one).

### General
Autosave | Localization
-----------|--------------
Interval | Language
Max Files | Auto-Update

### Sound
| Volume | Advanced |
| --------- | :-------: |
| Master | Sound Device
| Music | Locational Sound
| Game |
| Alerts |
| UI |

### Video
General | Advanced | Window
---------|------------|---------------
UI Skin  |                  | Mode
Quality  |                  | Resolution
V-Sync
. | Particles  
. | Shadows 
. | Buffering 
. | AA             

### Developer
General                    | Console              | Utilities
----------------------|-------------------|----------------------
Developer Mode     | Scroll Sensitivity | Performance HUD
Logging Level         | Time Stamps       |
. | Enable                 |
. | Font Size             |

BraedonWooding:
Closes #1658 

Just has all the options in #1658 and a few changes listed at the end.

#### Autosave
- [x] Interval
- [x] Max Files

#### Localization
- [x] Language
- [x] Auto-Update

#### Volume
- [x] Master
- [x] Music
- [x] Game
- [x] Alerts
- [x] UI

#### Advanced
- [x] Locational Sound
- [x] Sound Device

#### General
- [x] UI Skin - Currently a placeholder till someone builds a system for it
- [x] Quality
- [x] V-Sync

#### Advanced  
- [x] Particles  
- [x] Shadows 
- [x] Anistropic Filtering 
- [x] AA  

#### Window
- [x] Resolution
- [x] FullScreen

#### General
- [x] Logging Level - Implemented but needs behaviour.
- [x] Developer Mode

#### Console
- [x] Time Stamps
- [x] Scroll Sensitivity
- [x] Enable
- [x] Font Size 

#### Utilities
- [x] Performance HUD

## SettingsKeyHolder
Now one of the changes is to add something called SettingsKeyHolder, all it is a static class that holds all our default settings keys in a nice format all you do is call it like SettingsKeyHolder.DeveloperMode and it will return a boolean of the value, if you wish to change it you just call SettingsKeyHolder.DeveloperMode = value.  It is just a wrapper of the settings.getSettings, the reason?  Because I changed the naming scheme to reflect its position in the settingsMenu for example developer mode is located under Developer button under General so it's developer_general_developerMode and so on.  Now I know this may form some conjecture but the only places I found it used was where I used it to get the settings for things like PerformanceHUD and the DevConsole, and where it was used quite a bit for DeveloperModeToggle (which was already accessible through some static system for DevConsole commands, which this removes.  This removes CommandSettings.cs and replaces it with a more general solution).  The hard coded nature maybe the most contestable part of this but the whole point is so if we ever change the key name or we change where the key points we only have to change it in one place (instead of like 15).

## Checklist
- [x] First Stage: Get options in the data file
- [x] Second Stage: Get them in code
- [x] Third Stage: Fix up layout for the options (if any problems happen)
- [x] Uncommit any changes like ProjectSettings

## Should be merged after
#1727, #1647 

## Remarks
It's not too late btw to request or contest current suggested options (I got the list from the discussion issue I opened) just comment below :D.  Also this will need tons of testing to make sure each option is referenced right and they all work so when second stage is done if you have any free time pop on and try it :D.  

Also I fixed up the animations they look much better now :D, I will do a little more colour work since it's kinda irritating me lol.


BraedonWooding:
Note: the update fmod.cs was just to remove it as a commit (accidentally pressed ctrl+s with it open and it reformatted)

koosemose:
I can hook up the remaining sound stuff in a follow-up PR, as long as it's reasonably obvious where to do so and how to get the values I need. Probable a good idea for the positional sound to be optional, though I'm not absolutely sure there's a reasonable way to turn it on and off.. if all else fails, I can just have it play sounds at the listener's position.

As far as Sound Device, I'm not to sure on, I'll have to look back into the documentation to see if it's selectable, and if there's a way to get a pretty version of what the sound devices are. Alternatively, if I can just get it to update the sound device when default is changed, that would do fairly well (that's what most other games I've seen recently do), though that may actually be trickier than getting a pretty list of sound devices.

Also... what's up with your last commit? The commit message is... strange...

BraedonWooding:
Yeh they are easy its just SettingsKeyHolder.Master/Music/UI/Game... so on for each option (its a computed value, you can get it or set it, just a wrapped up settings.getsetting/setsetting so names can be verbose).  Next point have a look if not then it can be removed just a thought i had.  Also yeh idk what was up with that commit 😂 i was excited cause i got animations to work smoother for the buttons and for some reason it added it (it could be me accidentally copying it then later pasting it idk 😂). Oh and the update fmod was just to remove it as a commit (since i accidentally ctrl+s it when i was searching for things and it popped up).

BraedonWooding:
Last two commits to undo my previous commits to those files (sometimes easier when adding a bunch of new files since it means I can just do git add .)

koosemose:
Conflicted

bjubes:
You have a compiler error:
 ` ` `
Assets/Scripts/Models/Buildable/Furniture.cs(870,13): error CS0103: The name `CommandSettings' does not exist in the current context
 ` ` `

BraedonWooding:
@bjubes @koosemose there is no reason it should be erroring out... That line does not mention CommandSettings anymore, and all the unit tests succeed on the desktop variant.   So I'm confused to how it's gotten that error?

koosemose:
You need to pull in latest master. it's from SetUninstallJob, which I assume you haven't pulled master in since that was added, so while you don't conflict with master, master is still using the "old" form in an area that you haven't changed (since it doesn't exist for you).

In general if we (or automated testing) are getting an error you aren't, there's a good likelihood you need to pull in master, because something disagrees when combined.

BraedonWooding:
Oh okay, fair enough I'll pull in master and see if that fixes it :D.

koosemose:
Well... it won't fix it, but it will make what's broken actually be in your branch so you can fix it

BraedonWooding:
Yeh you were right it's from the uninstall stuff.  Okay now I know for next time too :p.

BraedonWooding:
Note: #1727 should be merged before this.  Also I still have to hook up sounds which I'll try to get to today or tomorrow (it's easy just have to slot it into busy schedule).

koosemose:
Conflicted

BraedonWooding:
Now I think is fine to merge, I do want to add a system (at one point) to the C# modding that doesn't require the get commands (basically instantiate object x) but that won't be anywhere near this current time frame due to other commitments.  I did test it and all the localization seems fine but maybe a final run through could be beneficial.

koosemose:
Something's awry, it claims developer mode is off, when it is obviously on (due to presence of spawn inventory ui):
image

Second... and this isn't *directly* related to this PR, just it is brought to light again by the large changes in Settings, but the console gets spammed with a lot of `Attempted to access a setting that was not loaded from Settings.json:` Which from looking at the code, I would call incorrect behavior, if a function has you pass in a default value (as this one does) it shouldn't be generating an error or warning when it has to return that default, as that would seem to be correct behavior.

BraedonWooding:
I will have a look at that, I did just merge in a bunch of localization changes and could have changed one of the keys by accident :P.  On the second note yeh that's not really related to this PR, now I agree with the fact that if you pass in a default you shouldn't get an error, but I feel that should be a separate PR since it's not relevant to this PR (I believe I don't even change a single line in settings in this one), I'm more using the settings 'api' (so to speak) in this.

koosemose:
It's also possible the first issue is related to having a settings file with "old" settings keys in it, in which case we need better behavior when settings are changed/added (Maybe rather than having a GetSettings with a default, we auto create settings with a default for any that don't exist, and if any get added we mark it for saving so the new settings are included in the settings file.

Also, I notice you didn't update the FallbackSettingsJson, it still has all the previous settings names in it. And perhaps the fallback settings shouldn't be implemented as a json string, and instead just directly set in some fashion, and those can be used as the defaults even if the file already exists (but the specific setting doesn't). Which would also fix the potential issue coming from different callings of GetSetting using a different default value, which could cause unusual behavior when a setting doesn't exist (one call having one default and another having a different), which may also be the cause of the behavior I'm seeing where checkbox says one thing, and existence of SpawnInventory UI says another (they may have different defaults, and since I don't have the new dev mode setting in my settings.json, it's doing two different things)

BraedonWooding:
The first thing yeh that's what the problem is.  And maybe it's more of a case of changing the message?  Saying "Couldn't find key {0}, so creating key using default value: {1}'.  And yes I can make it so it checks if the key exists and if it doesn't sets a default (when loading settings), that's easy and I'll do it in this PR.  I can also change the message in this one since it's very minor.  But I'm not so sure about adding new functions and methods that's not directly related (I do want to at some point add an instantiate object as I said before but that's out of the scope of this PR).

And yes I agree.  I'm wondering if it's better to do a Settings 'overhaul' where it's just changing up the behaviour of defaults rather than trying to incorporate it into this.  Though I guess since it's 'kinda' related and it's agreeable that it's better that way I can implement some kind of default value dictionary, instead of individual calls having defaults ?

koosemose:
Well... I do notice that SettingsTemplate has defaults listed in it, which I assume are what are used for the menu? It's not really clear... Is the fallback_json still used? There's an awful lot of places that adding a setting seem to need to touch, and really needs consolidated. 

Regarding the message, in my opinion it is better to remove it, as having a warning about normal behavior seems pointless... especially if that normal behavior can happen every frame and flood the logs (which is what happens for me), which means that either you turn that channel off completely (which is fine, as that's partially the purpose), but if you need to debug anything on that channel, anything else will be drowned out by something that seems to be treated as normal behavior. 

As for what should and shouldn't be done in a PR, I don't know. I usually will try to include small changes if they're directly related to the broader topic of the PR. At the very least I would say disable the message for now so it doesn't create log spam during the time between this PR being merged and a better fix being put in place.

BraedonWooding:
Yes the SettingsTemplate.xml defaults are what they say on the tin defaults for the menu, they were introduced since Settings didn't have them inbuilt and I needed them for default values for elements.  Now on the second point it's not very clear and I didn't implement it but from what I can see it seems to happen if it can't find the file or some loading problem occurs.  Now I think that removing this and rather just populating the internal dictionary with default values is a 'better' way.

But (always a but lol), I don't 100% think that the settings.cs class should rely on the settingsMenu for defaults.  But I do hate adding new xml for small things like defaults or by having it navigate the same options, I guess the only problem I can see is if SettingsMenu hasn't yet initialized with its data and done defaults then problems occur and so what should happen?  Cause it could occur on every load due to something firing off early?

Now I don't know if you meant consolidation as in the Settings class or as in the SettingsMenu and I agree with the vibe that the Settings class needs a lot of consolidation, this was one of the reasons why I built the SettingsKeyHolder as an interface to the Settings, this way if we changed a key name (for whatever reason) we only have to change it once internally (that's presuming we remove the fallback json).  But again I think that's out of scope.

However, if you meant SettingsMenu I do feel it's fairly consolidated there is just:
- name for display which is really just a way so you don't have to create a string literal of the name in the code and means that in some cases you won't have to create any custom class at all :D.
- key name for access
- default value
- and the class name for custom use
It's also completely internal and doesn't touch much (other than the UI systems and the Settings system).

However, in writing this I do have an idea for an improved settings system (just an idea rn but it's actually decently relevant to this PR).  The general idea is this:
- SettingsTemplate remove Key and Default value and has an object called Key, just for cleaning up
- Key has both KeyName and DefaultValue fields
- Key also has ValueType (int, bool... I already have the code to get this with support for UnityEngine and other assemblies from my DevConsole so we can use that)
- We have a LUA document with enums/flags (so it can be modified, and we can concatenate them) therefore one can just tell the system it wants to know the value of x or whatever.  Now still this requires strings (stringifying things is often called 'the root of all future bugs') but then again we could create a C# enum (well tbh it would probably just be a class) that would prevent bugs on our end.  And since LUA is translated (and 'technically' one could see C# modding as real-time compilation) it means that those could access a 'dynamic' enum.

koosemose:
I was referring essentially to the whole thing, for example SettingsMenu shouldn't *really* be handling the defaults, that should be the settings themselves. Which also makes it rather misleading, the defaults aren't *really* defaults since they're basically only affecting what the menu displays not what the settings actually are (or what the specific call chooses to have as a default). 

I will warn you on your planning though, be careful of overcomplicating any mod-facing stuff, particularly things that may be commonly needed by mods (such as settings), even if the simpler solution may not be perfect. And also watch for requiring C# mods for "basic" things. I'm not a fan of anything requiring C# for modding, but I do realize some things may require it (particularly if it needs to play with classes and objects), but it should only be done when it is really required.

BraedonWooding:
That's exactly what I was meaning, but I was also saying that it's kind of one of those flip flop no good situation things.  Where in one case you have an xml just for defaults and it means if I want to add a key I have to add it in two xml documents or two we have just a single xml and well it's used for the SettingsMenu but also used for the default settings.  Kind of why I called it SettingsTemplate it represents a template of all the settings options that are used to formulate the SettingsMenu but may also be used for defaults.  Now I'm not entirely against adding the extra xml but I do think it could be unnecessary and cause more problems??

Now I will admit my bias against LUA (to me it fails in its core mission but that's for another discussion lol), but I'm working to allow all C# classes to also be created in LUA for all my stuff, this will include the SettingsMenu, the PerformanceHUD and anything else that uses the malleable base UI system.  For the flags if it was a C# enum/class it wouldn't be external, you would add the options either in a lua file or via a C# command or xml file or whatever right it wouldn't just be limited in this case. 

Now despite this I will point out that using LUA for the UI stuff isn't recommended since well Unity is coded in C# (well actually a lot of it is C++ but that's beside the point) so interacting at a gameobject level is harder in LUA (though there are the helper functions too).  But for things like accessing settings and stuff will be universally possible and usable (easily) across all languages supported.

I'm still up in arms a lot about the idea of the enum thing since well strings have problems but maybe we just have core options being implemented hard coded (like `SettingsKeyHolder.EnableConsole => return Settings...GetSetting("Developer_Console_Enable")`) and the rest are up to the mod author to either add a static class similar (or perhaps to extend from it or similar) since well our code will never call their stuff anyway so really it's kind of 'not our problem' since it's really a style thing.  If you add a single key and you reference it once or twice just by string is fine???  So basically TLDR on the last point maybe we just keep it like it is for the whole accessing options, any LUA/C# script can access the key holder (I'm pretty 100% sure) and they can either implement a similar solution for their options (for maybe other modders to use) or just call by string.

koosemose:
It kind of makes sense for me to get the contents of the settings menu from settings (though in our case, your SettingsTemplate is more robust than the Settings, so it makes sense to use that).

And interacting with UI on the level that mods need isn't that difficult, things just have to be coded for it on our side (basically what ModDialog does) to ease the transition.

BraedonWooding:
On the second point I agree, and that's exactly what 'my' UI system does too (ease the transition) it adds functions such as createText() or whatever that rely on chaining of layout groups rather than pixel points to enable precision of display :D.  I was more talking about really deep stuff like disabling other elements and other stuff that is only possible with a C# script.

On the first point so what's I guess our ending point?  Should I just leave it as it is (the template) and remove the fallback json (and when it's creating the settings originally, it will go and take the defaults from the mod manager instead of the settings menu setting them) and also remove that log?  That to me just seems simpler?

koosemose:
Remove fallback Json (just be sure things are still handled properly when there is no settings.json). As far as future, when mods are involved, I suppose it won't be so much a matter of getting the settings from the mod manager, as the mod manager getting the data from file, and setting things where they need to be set... which may be what you meant just said the other way around...

BraedonWooding:
Yeh that's what I meant.  I'm aiming to get my list of things done by end of weekend since I have time

BraedonWooding:
Currently the logic is as follows:
- No fallback json, it loads in the template first to set defaults then the other file goes ontop
- Therefore if the key can't be found it doesn't exist at all, anywhere...
- The function now returns a boolean to whether or not the key exists (this includes both the settings template and your settings file), you now use an 'out' variable to get the value.  This does make it a little messy (requiring a 'temp' variable in some cases) but it when Unity updates to a newer version of mono and we can create out variables inside the out statement it will remove that messiness.
- If the function can't find it, it will set it to the default(T), you really should use the boolean to indicate whether or not you should use it and you can do most calls in a line or two so it's simple.
- SettingsKeyHolder has been updated a little bit and works just the same as before (access the properties instead of having to use strings that could change) but now will also return a proper default value if the key can't be found (instead of default(T)) this just means if we change either the default value in the template or the key we also have to change it in settingsKeyHolder, which is only 1 place instead of everywhere (as you can see significantly cleaner).

Other than that there was just some general nicer code :D, this is a much, much more consistent system since the old one could allow different default values for the same key this keeps it completely consistent :D, we should never run into a problem where the key doesn't exist on our end (else if they - the modders - remove that key in which case we have a literal default as a backup to prevent any weird stuff) :D.

koosemose:
If I understand correctly, as you have it now if a key doesn't exist that you're trying to access, it should be a legitimate error; if that's the case then perhaps we should just allow it to fail as such and throw an error when someone tries to access a nonexistent key, and leave it to the developer or modder to fix in their code, rather than having to use temporary variables for an out param?

Do you know what happens if you load an old settings file that contains some old settings keys and doesn't have newer ones? Does it still work properly in setting defaults?

BraedonWooding:
Oooh yes that's a fair idea, I'll add back in the error.  It will not migrate over the new settings but it won't cause any problems they will just kind of sit there and collect dust in the dictionary this is mainly so u can disable mods and re-enable them and the settings won't be gone.

koosemose:
I was mostly worried about the possibility of the presence of an old settings file somehow getting in the way of loading the new defaults. Of course if there is an issue, I'll see it as soon as I get to testing, since I'll have old settings, but thought I'd mention it now just in case since you seem to be currently involved in your code.

BraedonWooding:
It shouldn't cause any problems.  Though to be fair you never know 😂

koosemose:
Everything looks good, just need that error added back in, since nonexistence of a settings key is now an error rather than just how it handles falling back on the default, then it should be good for merging.

BraedonWooding:
Yep should be all done now, it is now an error since it never should happen (unless the user hasn't done template right or uses the wrong string).

koosemose:
Alright, everything seems good... any further tweaks we can do in followup PRs. Merged

PR #1738, opened 2017-01-04
Remove FeatHub link from Readme

koosemose:
Feathub is barely used and we have little control and can't police it (such as removing the "memes" suggestion), or remove already implemented features.


bjubes:
I agree, features should be proposed on GitHub for now. Although I think the upvote system is useful for community driven development, we need to use platforms we have control over. simple thumbs up/down has worked so far and consolidates where PP is

PR #1739, opened 2017-01-05
Install/Uninstall furniture

koosemose:
Adds context menu to uninstall furniture with an Uninstall OrderAction, along with an accompanying Install on the matching inventory.

I wanted to generalize having non-Inventory as inventory, but each potential thing that might want to exist as inventory (only furniture at present, and gear in the future) has such drastically different needs in how it needs to operate and look that I'm not sure that's possible

Still need to sort out the localization particularly for uninstall, as it is based on deconstruct which seems suboptimal (formed dynamically, requiring a key for each deconstruct action.)


dusho:
so this creates packed (crated) furniture in order to move it? What is the gameplay usage for this?

koosemose:
Basically the same as it is in Rimworld, not making it a pain and having to lose materials if you misplace an expensive furniture. Additionally, it means you could order furniture/ buy from a trader. Or rather than having a completely built and set up base, players could get just the frame (walls and floors) and a basic set of furniture, without needing to be told exactly what they need to build from an assortment of resources (because they just have everything already there, just in boxes). And it also allows furniture to be rewards in the quest system. But the first reason was my primary reason for doing it.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -795,6 +795,16 @@ public void ReadXmlPrototype(XmlReader readerParent)
...
+
+        if (orderActions.ContainsKey("Uninstall"))
+        {
+            InventoryCommon asInventory = new InventoryCommon();
+            asInventory.type = Type;
+            asInventory.maxStackSize = 1;
+            asInventory.basePrice = 0f;
+            asInventory.category = "crated_furniture";
+            PrototypeManager.Inventory.Add(asInventory);
+        }
```
dusho: what does this do? If it's needed can't it be placed in UnInstall.Initialize() ?

dusho:
ah.. ok. Guess I haven't played Rimworld for some time now.
Is UnInstall element in furniture used also for Install ?

Assets/Scripts/Models/Buildable/Furniture.cs
koosemose: This is what actually creates the Inventory prototype (and sets some baseline info). It could *probably* exist in Uninstall.Initialize, but wasn't put there simply because the Uninstall orderAction didn't exist at the time I made this piece of code.

The one potential sticking point I see is basePrice, I want to have an actual base price set in xml (rather than always being 0), and it feels unintuitive to have it exist in the uninstall tag. I'm not entirely sure how I could get that info to the Uninstall orderAction without having it a property in the orderAction tag, though I will look into seeing what I can do on that front.

koosemose:
Install just uses the standard building method, just overriding the required items. While I can see the sense of either somehow using the Uninstall action, or having an Install action, that would necessitate having an entirely separate buildMode for the buildModeController (I think), which would be nearly identical to the furniture build one, so a lot of duplicated code. Aside from that, I don't like the idea of requiring a separate Install OrderAction (because it makes no sense to have one without the other), and I'm not sure how/if the Uninstall could be used.

And I'm personally quite fond of the quality of life aspects of Install/Uninstall, having switched from games that have it something equivalent to ones that don't, it's amazing how much more frustrating it is to have to fully deconstruct something because, either you put it in the wrong spot or later need to rearrange. 

Though it could potentially be an interesting gameplay aspect to have at least some things that can be uninstalled require skill to do so (when conceptually you are actually partially disassembling it to transport it, rather than just picking it up and moving it).

Though now bringing up the potential of skills playing into success and failure of things, if eventually there's a chance of building something failing ( losing *some* resources ), that would be a good argument for having a separate install action, rather it requires being added to the xml, or automatically created when you have an uninstall  orderaction, because even if you have the same failure chance, at worst you'd probably want to, rather than just potentially losing the entire crated furniture, having it either be installed broken, or just add some portion of the build order required resources to represent it being slightly broken... But I'm getting ahead of myself.

dusho:
Just thinking.. if we want to introduce e.g. some explosives with timers and stuff - that would be inventory as well, right. Shouldn't we add some kind of components system to inventory as well? Otherwise we will keep doing what we did to Furniture class - adding more and more crazy flags, more functionality into it and stuff not working that well together. If Inventory would have Params with dynamic things (as Furniture does) that would keep info about crated furniture (or actual time to explode, or expiration time of food).
In this case then we would have inventory component - say Unpackable or Installable.

koosemose:
In this case, I don't think a component would quite do it, simply because as far as the data (xml files) is concerned, these inventory don't exist, and I *really* don't like the idea of having to explicitly define all furniture that can be uninstalled separately in Inventory.xml.

There's however probably a lot that could benefit from having Params though (that *might* be a place that info about the machine uninstalled could be stored, such as a damaged machine staying damaged if uninstalled and reinstalled).

I would also at least prefer to highly discourage inventory having updates if at all possible, simple because it's so much easier for a base to end up with a lot of inventory lying around, and where a user with a weak computer can build to account for that (trying to keep furniture to only what is absolutely needed) keeping inventory under control will be a lot harder.

dusho:
Ye.. no list of furnitures in Inventory.xml. I would just introduce tag 'Installable' to furniture for all that can be packaged. (This would mean than that packaging time will be same for all - but no need to have UnInstall tags for furniture)
Regarding updates - I already saw some PR with Components having IsSlowUpdateRequired and IsFastUpdateRequired.. that can be nice generic way to call update only on stuff that needs it - so Inventory without component -> no need for any update.

koosemose:
I much prefer being able to have variable times for un/install, since you were the one that coded orderActions, and components both, can you think of a way with either, or by combining them (with only one entry in the xml, such as maybe a component that creates the orderActions) to have un/install having variable times (even better if install doesn't rely on construction time, as it should in theory be easier to build a thing from dissassembled components/a boxed version of that thing than it is to construct it from scratch)

dusho:
Hm.. not sure. Also what would be the benefit of having variable install times? Gameplay wise..
I would much prefer having mechanic you mentioned here before - having just some base time for install/uninstall and then factor in skill of the guy. Maybe basetime can be calculated from size of the furniture (so 2x2, 3x2, ... )?

koosemose:
First off, sense of believability (not necessarily realism), for example, let's say I have a simple wooden chair and a fission reactor, assuming both are un/reinstallable, I highly doubt there is any time that won't feel jarring to immersion on one or the other (either the chair takes way too long, or the reactor is way too quick, or both are somewhat too slow and too quick respectively), basically the same purpose that variable time on con/destruction serves (even though we don't really make proper use of that at the moment, as we've not given any attention to specific numbers and are currently focusing on the bare mechanics).

It can also serve for balance. For example, let's say we have defensive turrets that can be set up. We either have short un/install times, in which case it's going to be annoying to do a simple rearrangement (of chairs for example) or we have short times and players can easily game the system by having weaker over all defenses and move defenses to where they're needed very quickly (not to say that there might not be quick to un/install turrets specifically for this purpose), and/or if we were to implement something like Rimworld's feature where the difficulty of attacks is based on your base's strength (number of people and defensive structures), the could keep base-line defenses just in case they're slow to respond and have a stock of crated turrets to set up when an attack happens, meaning they have a lower defensive rating when the attack is determined (so a weaker attack) but a strong defense when the attack shows up.

dusho:
ok, but then why not to use some fraction of build/deconstruct times we already specified. If believability is the factor then surely there is relation between furniture being built from scratch and installed from the parts. What if installing/uninstalling is purely 30-50% of time it takes to build/deconstruct. Further influenced by relevant skills.
Alternatively if you really want explicit install times, then we can keep still way you defined it here (so instead of furniture having tag to identify if it's installable, have just existence of Install component), just additionally introduce components for inventory - with Installable having in Params furniture that has Install component. Install times for furniture would be taken definition in Furtniture.xml.

koosemose:
For now I'm going to keep it as is, as usual, if a majority of others are against it we can change it. But I'll look into further modifications in how things are handled, I will see about adding a separate Install OrderAction for the time being, until we can figure out how/if it's feasible to do an OrderAction from the component, just so that Install/Uninstall data can be set from within a single tag.

koosemose:
On second thought I'm going to stick with the single OrderAction for now, things seem to get overly convoluted doing other ways.

dusho:
as you wish..
I still think clean component system for Inventory would make this feature cleaner and more flexible (control over parameters). Surely we can make OrderActions work with InventoryComponents - even loosely through names of components (strings) so there can be new components loosely added.

koosemose:
Yeah, I'm not saying no to the whole component thing, but while that *may* be able to support this, it is also vastly different from the main focus of this PR. I do think Parameters at least will be required for full functionality in the future, but at present there isn't any sort of constant state to furniture outside of things that it would make sense to not be carried through un/reinstall.

PR #1742, opened 2017-01-05
New asteroid generation

koosemose:
This is mostly just a demonstration of my work on generating asteroids in a different way, rather than the current Perlin Noise based method (which is extremely unsatisfactory for 3d asteroids), it picks random seed points along the map and "grows" the asteroids from those points, generating roughly roundish objects, with the occasional odd shaped one as multiple asteroids overlap.

AsteroidSize (average) is defined as radius in the appropriate worldGenerator file, with a density equaling an approximate percentage of the map covered by asteroids (assuming asteroids are perfectly rounded ellipsoids, evenly spaced, and packed "optimally")

Asteroids seeds are spread using an approximation of farthest-first traversal, while this (with the number of options available [4x the number of asteroids desired]) doesn't absolutely guarantee asteroids not overlapping, it does tend to minimize it, and it seems rare enough that the times it does happen is desirable, as it will lead to an occasional peanut shaped asteroid.

While the spreading algorithm could be improved (though not too much, occasional overlaps are desirable, and too evenly spread will lead to too much uniformity), it is functional and is separated enough from the rest of the code that it *should* be trivial to replace it without drastically altering the rest of the system.

The growing of the asteroids could potentially be improved, I can't come up with a better solution, and this version does generate better looking asteroids with a bit of culling (bits that stick out too far -- i.e. don't have enough neighbors -- aren't placed).

This can also (as @dusho mentioned) potentially lead to such things as asteroids being more consistent in ores (i.e a single asteroid has only 1 or 2 ore types), and possibly placing the ores in vein like structures, since, during generation, each asteroid is aware of the tiles that it consists of. However, I consider that work for another PR.


dusho:
this sounds nice.. with this we can have asteroids only with specific range of ores.. and having fields with mostly iron ore and such
you can extract those hardcoded parameters into .xml where amounts are defined.. or have different .xml that would represent worldTemplate (world_asteroids_iron.xml, world_asteroids_mix.xml, ... )

koosemose:
I had considered having asteroids be more consistent with their ore, but the one issue, is, at least at PP's current state and possibly for the foreseeable future, there's entirely too much likelihood of generating a world that is completely missing a given resource, and while I'm usually fine not limiting features because they require things that our current version doesn't have to be properly playable, multiple worlds/areas seems like a much more distant target.

Though I can think of some potential ways to have more consistent asteroid resources, but still ensure there's at least some of every resource (the first thing that comes to mind is generate a list of length matching the number of asteroids generated, and each slot in the list get's a list of what resources that asteroids contain (tending towards one, maybe two), and then before actually generating the asteroids and their resources, squashing the LoL, and ensuring every resource is represented at least once).

dusho:
just now see the asteroids.. it looks a bit messy..
this PR needed screen

PR #1743, opened 2017-01-06
[Conflicts] TileByTile Temperature System

BraedonWooding:
Here is my proposed system for replacing the old temperature system.  @abackwood is building the other one, since we want to get the best system and we had different ideas.  Mine using a tile by tile tracking system (with a handmade equation to simplify and allow it), where as his will use a room by room tracking system.  Image of mine below in effect (let it run for a short time - a few mins):

image

If a furniture has heat requirements not met (not provide requirements only receive ones, so if it reaches a max/min it won't output this requirement :D, just wanted to throw it together to show what happens if it's not met!)
image

This completely replaces the temperature class with basically a 100% better system, it doesn't have internal maps of things and rather replaces it with a cache (this is very important, I'll talk about this in a moment), and essentially sitting ontop of the world class and getting its data from that.  This also needs 0 (yes 0) updates on real time (it doesn't have any 'fixed freq.'/every frame update functions or whatever), rather you tell the system where you want to dump some heat and how much heat (and you pass a delta time, but if you just want to dump a one off then just make delta time 1), then it will process that independently of time.  This means technically it can run in a thread (though it does require access to the world class so it's probably recommended against). 

Okay I mentioned caching but I don't mean caching instances I mean caching heat maps basically, so the system uses generalisations to gain speed (generalisations is a form of coding, where you... well generalise something; i.e. say in 99% of situations this is true, or in this situation this is true IF this is true ELSE this is true...).  So the first generalisation it makes is with the formula (more on that in a moment) the second is with how it handles heat dispersion, the term I coined for this is majoritively circular sub polygonal dispersion generalisation.  Essentially it's a generalisation that the area that will be expanding will be circular majoritively (by circular it means 1x1, 3x3, 5x5, aka increasing by 2 every iteration; or one on each side of the square), but some of those squares won't be absorbing that default energy and rather will be absorbing a different value (because they have a different thermal conductivity index) so therefore it's 'sub' polygonal (like a subtree, it's mainly circular but it is also polygonal), then the dispersion bit is because its dispersing heat.  So why is this useful well we can cache the results of the energy loss (till it reaches < 0.1) and never have to redo those results (never ever, even with new maps, not currently a static but while writing this realised it should be), so you're already saving having to run this formula on the majority of your tiles and only have to run it on the tiles that don't follow this standard conductivity value (which is the value of air).

The formula is essentially a two part formula the first one is where it works out how much energy is lost per m (based off either the general thermal conductivity index, or the specific index if its a non conforming square) then the next formula just reduces this depending on the current temperature (currently it makes it a little too hard to go from < 0 to > 0 and too easy to increase above 0, though this is just about fine tuning).  Note: the tiles which trigger the polygonal method also make sure to effect their neighbours (resulting in an energy loss), though this has to be a little more developed since it doesn't carry on as nicely.  Also there is a component for temperature now, this component will output the heat you set it as, though this is changeable as it's a parameter :D (in-case you are outputting consistent heat with its value depending on something else), else if you want to just dump some random heat (or non-random) just call Temperature (you will need to go through a world) then you have 3 choices:
- ProduceTemperatureAtFurniture
- ProduceTemperatureAtLocation (uses a vector 3)
- ProduceTemperatureAtTile (uses a tile)

All furnitures now only need a single thing to implement temperature (the temperature component) rather than having to have a base_heating param along updates.   Oh and I removed sinks/sources no need for that you just tell it when you want heat and it complies.  I also added a good temperature unit class so you can get the celsius/fahrenheit/kelvin value from a single class (all tiles have one of these unit classes) instead of having to do a function to convert...  

Uhhh nothing else really, It's like ready to be convert to use multi-maps instead of 1 (well it already will work, since when you create it with a world it grabs a copy of that world), so that's good I guess (though that mostly came from good coding practices).  There are literally tons of useless commands (and useless xml data) that just sits there currently, which I haven't yet removed but yeh (so note: if you find a function that looks odd it's probably not used and going to be removed).  The system runs from kelvin and that's basically it :D, enjoy playing around with it.

Checklist:
- [x] Add equalisation
- [x] Make polygonal calculations more far reaching (continue for a little longer)
- [x] Add 3D (there is already 'support' just needs to use the support)
- [x] Fine tune
- [x] Add average room temperature
- [x] Add image for things that don't have the heat required
- [x] Remove changes to asset files

Ways to improve (either in the future or if it's deemed needed now):
- Cache smarter and more accurate maps
- Constantly fine tune the k value
- Improve the tile retrieval
- Put in a thread by itself, since all of its work is just touching functions, could store all changes wanted, then just carry those changes out :D.


BraedonWooding:
Just want to add a comment note that it's a little square right now due to neighbour system being a little cranky with me :P, it'll handle it better in the future.  Also haven't added equalisation yet (so that's why it can seem to not really equalise properly with rooms).  Anyway that's why it's called a demo!

koosemose:
The rate of spread looks pretty good (though perhaps we should have the base start off a little bit warmer than absolute zero, but that's unrelated to this. Assuming things are fixed so that heat isn't leaking into space at such a high rate, this just might work out.

Also, I'm assuming you haven't presently reimplimented RTG's heat generation? If you have, then somethings gone wrong because they aren't.

koosemose:
Also, just to test the functionality on the flip-side of things, would you care to do a quick heatsink type furniture (don't care about the sprite side of things... just a white box that says heat sink will do for now)

BraedonWooding:
Sure ill raise the default to like may 0 degree C / -273.15 degree K. No havent done rtg but will do now 😀.  What do you mean by sink?  Do they take in heat ?  In that case does a cooler fit the bill or are you talking about a literal sink like as in it absorbs its surrounding heat?

koosemose:
In theory either... I'm not sure they need to mechanically work differently... either a cooler (presume effectively it cools things off at the cost of power, or some kind of radiating fin thing (to radiate excess heat into space), or pretty much any method to lower temperature.

Any way to get rid of excess heat... because that will end up actually being the more important gameplay element, honestly (once we get past 0k base, anyways)

BraedonWooding:
Okay got it will do that and fix the other things today 😀

abackwood:
So when you say that you have no updates, what do you mean? When do you run your calculations and how heavy are those?

Assets/Scripts/Models/Buildable/Components/TemperatureConnection.cs
```diff
@@ -0,0 +1,149 @@
...
+            get
+            {
+                return FurnitureParams[ParamsDefinitions.OutputRate.ParameterName].ToFloat();
+            }
+
+            set
+            {
+                FurnitureParams[ParamsDefinitions.OutputRate.ParameterName].SetValue(value);
+            }
+        }
```
dusho: Does OutputRate needs to be dynamic? Will it be set to different value? If not, you can have just float property OutputRare in here - then in .xml it will be defined inside Component tag.
Or if you want to keep it as furniture param, you can still define if as property, and then during initialization set it though this OutputRate setter in initialization.
This is just to have input related to component have defined inside that component definition..

BraedonWooding:
Dynamic.  I.e rtgs at one point will base it off their energy output.

BraedonWooding:
Was basing code of other components, and its rather that its optional which is what i felt the furniture params was used for?  If the second way is better then I'm happy to change it to that.

koosemose:
Yeah, having looked over the code, I think the way it's actually supposed to work with the ParameterDefs, is that the component defines what the parameter is named, but it's not actually set in the Params, but rather from a Property in the component xml itself.

dusho:
yes, as @koosemose said.. 
also explicit definition in .xml is not needed unless some other components will use that value (but generally for learning and clarity purpose it's cool to have it there)
If it's dynamic, you can still have property in component called say InitialOutputRate that will populate that param in Initialize()

BraedonWooding:
Okay so just to clarify I should have another property of the component called something like 'InitialOutputRate' and that sets the furniture param output rate in initialize so other scripts can change it when they need to?

dusho:
yes.. and don't have to set OutputRate in Params in furniture.xml

BraedonWooding:
Yep got that but okay sweet will implement that 😀. Sorry just learning the component system now really.

koosemose:
Something you are doing keeps squashing the furniture xml file down, removing all blank lines, which drastically reduces the readability of the file, and causing conflicts with any changes to furniture.xml

BraedonWooding:
Oh sorry, it's cause I'm editing it using a xml editor (which is obviously making it squishy).  I will use a prettifier to make it more readable :D

koosemose:
Just marking this as WIP as you said elsewhere for it not to be merged yet... just remove when ready.

BraedonWooding:
@koosemose okay so I'm pretty happy with the balancing, and it diffuses a nice amount through walls, smaller rooms will heat up quicker (since less equalisation will occur) and generally the hotter something is the more it resists heat (among other little tidbits that are simulated quite well IMO).  While I feel there are some sections of the code (like equalisation getting all its neighbours every time it runs and uses delta time, where it could equalise maybe once a second or so) but often those would require some code changes elsewhere - mainly TimeManager (like temperature really could benefit by being on a slower timer that updates less often since it doesn't need to have 4-5 updates a second could easily work on 2-3).  Though I think this is a case of letting it run and see what people think (if people think that x material lets too much heat in then we tweak it a little, or that heat diffuses too far then manipulate k...) and so on.  So I would say merge it (if it passes checks and so on), and we can go from there?

koosemose:
Yeah, I wanna give it a run at some tests, and want to test out a few stress test scenarios and look for any edge cases to be sure it functions how we might expect in those cases.

Assets/Scripts/Models/Buildable/Components/TemperatureConnection.cs
```diff
@@ -0,0 +1,188 @@
...
+                }
+
+                // Else we are fine :D
+                World.Current.temperature.ProduceTemperatureAtFurniture(ParentFurniture, OutputRate, deltaTime);
+            }
+        }
+
+        protected override void Initialize()
+        {
+            componentRequirements = Requirements.Temperature;
```
dusho: this should be maybe set only if Requires != null and has some limit it needs? e.g. heater doesn't need input temperature and I guess you don't want to show temperature status icon

BraedonWooding: Will it matter...?  Because it won't ever tell the system to display the icon unless `Requires != null`.  I was just following the pattern of every other component where they would set that requirement, though this may be  a bridge case?

BraedonWooding:
@koosemose please do.  A good edge case is placing the heater two spots East of the door (just the standard map), one of the tiles won't heat up, this is just due to it not reaching that tile due to its weird location, and it not being properly dispersed into.  Now I would claim that in reality it's just a weird cold spot, though I could try to improve the equalisation stuff to try to equalise better against this tile?

koosemose:
That is indeed a peculiar case, I'll be sure to check it out to see if it's a significant enough difference to cause issues in the future.

Assets/Scripts/Models/Buildable/Components/TemperatureConnection.cs
dusho: ah.. you're right. CanFunction() has to return false in order to get notification and there already is check for Required.

crazyfox55:
Have you considered doing equalization on the GPU with a Shader? It should be simple in 2D it is similar to a blur operation. Though I haven't fully mapped out how to deal with 3D.

BraedonWooding:
The colours are only a representation of the values of the tile's temperature, so each tile stores its own temperature.  Now this doesn't exclude using a GPU but I think avoiding complicating the equalisation of a tile's temperature by spreading its values out to its neighbours would be more beneficial.  3D wise would also overcomplicate it way too much since in reality the only 'problem' with the system is the fact it has to generate a list of its neighbours every 'run' (causing GC, which is only a problem since we already have problems with GC so any extra GC causes further problems - regardless of the amount).  Though you saying that does spark an idea to essentially offload countless operations to a GPU to speed up, though I really doubt it will gain much friction in PP, since I want all my code to be simple enough for anyone to read and understand it (incase a bug pops up and I'm not here to fix it), and I think GPU would just overcomplicate it.  Though I can definitely see some benefits for GPU in PP.

Assets/Scripts/Models/Buildable/Components/TemperatureConnection.cs
BraedonWooding: Yeh, though this probably raises the question of whether or not we should even call CanFunction() if its only for showing that indicator?

crazyfox55:
Now that you mention it moving to the GPU could be problematic for the target coding level of PP and maintainability.

crazyfox55:
In tile you have equalize temperature and it modifies both the current tile and it's neighbor directly when the function is called. This will cause inconsistent temperature flow based on the order of neighbors. It will also cause inconsistencies based on the order in which each tile has equalize temperature called. Instead changes should be stored in a buffer and applied after all tiles have been calculated.

BraedonWooding:
Yeh that was one of my main issues with the GPU stuff.  Neighbours are always consistent (its always N,E,S,W, then the diagonals and so on) so I use this to influence it (meaning it spreads less to the diagonals and verticals).  However, I did take what you said onboard and now it sorts it based off value (so the coldest tiles will heat up sooner) this was mainly just so things will be more equal.  The idea of a buffer though is a good idea and made me think.  @abackwood what if we actually used this to maybe combine our ideas, so we could essentially store a buffer of like a room state then perhaps do an average of that room state (not direct average but rather like essentially using a room formula) to get essentially your room value, in reality we wouldn't get any/much slow down and wouldn't need equalisation, and wouldn't need update functions?  Also this would solve the idea of updating rooms since temperature would essentially work everywhere (obviously we can disable it on tiles if there isn't enough gas/no furniture).

koosemose:
Conflicted

BraedonWooding:
Now I'm perfectly happy to update it to resolve conflicts (and spend time building it to equalise within a room) but I do feel that the room by room is better (I do think at one point I suggested we do room by room for the reasons that I feel tile by tile is weaker), anyways if people at one point want to try to figure out the issues with tile by tile (speed, equalisation weirdness) then this could be a good starting point :D.  I'm gonna close the PR because I do feel that the changes that Tile By Tile would need to do would essentially make it room by room for rooms and tile by tile for non rooms, now while this may work in a game set place on 'Earth' (or any planet like as in Rimworld) for our game it's not needed since if you aren't in a room you are probably in space (not probably you are), this means that no temperature is needed (since while space can transfer energy/heat its insignificant).  Regardless this means that while a planet system could benefit from a tile by tile and a room by room combination our game would most likely not significantly benefit (and would just cause problems), so therefore I'm closing this (for now atleast though I doubt we will ever go back to Tile by Tile).

Issue #1729, opened 2016-12-30
The Temperature system needs replaced or overhauled


koosemose:
Displaying the temperature in more understandable terms exposes some other issues preexisting with temperature, most particularly it heats tiles up insanely fast at the heat source but doesn't dissipate that heat to surrounding tiles nearly fast enough, so the heat source will end up at around 1000F, the next tile over at 256F and the next out will be around -256F. Even doing it on a faster timer (everyframe) it doesn't work in any vaguely realistic way (and bogs the game down horribly, as one might expect).

That doesn't count all the things it just does horribly wrong, from trying to patch issues in the past that now the temperature system is just juggling around meaningless numbers, wasting time on an allegedly more realistic temperature simulation, that takes too much time, is really more simulation than we need *if* it worked correctly, and manages to work less realistically than a simpler method such as tracking per room.

I think we need to look into tracking temperature in a different way, perhaps on a room by room basis (of course that brings its own collection of problems to solve, such as how to handle heat transfer through walls), or even just rewriting the whole thing using the same general type of system, just in a way that actually functions more properly.

BraedonWooding:
I was going to post a new issue, but this is a right place :D.  Okay I have a working example of it (it uses a simulation through javascript took maybe 6 hrs to build).  Here is the [link](http://htmlpreview.github.io/?https://github.com/BraedonWooding/HeatingExample/blob/master/HeatExample/empty-example/index.html).  Posted it also on discord but I'll do it here for right jollies.  It has a 3D mode if you so desire (idk) and in 2D it'll show the current values (the redder the square the hotter).  This uses the real world values for heat transfer (well its thermal conductivity) and applies a general formula (crafted by me :D), it seems incredibly accurate (and I'm constantly improving it).  

It uses a few different ways to balance such as it requiring more heat the more heat it has (exponentially) and it loosing heat based on index and starting heat (therefore things that are hotter will cool down quicker, theoretically though haven't tested).  It should work with negative temperatures (cooling things down, but haven't tested).  This is incredibly fast since it doesn't require a complex algorithm and only needs to update whenever there is a new heat input, then once the heat is done it can apply a simple dispersion technique (find your neighbours and average out heat) to average out the heat and disperse it which will eventually fully disperse it across the map (admittedly haven't added this yet, but I did a test and it worked well, just wanted to right it non-hackily).  

[...]

PR #1744, opened 2017-01-06
[WIP] Temperature and Atmosphere overhaul

abackwood:
This completely rewrites the Temperature system from the broken one in place and also refactors the way gasses and atmosphere are handled. The two are an intergrated whole as temperature is a property of gas.

### The issue this fixes

The current temperature system is broken and, in my opinion, overly complex for what it intends to do. We do want temperature to be a part of life support management so a better system has to be conceived of.

#### Temperature
Temperature is tracked on a room level, making the assumption that heat spreads instantly through an enclosed space. Might be the ceilings come with fans to help natural diffusion - not unlikely in a space station.

Instead, the system focuses on the movement of heat through walls (and other enclosing furnitures). The more wall surface there is between rooms and the more the walls are conducive to heat, the faster temperature differences will be equalised between them. Gas pressure is also a factor, where a higher pressure spreads heat faster (though not linearly) and vacuum hardly takes any heat. Some heat loss is always present even in pure vacuum, to account for radiation losses.

#### Atmoshpere
Rather than handling it directly in Room, as is done now, a dedicated AtmosphereComponent is hooked into Room that handles all gas content and the temperature thereof. This has two advantages:
- There is a central place that can do organised calculations on the effects of moving gasses, heating the air and info on the composition of the air in %. It provides intuitive functions to create, destroy and move gasses between these components, rather than having a mix of functions splitting, joining, moving and absorbing that we have now and that do much double work that is error-prone.
- AtmosphereComponent can also be used on furnitures that contain gas. Think of Compressors, Generators and even gas pipes once those are implemented. They can hook straight into the new atmosphere system without much hassle and used established functions.

To still provide some needed higher-level functions, like equalizing rooms, a static AtmosphereUtils class was made that uses the above functions. The move from Room to an Utils class also clarifies that they are convenience methods, not core functions.

Pressure is still a property of Room, since AtmosphereComponent has no knowledge of the space available to the gas. I am open to disagreement on this because I was on the fence about this myself.

heating1
heating2

There is a good chance the function of some furnitures related to gas and atmosphere may have broken in the process. I did my best to avoid this but one or two may have slipped through.

### Technical
WIP

TODO
- [ ] Implement cooler
- [ ] Have another look at when to update the diffusion table
- [ ] Look into optimisation of only recomputing rooms that have changed
- [x] Handle room merging
- [ ] Handle room splitting


koosemose:
I don't know if it's a recent commit or merging with master, but I'm getting 0 functionality when trying to test this, an assortment of DIctionary related errors.

The one in my most recent test gives the following:
```
KeyNotFoundException: The given key was not present in the dictionary.
System.Collections.Generic.Dictionary`2[ProjectPorcupine.Rooms.Room,System.Single].get_Item (ProjectPorcupine.Rooms.Room key) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System.Collections.Generic/Dictionary.cs:290)
Temperature.GetTemperature (Int32 x, Int32 y, Int32 z) (at Assets/Scripts/Models/Area/Temperature.cs:93)
Temperature.GetTemperatureInF (Int32 x, Int32 y, Int32 z) (at Assets/Scripts/Models/Area/Temperature.cs:113)
MouseOverRoomDetails.GetMouseOverString (.Tile tile) (at Assets/Scripts/UI/MouseOverRoomDetails.cs:29)
MouseOver.Update () (at Assets/Scripts/UI/MouseOver.cs:32)
```

Also, if this has some concept equivalent to heat sinks (things that would help with dissipating excess heat), would you be so kind as to implement a simple one as a furniture, to test that aspect. And if it doesn't have a concept of a heat sink, it kind of needs one.

abackwood:
Checked the error out. Thats caused by a bug when a new room is created. I'll get on fixing that.

Heat sinks for me are just negative heaters, so yes they exist. I'll throw something together.

BraedonWooding:
Just want to add mine has no updates, like no time manager updates (just saw you wrote mine has updates) 😀

koosemose:
Seems to equalize perhaps a little too fast, very easy for a single heater to heat an entire base... though *maybe* that's accurate.... there's not incredibly much data available on the heating situation of the space station.... either way it *feels* perhaps a little too easy.

Also need better handling of heat when rooms merge, as is it appears to just use the heat of one of the rooms for the entire room, look how it's handled for gas when rooms join, that should provide guidance (if not I can probably put a PR to your branch to do it).

Of course, it also has the same odd situation with door ways that gas does, there being no temperature in a door way, which may cause difficulties when crew pass through it, if the are to suffer any ill effects from low temp.

BraedonWooding:
You did mention using a cache for tiles.  Now if you did this would solve the second and third problems since you could cache values inside doors and such (and when merging rooms just get the value from the tiles anyway.  The first one is just feeling and perhaps you could make it so they don't diffuse so much?  And that they can only diffuse so far.  Spaceships often require quite complex heating systems since they are often quite cramped (especially the rockets) but in this case a simple heating system like a few heaters should be fine (if it was real life).

abackwood:
Good point, I overlooked merging rooms in the TODOs. The logic shouldn't be terribly hard. The new room gets the sum of both old room's energy.

As for doors, I don't think a temperature "bump map" is the solution for that. It's a bit shooting a fly with a rocket launcher. Making an exception for doors/airlocks - maybe by testing the Passable = Soon trait? - where their tile temperature is the average of the two rooms should be good enough. Characters don't spend their lives in doorways.
Making diffusion through open doors more than through closed ones is kind of a challenge with this system as is. Some changes could be made. I don't know if that's within the scope of this initial PR.

I'll move the heater power down even more, but I'm not overly bothered about balancing right now. It's a simple value in the XML that can be tweaked by anyone. As we see temperature used we'll get a better idea how to balance heaters and coolers. Make them more expensive to build. Make them weaker. Make them be electricity-guzzlers. All that stuff.

koosemose:
Regarding heat in doorways, I *think* there's a hack in place for doors... something about using an adjacent room if a character is in a null room, so it covers both doorways, and if a character ever builds a wall underneath themselves, both situations in which they shouldn't suddenly start freezing , honestly anything so characters don't appear to mysteriously take freezing damage (or whatever being at or near 0k does eventually).

As for changing the heat flow through doors as they open and close, if there is a value on furnitures that affects how heat flows through them, it is public, and changes to it actually affect the system, then doors can handle that their side, just putting it to a value relative to their openness. As long as those actual tools are in place, don't worry about actually implementing it.

koosemose:
Have you had any success in handling the splitting/merging of rooms?

abackwood:
I'm including an overhaul of how gas is handled within rooms. This feels justified as temperature is an integral part of atmosphere and it would be silly to first make a stupid system that works around a chaotic gas system, rather than do both right. This does make the PR a lot bigger however.

koosemose:
Alright, just drop a comment to let me know when you think it's ready for another look-at

BraedonWooding:
Now I do think that maybe making it so it supports multiple worlds would be useful (since it means one less thing to change in the future), it is just removing all World.CurrentWorld (or whatever it's called) from your code (though don't worry for things like room or whatever just the new systems introduced by this PR).  Also I see you have made changes to DevConsoleCoreCommands.cs, now this is my fault (and I'll do a PR about this today since I forgot :P), but it's going to be removed (as in sometime today) since all the commands are already located as a mod under StreamingAssets/CSharp/DevConsoleFunctions (or similar) can you just replicate your changes there (don't copy + paste since they probably won't match)?

crazyfox55:
Currently energy is linearly equal to temperature. I think you should consider making this relationship exponential. This change would keep reduce the chance of a catastrophic temperature. Imagine this graph www.wolframalpha.com/input/?i=y+%3D+x+%5E+(1%2F2) where the vertical axis is temprature and the horizontal axis is energy. This curve would make low temperature disappear with little energy, but would protect against large amounts of energy making ridiculous temperatures. 

The basic formula would be Temprature = Energy ^ (Fraction), but this could be tweaked to provide interesting temperature settling/stage characteristics.

abackwood:
After some discussion in Discord about this, @koosemose, @BraedonWooding and me worked out that there is some truth to what you're suggesting. Apparently, and I did not know this until now, it becomes harder to heat something when it gets hotter because of molecular-level physics. Energy still corresponds linearly to the temperature of a mass though.

Also bear in mind that the more the temperature difference, the higher heat loss is to other rooms and the outside. That in itself works as a balance to avoid extremes. Keeping rooms at a reasonable temperature isn't the job of the mechanics, it's the job of the player. At least in my opinion,

I think it would be better to keep the energy-temperature relation linear and leave the above effect to furniture mechanics. Whether we should make heaters and coolers that complicated is a gameplay discussion, but not for here. Feel free to make an issue on it of course. It's worth discussing.

koosemose:
I would suggest further conversation to be at #1729, as it relates to the replacement temperature system.

BraedonWooding:
I'm happy once I have time to take over and continue this work since I know quite well the temperature system.  Won't be for another few weeks but thought might as well state it.

BraedonWooding:
Merged in a similar variant.

PR #1748, opened 2017-01-10
Use more debug channels

koosemose:
Changes more of the debugging logs to use channels as appropriate. Also updates UnityDebugger to accept objects, rather than only strings, slightly improving usage.


gunthergun:
Seems working, if this needed approvement, I approve 👍🏻

PR #1749, opened 2017-01-10
Adds a seed option to the new game screen

gunthergun:
This PR adds the ability to type in the seed before the game starts. please test for bugs. I style copped the code and there are no violations.

screen shot 2017-01-10 at 21 11 32

### DIFF
--- a/Assets/Scripts/Models/Area/World.cs
+++ b/Assets/Scripts/Models/Area/World.cs
@@ -51,7 +51,7 @@ public class World
         // Creates an empty world.
         SetupWorld(width, height, depth);
         int seed = UnityEngine.Random.Range(0, int.MaxValue);
-        if (SceneController.NewWorldSize != null)
+        if (SceneController.NewWorldSize != Vector3.zero)
         {
             seed = SceneController.Seed;
         }
--- a/Assets/Scripts/Models/Area/World.cs
+++ b/Assets/Scripts/Models/Area/World.cs
@@ -50,8 +50,11 @@ public class World
     {
         // Creates an empty world.
         SetupWorld(width, height, depth);
-        int seed = UnityEngine.Random.Range(0, int.MaxValue);
-        Debug.LogWarning("World Seed: " + seed);
+		int seed = UnityEngine.Random.Range(0, int.MaxValue);
+		if (SceneController.NewWorldSize != null) {
+			seed = SceneController.Seed;
+		}
+		Debug.LogWarning("World Seed: " + seed);
         WorldGenerator.Instance.Generate(this, seed);
         UnityDebugger.Debugger.Log("World", "Generated World");
 
--- a/Assets/Scripts/Models/Area/World.cs
+++ b/Assets/Scripts/Models/Area/World.cs
@@ -50,7 +50,7 @@ public class World
     {
         // Creates an empty world.
         SetupWorld(width, height, depth);
-        int seed = UnityEngine.Random.Range(0, int.MaxValue);
+        int seed = UnityEngine.Random.Range(int.MinValue, int.MaxValue);
         if (SceneController.NewWorldSize != Vector3.zero)
         {
             seed = SceneController.Seed;
--- a/Assets/Scripts/Models/Area/World.cs
+++ b/Assets/Scripts/Models/Area/World.cs
@@ -50,11 +50,13 @@ public class World
     {
         // Creates an empty world.
         SetupWorld(width, height, depth);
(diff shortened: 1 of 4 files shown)


BraedonWooding:
Welcome back btw 😀, good work

dusho:
some screen, please?
we just add note to PR template to post some screens when possible.. it's more explanatory..

gunthergun:
sorry edited it, I didn't know that was required :)

koosemose:
It's not strictly required, but if you're making a UI change or something visual it's.... well, yeah if you're doing a visual change it's required, so people can get an idea of what you've done.

Also you don't need to tell us it stylecopped clean, we run the stylecop in automated testing now, so it'll let us know.

koosemose:
Only two concerns I can see.

The first is the hint text (the text shown when the input box is blank) is depth, I assume you just copied over the depth text box and missed the hint text.

The other is that if you create a new world with a blank seed, it hashes that ending up with 0. I *think* as a user the expected behavior would be to generate a random seed in that case, so perhaps generate a new seed (or *maybe* use the originally created random seed, though I think a new random seed might be best), when the text field equals string.Empty.

Other than that everything looks good and I can't find any way to break it.

gunthergun:
Ok I will change that tomorrow, as I cannot access a computer at the moment :) Yes I just copied that from the depth one, I did not check if there was more to change.

gunthergun:
@koosemose I changed your concerns :)

Assets/Scripts/Models/Area/World.cs
```diff
@@ -51,6 +51,11 @@ public World(int width, int height, int depth)
         // Creates an empty world.
         SetupWorld(width, height, depth);
         int seed = UnityEngine.Random.Range(0, int.MaxValue);
```
bjubes: Should this line be removed? Your setting the seed to a random value only to reassign it. (assuming the != Vector3.zero condition is always met.)

Assets/Scripts/UI/DialogBox/DialogBoxNewGame.cs
```diff
@@ -29,6 +30,9 @@ public override void ShowDialog()
         Width.onEndEdit.AddListener(delegate { VerifyNumericInput(Width); });
         Depth.onEndEdit.AddListener(delegate { VerifyNumericInput(Depth); });
 
+        // Generate Random Seed TODO: make it alphanumerical
+        Seed.text = UnityEngine.Random.Range(0, int.MaxValue).ToString();
```
bjubes: This can be done by hardcoding in every alphanumeric character and using rng to select them or by using 
`Path.GetRandomFileName().Replace(".", "")`

of course keeping it as a integer is fine for now.

Assets/Scripts/Models/Area/World.cs
gunthergun: Yeah you could remove that it is only there so that you can start the game without the main menu scene.

Assets/Scripts/UI/DialogBox/DialogBoxNewGame.cs
gunthergun: I wanted more input on that that is why I added that todo :)

koosemose: Perhaps we could use a word-list, and randomly concatenate two or so of them, though of course that might be more something for another PR

Assets/Scripts/Models/Area/World.cs
koosemose: I would call it necessary as it supports development, allowing us to start in the _World scene

BraedonWooding: Yes please from someone who starts a lot from _World this is very needed :D.  Also it shouldn't be from 0 to MaxValue shouldn't it be `int seed = UnityEngine.Random.Range(int.MinValue, int.MaxValue)` ? shouldn't we include negatives

gunthergun: @BraedonWooding I just used what was there before. Does Random allow negative seed numbers ?

koosemose: it takes an int, and the docs don't *say* anything about limiting the input range, so I *assume* negative is ok.

gunthergun:
@koosemose you should change it as well in the prefab script. There it is also created from 0 to maxint

koosemose:
Alright, Everything looks good, Merged.

PR #1752, opened 2017-01-13
[Fix] Fixed FMOD Memory leak in editor?

BraedonWooding:
For me this seemed to fix the memory leak.  The problem was it wasn't creating the sound system again in the editor (requiring you to quit the unity editor and reopen), this _seems_ to have fixed it.  If it hasn't fixed it for you then just post some basic computer info (version of system running/os.  I.e. MacOS Sierra Version x, or Windows 7 Version y so on)...

I didn't do much other than get the system to close (stopping all handles), then releasing the channels then finishing it with releasing the sound system.

### DIFF
--- a/Assets/Scripts/Models/InputOutput/AudioManager.cs
+++ b/Assets/Scripts/Models/InputOutput/AudioManager.cs
@@ -122,7 +122,13 @@ public class AudioManager
     public static void Destroy()
     {
         SoundSystem.close();
-        master.release();
+
+        // This will also release master, so we don't have to call master.release();
+        foreach (string key in channelGroups.Keys)
+        {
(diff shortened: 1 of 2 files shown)


koosemose:
Since you're explicitly releasing the master channelGroup, perhaps you should release the others as well?

BraedonWooding:
Yeh fair point, it now releases all channelGroups, I removed the call to release master since that is part of the dictionary so it won't get removed twice.

koosemose:
I'm going to assume this works, because it doesn't break anything, but as it can often take hours of starting and stopping to reach the failure state, it's difficult to be absolutely sure it's fixed. If it's not, we can go back in and give it another try. Merged

PR #1754, opened 2017-01-14
Save/Load Seed and Random.State

koosemose:
Simply saves the original world seed to the save file (both for later recovery of the seed, and in case it is needed to facilitate future multiple world generation of some sort), and the random state (meaning anything past world generation that depends on the seed won't be based off a random seed on world load, but instead on the original seed (and where it developed to after multiple Random calls).

Required some slightly awkward workarounds, as JSON.net is unable to automatically serialize Random.State, and none of it's fields are accessible, so it can't be manually serialized, so I use JSONUtility to convert it to/from a Json string feeds to/from Json.net to generate the appropriate object.

If a seed or random state isn't saved in the save file, it will leave everything as is, meaning that it won't work on old save files (but also won't crash, it will just generate a seed as normal), but also meaning that start areas can not have a seed or RandomState, so as to not set those on new world generation (but can also have them included, if someone wanted to create a scenario that always had the same world generation and other randomness).

### DIFF
--- a/Assets/Scripts/Models/Area/World.cs
+++ b/Assets/Scripts/Models/Area/World.cs
@@ -50,14 +50,14 @@ public class World
     {
         // Creates an empty world.
         SetupWorld(width, height, depth);
-        int seed = UnityEngine.Random.Range(int.MinValue, int.MaxValue);
+        Seed = UnityEngine.Random.Range(int.MinValue, int.MaxValue);
         if (SceneController.NewWorldSize != Vector3.zero)
         {
-            seed = SceneController.Seed;
+            Seed = SceneController.Seed;
         }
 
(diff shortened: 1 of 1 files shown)


bjubes:
I tested this and its working. Do you want this merged now or do you want to find an alternative to JSONUtility first?

koosemose:
It should be good for merging, I spent a bit of time trying to find an alternate way to serialize it to no avail. I can only assume either I'm being dumb about it and just can't see it, or they're using some under the hood stuff to allow JSONUtility specifically to serialize it. Either way, working around it with JSONUtility doesn't seem to cause any issues other than just being a little clunky, and if someone else can figure out how to do it without JSONUtility I believe I've got the relevant code separated enough to allow easy replacement.

PR #1755, opened 2017-01-14
Updated fluid furniture (art + setup)

dusho:
Added art for water_generator and water_tank (is kind of in sync with existing oxygen_compressor and smelter hopper idea). Hooked animations and status info.
UPDATE: new art, with different perspective
image

Also added metal_smelter into default save file, so it's possible to smelt mined ores and create stuff without DEV console for now.


koosemose:
Looks good... the perspective is wrong, but it's certainly an improvement (that goes without saying considering the extreme placeholderiness of the previous).

And I was about to do the same regarding the smelter. 

Did you make the water gen actually consume ice now? 

If so, we need a way to halt the processing of a workshop from another component (so that it doesn't consume ice when no water is being used)

dusho:
ye.. I used same perspective as oxygen_compressor, which indeed looks out of place now
That consumption of items when not needed - yes, that needs to be dealt with in generic way in Workshop component, though it's questionable whether fission reactor should also stop using power cells if there is enough power..

koosemose:
I'm not a huge fan of the "hopper" style input and the generator being changed to 2x1, but that's a broader visual/design thing than this PR, really. On a purely visual standpoint, I like the look... which was the point with the original art... that anything is an obvious improvement.

BraedonWooding:
I love it :D!  Looks absolutely amazing much much better than the previous.

koosemose:
Lol, if something managed to look worse than my intentionally bad programmer art, that'd be quite an accomplishment.

Anyways, despite my personal distaste for making the generator 2 wide, and this heavily requiring more modifications to some underlying systems (related to not consuming fluids and inventories when no fluid is needed), I'm merging this.

PR #1757, opened 2017-01-15
[Small] Fix Errors when reseting debug channel filter

bjubes:
### The issue this fixes

Without this work around, sometimes the ChannelSettingSO would be removed but the Editor was still refrencing it without creating a new one. 

### What this PR does
This prevents this by checking that it exists before accessing it when marking settings as dirty. It recalls the Awake function if the ChannelSettingsSO doesn't exist, thus preventing errors when interacting with the Debugger Channel Control when this edge case occurs.

### DIFF
--- a/Assets/Editor/DebuggerChannelControl.cs
+++ b/Assets/Editor/DebuggerChannelControl.cs
@@ -107,6 +107,10 @@ public class DebuggerChannelControl : EditorWindow, IHasCustomMenu
 
         if (dirtySettings)
         {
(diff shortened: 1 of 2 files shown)


koosemose:
Just pointing out that you've got a Stylecop violation.

bjubes:
somehow made a violation despite only adding three lines.... should be fixed

koosemose:
Feels like it's always my smallest commits that have the most stylecop errors.

PR #1759, opened 2017-01-16
Try to connect to power utility in all tiles of furniture

koosemose:
Connects Power consuming/generating furniture to cables in any of its tiles (both when the furniture is placed, and when a utility is placed). 

Also adds a utility function to furniture to support, `Tile[] GetAllTiles()` so that a loop doesn't have to be recoded every time all tiles of a furniture are needed, and to simplify breaking out of looping once a utility is found to connect to.

As the Fluid Network is still Universal at this time, the same logic isn't applied to the Fluid Connection, but should be easy to implement when Fluids through pipes is implemented.

### DIFF
--- a/Assets/Scripts/Models/Buildable/Furniture.cs
+++ b/Assets/Scripts/Models/Buildable/Furniture.cs
@@ -1306,6 +1306,20 @@ public class Furniture : ISelectable, IPrototypable, IContextActionProvider, IBu
         Rotation = rotation;
     }
 
+    public Tile[] GetAllTiles()
+    {
+        Tile[] tiles = new Tile[Height * Width];
+        for (int x = 0; x < Width; x++)
+        {
+            for (int y = 0; y < Height; y++)
+            {
+                tiles[x + (y * Width)] = World.Current.GetTileAt(Tile.X + x, Tile.Y + y, Tile.Z);
+            }
+        }
+
+        return tiles;
+    }
+
     // Make a copy of the current furniture.  Sub-classed should
     // override this Clone() if a different (sub-classed) copy
     // constructor should be run.
--- a/Assets/Scripts/Models/Buildable/Components/PowerConnection.cs
+++ b/Assets/Scripts/Models/Buildable/Components/PowerConnection.cs
@@ -8,6 +8,7 @@
 #endregion
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Xml.Serialization;
 using MoonSharp.Interpreter;
 using Newtonsoft.Json;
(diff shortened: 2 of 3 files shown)


dusho:
hm.. I had something similar when I was trying to solve it
the cases reported by you previously are not an issue anymore? ( #1657 )

koosemose:
I haven't been able to recreate it yet... I'll try recreating the setup that caused the issue with yours, and see if I can get it to happen.

Assets/Scripts/Models/Buildable/Furniture.cs
```diff
@@ -1306,6 +1306,20 @@ public void SetRotation(float rotation)
...
+        {
+            for (int y = 0; y < Height; y++)
+            {
+                tiles[x + (y * Width)] = World.Current.GetTileAt(Tile.X + x, Tile.Y + y, Tile.Z);
+            }
+        }
+
+        return tiles;
+    }
+
```
dusho: in my GetFurnitureTiles ( https://github.com/TeamPorcupine/ProjectPorcupine/pull/1657/files ) I used IEnumerable and yield hoping there will be less allocation of small arrays.. but not actually sure it's better (don't know Unity's mono that well)
 ```
   public static IEnumerable<Tile> GetFurnitureTiles(Tile startTile, Furniture furniture)
    {
        for (int x_off = startTile.X; x_off < startTile.X + furniture.Width; x_off++)
        {
            for (int y_off = startTile.Y; y_off < startTile.Y + furniture.Height; y_off++)
            {
                yield return World.Current.GetTileAt(x_off, y_off, startTile.Z);
            }
        }
    }
```

koosemose: As far as I understand it, using an Ienumerable leads to more object creation, of course my use of linq in PowerConnection may make that a moot point.

And at least how I'm using it (only when a furniture is placed or a utility is placed in a furniture's tiles) it will be such a small difference either way, I think it will be virtually impossible to see a difference either way.

I do recall reading that arrays in foreach loops don't have the excess object creation issues that other objects do in unity's mono, and also that that is how some of the linq statements are implemented, so that *should* be some savings there.

Either way, I think for now either one would be near enough to identical due to the small usage, and if the function starts seeing greater use, it can be revisited to see if one or the other has a more significant difference.

koosemose:
Drat's there they are.... acting as weird as always...

koosemose:
That seems to fix it, it may in fact actually be a bug in master, just only being able to connect to the base tile and the very specific circumstances that trigger this (only seems to trigger when coming from the right), concealed it possibly completely.

dusho:
Tested. Seems to be working fine now.

koosemose:
And I can't generate any oddball disconnections anymore, of course that doesn't absolutely guarantee there's not any weird edge cases, but the all tile connection seems to be working, and at least the previously known issue isn't showing up, so it's certainly an improvement.

PR #1760, opened 2017-01-17
[Fix] Properly return a path to an inventory that can actually be picked up

koosemose:
I made a mistake when implementing RoomPathfinding,  I missed an inventory.CanBePickedUp, so it was pathing to the stockpile, and being told it can't pick it up, and therefore requesting a new path, and being told about the same inventory so they get stuck in a loop of that, so all crew eventually end up stuck in stockpiles, hammering the pathfinding.

This properly adds a check that the inventory can be picked up before pathing to it.

### DIFF
--- a/Assets/Scripts/Models/Job/JobQueue.cs
+++ b/Assets/Scripts/Models/Job/JobQueue.cs
@@ -131,14 +131,14 @@ public class JobQueue
             {
                 if (CharacterCantReachHelper(job, character))
                 {
-                    UnityDebugger.Debugger.LogError("Character could not find a path to the job site.");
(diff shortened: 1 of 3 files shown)


gunthergun:
This should be merged fast 👍🏼

koosemose:
Yeah, that's why I've been trying to draw attention to it on Discord, but it isn't such a game breaking bug that I feel the need to immediately merge it, regardless of outside testing. If you have the time to give it a test and see if things break from it, it would be appreciated and would speed the merge time.

koosemose:
As no one's discovered any issues and it seems to work for me, I'm going ahead and merging this

PR #1761, opened 2017-01-17
use GameTime in inventory Claims

koosemose:
GameTime was created secondarily to in my TimeManger updates, this simply changes the inventory claim system to use GameTime rather than real world time.

### DIFF
--- a/Assets/Scripts/Models/Inventory/Inventory.cs
+++ b/Assets/Scripts/Models/Inventory/Inventory.cs
@@ -100,7 +100,6 @@ public class Inventory : ISelectable, IContextActionProvider
 
     public void Claim(Character character, int amount)
     {
-        // FIXME: The various Claim related functions should most likely track claim time in an in game time increment.
         float requestTime = TimeManager.Instance.GameTime;
         List<InventoryClaim> validClaims = claims.Where(claim => (requestTime - claim.time) < ClaimDuration).ToList();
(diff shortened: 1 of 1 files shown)


BraedonWooding:
Now the only thing I will say is it better to implement gametime in all aspects or in small PRs?  Just if we do it in a bigger one it can be tested maybe better??  Though this is me not fully understanding exactly what things will need game time.

koosemose:
Technically this is just using something that's already there, I slipped GameTime into the spread updates because for some iteration of that I needed it, and saw no reason to remove a tiny thing that would be needed anyways.

Now there may also be a confusion of terminology (especially since my use of terminology has gotten a bit cluttered. GameTime is simply real world time, unpaused, and sped up or slowed down as per time scale, the other thing I've been working on, I've been trying to use the term WorldTime to refer to, which is the time to scale with passage of time in game (so 1 GameTime second = 1 WorldTime second). There may be terms that would better differentiate the two things, but I can't think of any that are any better. 

This is using GameTime because GameTime is on the same time scale as the time in the delta time received from updates from time manager, and inventory claim is set up to function analogous to getting updates from TimeManager, but without actually getting updates on TimeManager (because sending updates to all Inventory would quickly slow things down)

BraedonWooding:
I quite like the terminology and in the end its not so important.  And yeh it should use game time.  Good work with game time by the way :D.

PR #1764, opened 2017-01-25
Varying input/output (power/fluid) for components

dusho:
This allows to define furniture with fluctuating power input - meaning, if there is power shortage, instead of whole grid going down for these machines, machines will work at decreased efficiency (depending on available power)
E.g. here 2 smelters with power requirement of 2 (needed power = 4) being hooked on grid that provides 2 power. Efficiency varies from 50 to 60%. 
image

CanFluctuate flags are currently applied only on Oxygen Generator and Smelter.1

- [x] Fix showing efficiency/electricity description when offline (now showing 100%)
- [x] Do not consume power when not working
- [x] Do not consume inventory when not needed

Changes:
- Requires with params is now RunConditions in components
- Power grid info now shows whether it's connected to grid with at least one producer (as Online/Offline is kind of confusing now with machines that have means to run but are not required)
- IPluggable now has more flags to support 'on demand' producers (fission reactor)

Testcases:
Power on demand
- RTGs are providing power to oxygen generators and pumps
- spawn power cell
 => power cell goes into stockpile (there is enough power in grid)
- remove wires from RTGs that are next to fission reactor (disconnect power from RTGs to grid where reactor is in)
 => power goes offline in pumps and generator
 => power cell is hauled to fission reactor and reactor is producing power
- add power cables back to RTGs (so they are connected again)
 => fission reactor should be idle again, (power is provided from RTGs)
- remove cables from RTGs again
 => fission reactor continues to spend fuel and provide power
 
Grid efficiency
- create RTG, create another smelter
- connect both smelters to one RTG, assign iron smelting, spawn raw_iron
 => see power efficiency fluctuate when smelters are smelting

### DIFF
--- a/Assets/Editor/UnitTests/Models/BuildableComponentTest.cs
+++ b/Assets/Editor/UnitTests/Models/BuildableComponentTest.cs
@@ -108,6 +108,7 @@ public class BuildableComponentTest
 
         Assert.NotNull(deserializedWorkshop);
         Assert.AreEqual("Raw Iron", deserializedWorkshop.PossibleProductions[0].Input[0].ObjectType);
+        Assert.IsNotNull(deserializedWorkshop.RunConditions);
     }
 
     [Test]
@@ -117,7 +118,6 @@ public class BuildableComponentTest
 
         //// serialize
         string workshopJson = SerializeObjectToJson(workshop);
-        StringReader sr = new StringReader(workshopJson);
 
         // if you want to dump file to disk for visual check, uncomment this
         ////File.WriteAllText("Workshop.json", workshopJson);
@@ -161,7 +161,6 @@ public class BuildableComponentTest
 
         // serialize
         string gasConnectionJson = SerializeObjectToJson(gasConnection);
-        StringReader sr = new StringReader(gasConnectionJson);
 
         // if you want to dump file to disk for visual check, uncomment this
         ////File.WriteAllText("GasConnection.json", gasConnectionJson);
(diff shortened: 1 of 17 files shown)


BraedonWooding:
Now the only thing I can say is that if your efficiency is only 50-60% that could be unintuitive and maybe have it so it's replicated to be based of 50-100%?  Unless it's explicit that it doesn't reach full efficiency.

dusho:
@BraedonWooding not sure I understand..
if there is enough power in grid, all devices will work at 100%
if there is not enough power then devices hat are allowed to have varying power input will work at specific efficiency (that will go up and down around some 'stable' point)

crazyfox55:
@dusho I believe he is saying get rid of specific efficiency and just have two modes under powered and fully powered. Underpowered machines work at 50% efficiency. Fully powered machines work at 100% efficiency. 

I also want to point out that this is a bag constraint problem and we should pass control to the user to decide what gets power. In bag constraint problems there are lots and lots of possible solutions but ultimately it comes down to the user's goals at the time.

BraedonWooding:
Nah I actually meant what dusho clarified with said just was confused with OP, so i do think having it based off power supply (% like what dusho said) is the best way since it means a massive power loss isn't going to be absolutely catastrophic (really short term) but will develop into something insane (late term)

koosemose:
@dusho Checking if this is something you expect to be able to return to in the near future, or if I should look into having someone take over this PR.

dusho:
still plan to finish this.. have to change bit more things than I expected and had little time

dusho:
updated, could use some more testing by others..

koosemose:
Heater sprite isn't missing, it's just because it was for some reason using it capitalized (possibly either an oversight or merging mistake). The sprite image files (and consequently the sprite xml) should be using the object's type, which is uncapitalized.

koosemose:
I still have some minor concerns (primarily having to do with the heater sprite issue, and consistency in naming, but as we are moving to a new organization, I'd like to get as much as I can in so PRs don't languish here in the (now old) project.

PR #1765, opened 2017-01-27
World time struct and display

koosemose:
This is a relatively basic display of the world time, semi-arbitrarily settling on using quarters over months/seasons or whatever else might be used, 15 days per quarter, and an epoch of Q1 Day 1 2999 12:00 am. It should be relatively trivial to set a different epoch, though making it changeable would require a bit more work (a separate class to hold the data, since a non-const can't be initialized in a struct), days per quarter would need a similar change to be changeable.

image
Time Display

The most complex portion of this PR is the WorldTime struct itself, which was designed to be similar to C# DateTime, with perhaps excessive ways to access the information, properties for every component, and total time in each relevant time span (i.e. both the day of the quarter of the year, and the total days since epoch), with the components being settable (both by direct access to the property, and a Set[time component] method for each component, and addable with Add[time component] methods. WorldTime has overloads for both arithmetic operators that make sense (addition and subtraction), along with comparisons. And ToString functions (with variants to display the entire time and date, or just one or the other, and convenience methods to directly get time or date).

Additionally WorldTime is synced to GameTime (the WorldTime Property isn't a consistent object, but rather a new object is created when requested, as it is a struct consisting entirely of value types, effects on garbage collection should be no more than if the object were stored and passed. It could also be done by every cycle setting the WorldTime to a value equivalent to GameTime, that may be the better way, but I'm not sure that it would have any significant performance differences, and would have the same caveats on how WorldTime needs to be set), and setting WorldTime (in the way advised in the xml documentation) will set GameTime to an appropriate value.

I've tried to comment this heavily with xml summary comments to make usage clear, along with an in depth xml comment for the WorldTime instance in TimeManager, as to how to properly access and change the time.

This is only marked as WIP as I haven't done a final cleanup, and am not absolutely sure I have cleaned up all old code from earlier versions.

Update 1:
WorldTime should now localize fairly well in the display, and it is now (mostly) formattable, and while I'd like to be able to simply put in a string like one can for DateTime (HH:mm:ss for time) and have the entire thing parsed through, that is entirely too complex for a PR more focused on implementing the time tracking and display, though I will note, that if such a thing is implemented, it would allow us to more automatically display time and date closer to how it is in the chosen language, by finding the matching format strings in DateTIme for the culture, however it is not strictly needed.

The one issue that I have with the functionality isn't fixable in this PR (without going far out of scope), which is to get the correct culture if one isn't provided, however, due to how our language files are named (they don't quite match up with the standard, using an underscore instead of a dash), I can't reliably get the chosen language, and instead have to fallback to the CurrentCulture, though, should we change how they are named to match, it is coded in such a way as to use it properly (falling back to CurrentCulture is implemented as a catch for when the culture argument isn't acceptable).

Note: I *think* I've got everything using the same Format based ToString time/date display, I may have overlooked something from the previous implementation, let me know if any don't fall through to using the ToString Localized format version.


koosemose:
Added image of the time/date display

Assets/StreamingAssets/WorldGen/DefaultStartArea.sav
```diff
@@ -2720,6 +2720,5 @@
       "Type": "power_cable"
     }
   ],
-  "RoomBehaviors": []
-  }
+  "RoomBehaviors": [],
```
bjubes: I assume this is an accidental addition (since the trailing comma would cause a JSON parsing error right?)

Assets/Scripts/Models/Time/WorldTime.cs
```diff
@@ -0,0 +1,508 @@
...
+        return ToString(true, true);
+    }
+
+    /// <summary>
+    /// Returns a string that represents the WorldTime, separated by a linebreak if both time and date are present.
+    /// </summary>
+    /// <returns>The string representing the WorldTime.</returns>
+    /// <param name="time">If set to <c>true</c> include the time in the string.</param>
+    /// <param name="date">If set to <c>true</c> include the date in the string.</param>
+    public string ToString(bool time, bool date)
```
bjubes: just pointing out that this will have to be localized in the future. #1513

Assets/StreamingAssets/WorldGen/DefaultStartArea.sav
koosemose: I accidentally left that in from testing if setting start time in start area save file would work (it does)... but surprisingly doesn't cause a parsing error.

Regardless, I'll see about removing it next time I'm editing the code.

Assets/Scripts/Models/Time/WorldTime.cs
koosemose: Yeah... and I somewhat dread trying to figure out the localization on that since it is a somewhat more complex string.

BraedonWooding: Can you have localization as a format string??  Though that has problems in itself maybe localize parts of it?

koosemose:
WorldTime should now localize fairly well in the display, and it is now (mostly) formattable, and while I'd like to be able to simply put in a string like one can for DateTime (HH:mm:ss for time) and have the entire thing parsed through, that is entirely too complex for a PR more focused on implementing the time tracking and display, though I will note, that if such a thing is implemented, it would allow us to more automatically display time and date closer to how it is in the chosen language, by finding the matching format strings in DateTIme for the culture, however it is not strictly needed.

The one issue that I have with the functionality isn't fixable in this PR (without going far out of scope), which is to get the correct culture if one isn't provided, however, due to how our language files are named (they don't quite match up with the standard, using an underscore instead of a dash), I can't reliably get the chosen language, and instead have to fallback to the CurrentCulture, though, should we change how they are named to match, it is coded in such a way as to use it properly (falling back to CurrentCulture is implemented as a catch for when the culture argument isn't acceptable).

bjubes:
Looks good. This points out a weird problem, that time seems to run a bit while the game loads. When you save at say 9:05 it loads at about 9:07, as that time occurs while the game loads up. perhaps we need to pause the game in a sense during loading.

koosemose:
I suspect that is several things combined.

First is that, of course as far as the game is concerned, it isn't paused, so the time contributes to game time. And to the best of my knowledge, loading is still happening all in one frame, so that first frame, when it finally loads, has all of that time passed to the next frame. And the final aspect is that that first frame that it's unpaused has *any* time on it at all, it's just unpaused that frame, so I'd expect elapsed time to be 0.

Of course, the last will be an inconsequential amount of time passed if the first two are taken care of.

PR #1769, opened 2017-01-30
Performance HUD Changes

BraedonWooding:
The first version of the Performance HUD was using essentially a more hard coded version of the current settings menu modding implementation.  This merely updates Performance HUD to use the same current implementation :D.  

## New Stuff
It just implements a base class called `BaseUIElement` then for each implemenation that needs a highly flexible UI you just can subclass from that to implement your own 'unique' solution, the base class has ways to build some preset things like dropdowns and text.  This PR also implements the PerformanceHUDBaseElement which has an update function attached to it, now currently the actual generating of strings and so on is not particularly efficient and so I'll also improve that in this PR.  Note this system could also be used for any counters or money (and maybe at one point we can just call it 'updatableUIElement').

## Nice Changes
This removes about like 80% of the old classes (though some of them are in the single mod class) and aims to clean up a lot.  Things are cleaner and it's easy to add new options.  Now whether or not an option like this really needs to be modded can be disputed but I think if we ever move to the whole UI being generated this would just fit into that so idk and I was kinda annoyed at my old implementation.

## Checklist
- [x] Implement modding
- [x] Improve string generation using caches and string builders
- [x] Implement the xml file for the modding
- [x] Implement the base stuff (minus the improvements to strings as stated above)
- [x] Fix formatting (currently formatting for it is quite off since I didn't change the root to be more nice with the system)
- [x] Remove weird commits like spaces


koosemose:
Conflicted

koosemose:
Can I assume that since this still has unfilled ticks, it isn't finished until those are filled in, or are those intended for a later PR?

BraedonWooding:
Not finished yet (I will put a WIP) :P

BraedonWooding:
Now this should be merged after the settings options since it changes up one of the options (to allow it to use modding for the HUD) and it would be easier just to merge this in after then the other way around :D.  Also all weird commits should be done (it's saying I've edited ProjectSettings.Asset but it seems to just be a newline thing that doesn't exist since both versions don't have a new line after it and it doesn't change any values).

koosemose:
From looking at the most recent diff, it is a little unclear what the root objects in the list of root objects are. A root object seems like something there should be only 1 of, so it's not immediately clear what exactly rootObjects is referring to. Is there a way you can make this more clear, either with more detail in the summary comment, or perhaps an alternative variable name?

BraedonWooding:
Fair enough yeh I will, in essence it just acts as like a column of elements.  Will do tomorrow ish 😀

koosemose:
Perhaps just a change to rowRoots, or something along those lines... assuming I'm parsing your explanation correctly as it being that all together they make a column and each is an element in that column (and therefore a row)... Of course if I'm wrong and each is a column then of course columnRoots would be more apropos.

BraedonWooding:
Fixed the typing (also found an obsolete prefab :D)

koosemose:
Unsurprisingly, this is heavily conflicted, after the latest merge.

BraedonWooding:
Yep, actually not that heavily conflicted will take like 10-20 mins for me to fix :D, I already did precautions.

koosemose:
Just to clarify, is this solely changes on the code side? Should I expect/look for any changes as far as functionality/appearance goes?

BraedonWooding:
It should generally be better formatted, and will be more clear (and faster).  But mostly it's an internal thing, it doesn't have any functionality impact but as said before maybe a slight appearance.

koosemose:
Alright nothing seems to be breaking. Merged.

PR #1772, opened 2017-02-03
Remove InventoryCommon

### LINKED ISSUE #1735: Justification for InventoryCommon
We've had InventoryCommon in the code for quite some time, and as best I can tell it basically functions as a class to be Prototypes for Inventory, unlike every other prototypable which is its own prototype. I don't see the point of having it separate, it effectively just duplicates some of the code in Inventory, seemingly to no gain.

Can anyone see or come up with a reason that InventoryCommon is useful in some fashion and shouldn't be removed (with Inventory moving to serve as its own prototype)?

koosemose:
As no one presented any reasons that InventoryCommon should remain ( in #1735 ), this removes it and alters Inventory to function as an IPrototypable. 

There are still some odd constructs that I left in as I couldn't be certain if they were simply in place to deal with the odd way Inventories were prototyped through InventoryCommon, or if they were needed in and of themselves, most notably ImportPrototypeSettings. 

Additionally a small static convenience function is added, CreatePrototype to simplify creation of inventory to use as a prototype, used by Furniture that can be crated to add an inventory prototype of its crated form.

As there was absolutely no opposition to the initial proposal to remove InventoryCommon, I intend to have this PR merged in short order.

closes #1735


bjubes:
closes #1735

koosemose:
Everything seems to check out, no unforeseen issues have popped up yet, going ahead and merging this.

PR #1774, opened 2017-02-04
Localize inventory names

koosemose:
Adds localization tags to Inventory, and uses that localization for the SpawnInventory UI.

I'm not sure if there's anywhere else that needs localized inventory names, but if there are, just point them out, and I'll see about using them.

Also, as a matter of cleanliness of UI, I've sorted the Spawn Inventory UI based on category.

Spawn Inventory UI with localized names:
image

One other issue this shows, is that with longer inventory names, some wrap, leading to odd situations, such as making it unclear what an item is (Power Cell Press only shows Power Cell, making it look like Power Cell is listed twice). To solve this I've just changed it to overflow rather than wrap, which I think looks better. 
For comparison:
image
Before Overflow change

image
After Overflow change


BraedonWooding:
Idk if this is relevant to this PR, but I think also a scroll bar could be beneficial, it should be fairly trivial to implement?  It just means it won't take up 80% of the screen height.

koosemose:
Not relevant (I think) but maybe needed.

Also not quite as trivial.. at least for me, as the SpawnInventory UI is built entirely in code, and I'm not the best at dealing with Unity UI in code.

It honestly needs more than just scrollable though, because that still leaves it difficult to navigate. For another PR, a really good solution (in my opinion) would be to have, in addition to scrolling, have foldable categories (i.e. above all the raw resources it says "Raw Resources" or something like that, and it's clickable, when clicked it shows/hides all the things under it, that way it's easier to find a given Inventory. Further sorting could be useful as well (i.e. past being sorted by category, also be sorted alphabetically)

BraedonWooding:
Yes I agree with the relevance tbh, I felt it was still a prefab that you could just place in a scrollbar but I may be wrong (but if not then I can see how it won't be trivial and should be separate).  Currently we aren't splitting those up in xml (I believe) and we shouldn't do it manually.  Actually tbh it should be an xml document that lists all the items so modders can add to it...

koosemose:
Nope, no prefabs in there, pure code. and Inventory *is* categorized in the xml. And as we get more complete gameplay we will end up with more and more inventories, to the point that even scrolling the list is unmanageable. It's really just a simple matter of storing the last inventory listed's category and if it's different, make a header entry with the new category name (of course the folding functionality will be a bit more difficult). 

I'm not really sure what you're suggesting should be in an xml document, as all inventory already is, with their categories.

BraedonWooding:
Oh okay ignore that point since they are categorized one step ahead of me lol.  It's not really storing the last category because if it goes: x, x, y, x then it'll create the x header twice??  However again if it's sorted by the category sure that's a valid solution.  I would suggest maybe have a dictionary leading to a list of items for the inventory list and that way you can just see if key exists if so then add to its dictionary else create dictionary and then add to list.

koosemose:
It's already sorted by category (well I sort it already for display purposes), and has to be sorted for headers to function properly, else things wouldn't be under the proper header. I could in theory do things with dictionaries to ensure unique keys, but I'm already sorting it, and to do it without a sort would require doing it in two passes, which would ultimately end up sorting it just with an additional layer of obfuscation.

BraedonWooding:
Yeh don't worry about dictionaries if sorted (that's why I added that 'afterword' to the situation that may happen).  Coolies tho!  Goodjob

PR #1776, opened 2017-02-04
Small UI Fixes :D

### LINKED ISSUE #471: On high res display UI gets very small

BraedonWooding:
Fixes #471 (gosh that's a very old bug)

Note: this changes not only a lot of prefabs but also changes the .World files this is just to fix things like alignment and anchors :D.

Okay so tldr:
- The PerformanceHUD won't be fixed in this it was however fixed in the PerformanceHUD changes since we mixed up how the UI is displayed :D.
- Fixes the time scale going too far right and overlapping with the overlay
- Fixes the overlay option size staying constant where everything else scales
- Fixes text here and there scaling (mostly prefab changes)
- Fixes the currency display bug (#1771 @bjubes), it now will add the currency display (not it's a prefab) only when world is run
- Amongst some other small things :D 

IF there is any other ui fixes that need to be done please comment below and I'll fix them up :D.


koosemose:
You seem to have a few problems with the time scale still. 

First, You've got an extra copy of the timescale in the scene itself. 

Next, there's an odd little graphical glitch that only shows up on smaller screen sizes:
image
The top is original in same size, bottom is your new version. You can see like a little extra bit of bubble on the left, that almost looks like there's another bar below it (there isn't, I hid the extra time scale control... unless there's a third I overlooked), it shows up both with an unmaximized play area, at 1024x768, and 1280x720. The unmaximized play area isn't a problem, but but the other two are, especially the 1280x720, as that's a pretty reasonable size for someone to play at, especially if they choose to use windowed mode.

The final issue I notice, is that, unlike seemingly all other UI (as far as I can see, it completely scales to aspect ratio, which isn't a huge problem if we could guarantee it's always played at 16:9, but we can't (as far as I know). It wouldn't be so bad if it were a box like other things, but since it's meant to be a certain shape, squishing its size is kind of obvious, and in my opinion looks kind of bad. And since it also scales differently than the text, it also messes up the layout on different aspect ratios.

BraedonWooding:
OOOOh sometimes I'm a little stupid and forget to remove it from the _World scene that I was playing with (though I may have done that but then proceeded to forget to push it), that weird thing is mainly me screwing up the anchors (I again thought I had fixed it ugh was tired when I did the time scale stuff therefore it's a little odd should have better vetted it).  Now I will say that our UI as a whole scales quite bad this wasn't meant to fix that scaling as a whole and text really should be fixed scaling this was mainly to fix it when it gets big.  Now I will go and play around with the squishing thing its just a conflict between making it seeable and useable and making it not too malformed upon resize.

koosemose:
Quick note on procedure, you shouldn't be doing such things in the actual scenes, aside from the potential for little oopses like this, it leads to changes in the Scene file unnecessarily which, if anyone else does the same thing will cause conflicts, which will be a severe pain to resolve if the other person is legitimately changing the scene file. Of course, if you're already making actual changes to the scene file it's not really avoidable, but just to keep in the habit of following proper procedure, you should try to always use a duplicate scene file or an empty one to change prefabs.

On the squishing issue, if you can't find an acceptable solution, we may have to just accept it for the time being, and sometime down the road look into semi manual scaling (have a default that scales to a reasonable default, and allow users to further customize scaling) and do it in a way that keeps images and such at the same aspect ratio (along the lines of Minecraft's GUI scaling options, but perhaps a slider rather than a preset list). It may even be desirable even if you find an acceptable solution, since a GUI scale that works for one person may not work for another even with the same resolution (3 use cases being someone with a setup like mine, where the computer monitor is a TV screen in a media/living room viewing set up where a somewhat larger gui is desirable compared to other users; someone with visual disabilities that needs larger gui to see what's going on clearly; and a power user that has no problem reading smaller text and wants to maximize their visible play area)

BraedonWooding:
Yeh it's only cause I'm already doing changes that I didn't revert them in my commits... Cause to fix some of the things World need a change here or there.  Yeh and I understand about the whole resolving stuff (which was why this is done to be a quick and easy to not really affect other things).

Now I made it better but it still isn't perfect and later on honestly that GUI scaling option idea sounds great, though I do like the idea of a slider (we can control the steps though also so it can maybe be only certain % options idk that kinda seems unintuitive anyways that's for later), so since it's much better I'm happy with it for now.  :D.  I think i've fixed all issues (I'm not facing any issues now and it should be updated with all the time scale changes)

koosemose:
Next chance I get at looking at things I'll check it out again and see if there are any remaining concerns, though depending on what they are it may just be stating what issues I see that still need to be fixed in the future rather than things that need to be done in this PR.

bjubes:
I dont think this actually fixes #471, at least not on the newest iMacs
screenshot

BraedonWooding:
Time scale is the only thing that scales there mmm.  Btw thanks for this :D I don't have the ability to test it on bigger screens (though I guess I could fake super sample).  I will have a look and see how it goes :D, this does fix small screen scaling in most cases.

bjubes:
Since this fixes small screen scaling, do you want to merge this as is and leave the high res display problem for another PR?

BraedonWooding:
Mhmmm I think so since I am working on fixes for bigger screens but that will take a little while longer.

koosemose:
You still have duplicate TimeScales (the code created one, and one in the scene)

And most of the UI doesn't really seem to scale at all even at smaller sizes:
image
640x480

BraedonWooding:
Can you just make sure you have pulled the latest?  Cause it is removed from the scene quite a few commits ago...?  Also the other objects scaling was also fixed a few commits ago.  If they aren't fixed I'll have a look I may have reverted or forgotten to push but I'm pretty sure they're fixed.

koosemose:
Just cleared my pull of the PR and re pulled it, just to be sure I wasn't stuck on the old one for some odd reason. After doing so it's still there, and things aren't scaling as I'd expect them to (aside from time scale, of course). Most recent commit shown on my copy is "Fixed timescale weird things and improved aspect ratio scaling" which matches with the latest commit, and the weird double bubble looking thing is in fact fixed (which I assume is covered by "timescale weird things"

BraedonWooding:
Okay I'll have a look tomorrow, I may have actually forgotten to remove it and gotten distracted.  The scaling thing is weird, I can only see the current time scaling to be particularly wrong (everything else seems to be more about tuning) and that was a recent push that I haven't yet scaled (so recent I hadn't pulled it in to my local I believe) so I'll also fix that and do more fine tuning :).

BraedonWooding:
The scaling for PerformanceHUD by the way, has to kind of be ignored in this one since it's fixed by the PerformanceHUD changes :D.  I removed the time scale from scene (it was still there so my bad... brain fart xD), I was running on a lowest of 840x640 which looks pretty good, and I forgot to check for lower width wise ones.  It seems to be the MenuRight that fails to scale properly but this is a harder issue to fix cause that menu item was built different from the others (different not badly, though personally I don't like it :P), it doesn't scale and it's harder to make it scale properly (without it looking bad generally).  I fixed the date time (just wasn't set up properly with anchors and that meant it went crazy on different resolutions).  And fixed some of the sprites (making them fully expand in their area) which generally makes them look nicer and less cloudy to how big they are.  Just made them a multiple sprite then set the size to the size of the sprite (in the end has 0 impact on the game but has a pretty big visual boost.  So to recap the MenuRight items (the selection indicator and the mouse over info) needs some work, but doing that work would require changes to the actual internal stuff since the text scaling needs to work as a single text source rather than multiple for it to scale nicely.  And that kinda makes it out of scope for this (and people may not be happy with it being a single source, I'm not particularly happy with it either).  So idk it's up to you I can make the single text source changes but that may take a while to get it to format properly or this can be merged (after a final check I would ask) and that can be left for later.

koosemose:
Your changes to time date have made it worse, not only is its relative location inconsistent, but at differing sizes it ends up overlapping selection and tile info, it was anchored to the bottom right for precisely that reason. 

Additionally whatever you've done with it makes the text scale in a way that leads to inconsistent appearance. For example this is it on an actual 1080p:
image

It does the same thing with it overscaling the text at 720p as well.

BraedonWooding:
Actually it wasn't anchored anywhere... if u run the previous version it flies everywhere often right to the top right of the screen because it had 0 anchors...  I merely positioned it to what I felt was right, the text scaling won't be perfect because again it wasn't setup to scale, I quite like how it looks in the picture u attached to be perfectly honest.

koosemose:
Anchored Bottom right in master: 
image

And that picture is just to demonstrate the bad scaling of the text. 

This is its placement on 720p:
image

And this would be it on 640x480
image

BraedonWooding:
I'm sorry but this ain't right anchors...
image
(640x480 on upstream/master, which is why time scale is also off)

Now... honestly, I have around 5-15 mins for PP most days (and nothing more currently for the next few months) so I don't really have the time for a full overhaul of the UI placement (just fixes here and there which was the aim of this PR), now as you can see in the picture you provided above, the date time is actually perfectly positioned, it's just the selection stuff that is off... which is exactly what I said previously that I can fix them but they will take time and do require doing text as a single source (due to scaling) rather than multiple which people may/may not like (personally I kind of dislike)...  Now personally scaling of text is always hard, you can set a reasonably low point like 24 for maximum but that won't scale amazingly on big screens... you can set a much higher point but that'll make it scale too much on small screens.  So as I said before this isn't really here to fix all the text scaling in reality that needs to be a text behaviour that needs to be attached to localisation (which really should be on all text) that then scales a less sensitive variant.

Recap: The positioning of the DateTimeDisplay is broken in master, this does fix it (the text is a different point), the selection stuff (the MenuRight items) is a different matter that I've said before, it will require more internal stuff (which I didn't want to do) if you can fix the stuff (internally or otherwise) then you can go for it :D, and I'm happy to carry out fixes for them but it will take time...  Text scaling is a difficult business and PP really needs it's own text system (or at-least an interface sitting on top) for what it wants to do in some cases.

koosemose:
At best they're both wrong in one way or another. The problem is, as you point out, this is a half-measure, which of course has the problem that it is incredibly hard to judge if the things changed truly fix things.

BraedonWooding:
I'll be honest I like the canvas scaler much more.  Therefore closing this :D

PR #1780, opened 2017-02-10
Add comments to localization

### LINKED ISSUE #1778: Comments in Localization Files
There has been some discussion on putting comments in localization files. Currently they are of a custom format where every line is read and whatever value is to the right of `=` is asigned to the key on the left. 

I think it could be beneficial to add some structure to the files, which currently sit at just under 200 lines, so its easy to find keys other than just CTRL F. However the implementation has to be fast.

I'm suggesting a format where comments must be the only thing on a line (nothing like `function() //comment`) and are denoted by having an initial character that isn't allowed in localization, like `-` .

so a file may look like this:
```
------------------------Power------------------------
power_grid_status_online=Power Grid: <color={0}>Online</color>
power_grid_status_offline=Power Grid: <color={0}>Offline</color>
...
------------------------Furniture------------------------
furn_steel_wall=Steel Wall
furn_steel_wall_desc=A wall made out of steel.
...
```

Thoughts?

bjubes:
### What this PR does
 
Adds the ability to use comments in lang files, delimited by a `#` as the first character on a line. Localization files can now be organized by sections.

I have started to organize en_US.lang with the following headers

- Menu Items
- Furniture
- Utilities
- Inventory
- Job
- Overlay
- Status Messages (mostly dialog prompts and things like `Furniture: {0}`
- Power
- Settings

This also fixes a bug where localization files were being read twice on startup.
Closes #1778


Assets/Scripts/Localization/LocalizationTable.cs
```diff
@@ -265,6 +265,12 @@ private static void LoadLocalizationFile(string path, string localizationCode)
 
                     foreach (string line in lines)
                     {
+                        if (line.Length < 1 || line[0] == '#')
+                        {
+                            Debug.Log("comment");
```
koosemose: I assume this is just leftover from testing?

bjubes: yep, ill remove that

koosemose:
And everything should be good after that conflict resolution, merged.

PR #1782, opened 2017-02-24
[Small] Fix Warnings

BraedonWooding:
This will just deal with the warnings we currently have, I have either commented them out, removed the assigned variable for a function call or remove the line (mostly for useless variables including some lists that were retrieved during pathfinding but never used so that could speed up and reduce GC).  

I did a few changes WorldTime just to add an equals function and a get hash code, the equals checks instances, then checks if they have the same seconds (if both are of type WorldTime), and get hash code just returns the seconds.getHashCode() call.  Other than that I commented out offsets that were allocated (using a random range) but were never used, and made sure that it was just American Spelling (to fit the common standards, and most of the changes were in comments made by me :P, but some of them were variables such as in needs which seems quite old).  

Anyways pretty small changes :D.


Assets/Scripts/Models/Time/WorldTime.cs
```diff
@@ -294,13 +315,13 @@ public int Year
 
     #region IFormattable implementation
 
-    string IFormattable.ToString(string format, IFormatProvider provider = null)
+    string IFormattable.ToString(string format, IFormatProvider provider)
```
koosemose: This was here intentionally so that it can be used without having to pass a provider (even if it is null), to allow for cleaner usage (imo intentionally passing nulls looks bad, and is pointless when it can be done otherwise like this)... 

Unless, of course, IFormattable does this on its own behind the scenes...

BraedonWooding: As per the C# Language Specifications in 10.6.1:
"If optional parameters occur in an implementing partial method declaration (§10.2.7) , an explicit interface member implementation (§13.4.1) or in a single-parameter indexer declaration (§10.9) the compiler should give a warning, since these members can never be invoked in a way that permits arguments to be omitted."

Basically you aren't meant to add parameter options (like ref, optionals, params...) to interfaces if they don't have them, because it results in cases where you can't actually omit the parameter.  However, as per the IFormattable.ToString [method](https://msdn.microsoft.com/en-us/library/system.iformattable.tostring(v=vs.110).aspx) there is another function you can override or whatever that just has the format (that is `string IFormattable.ToString(string format)`) and normally it just calls the provider one with `CultureInfo.CurrentCulture` or null).  I can add this if you want?  Hopefully this however clears up why you shouldn't have default parameters on interface implementations :D.

koosemose: Yes, if you'd be so kind as to drop in the non provider one. Just have it set to null, so that the picking the right culture thing will work right (well, will work if/when we rename the language files).

Also, in case you didn't notice, you have a Stylecop failure.

BraedonWooding: Yeh I forgot to run a test I think :P.  That's why we have checks

koosemose:
Purely from a visual overview the code changes look fine. I'll still, of course need test things, to make sure there are no unintended side-effects, or unobvious side-effects that were intended and lost.

However, without exceptional reason (code that should be doing something but for some reason isn't or isn't working as intended, and needs reimplemented, is all I can think of off-hand), code should be removed rather than commented out

BraedonWooding:
Yep understand about the testing :D.  About the next bit yeh already did another commit earlier for that reason the only things I left where things that obviously at one point may have an effect (like in settings the path to the streaming assets settings file) but doesn't currently.

Assets/Scripts/Models/Time/WorldTime.cs
```diff
@@ -292,15 +292,41 @@ public int Year
         return Math.Abs(time1.Seconds - time2.Seconds) > float.Epsilon;
     }
 
+    public override bool Equals(object obj)
+    {
+        if (obj.Equals(this))
+        {
+            return true;
+        }
+        else if (obj is WorldTime && ((WorldTime)obj) == this)
```
bjubes: technically you just need two ifs statements followed by `return false`. the elses are redundant since if it had been satisfied it would have already returned.

BraedonWooding:
Yeh sure, just did the else to be explicit, can remove if you want?

Assets/Scripts/Models/Time/WorldTime.cs
BraedonWooding: Yes sure, it was more a redundancy thing to be purely explicit but I can change it if it really bugs you :P.

koosemose:
Conflicted

BraedonWooding:
Ooops conflicts were fixed ages ago forgot to remove [conflicts] from title

koosemose:
Hah! ... well that's good that it's fixed nonetheless, I'll have to give it another quick go to make sure nothing untoward happened from conflict resolution ... later, as I've been bashing my head against a Unity bug for far too long and need to stop looking at code and Unity in general for a bit.

koosemose:
Nothing seems to spontaneously combust, Merged.

PR #1785, opened 2017-03-06
[Small] Make Settings accessible to lua

koosemose:
Previously Settings was inaccessible to Lua, and after being made accessible the standard GetSetting could not be used (some combination of generic function and an `out` parameter did not play well with Lua).

So this makes Settings accessible to Lua, and adds a simple GetSetting that just returns the base string, which lua can handle just fine as a number if that's what it is, and should be able to handle as a boolean (though I haven't tested that side heavily). Though simple this should work quite well for Lua.

### DIFF
--- a/Assets/Scripts/Models/Functions/LuaFunctions.cs
+++ b/Assets/Scripts/Models/Functions/LuaFunctions.cs
@@ -43,6 +43,7 @@ public class LuaFunctions : IFunctions
         RegisterGlobal(typeof(Scheduler.ScheduledEvent));
         RegisterGlobal(typeof(ProjectPorcupine.Jobs.RequestedItem));
         RegisterGlobal(typeof(DeveloperConsole.DevConsole));
+        RegisterGlobal(typeof(Settings));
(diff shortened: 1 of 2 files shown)


BraedonWooding:
Now the only change I would say would maybe call it "GetSettingAsString" or "GetSettingString"?  This way it's more descriptive of what it actually does.

koosemose:
While I can see the idea behind it, I think it is overly verbose for minimal gain, since String is its storage form, quick info will show the return type, and in the main (maybe only) place where it won't (lua) it doesn't matter, as it can be treated as a number if it is a number-like string. 

Now if there were a Setting object (an individual Setting as opposed to the Settings as a whole) I could see the sense in it, as a unique object that could be treated in different ways. but I don't think that would add anything of significance to settings (though perhaps some more complex systems might be able to be built using such a thing, but that, of course, is a discussion for elsewhere.

But I feel AsString is either needless typing of the function name, or implies a conversion of some sort when there isn't.

PR #1786, opened 2017-03-06
Custom Parameter support for my UI System

BraedonWooding:
**Note: only touches the code side shouldn't have any influence on the visual appearance**

This is not only for settings menu options, this is for all systems that use my UI system (currently just settings menu and performance HUD but hopefully it'll expand :P), it now uses a more code based approach.  This is done through adding support for custom parameters, you just define them like every other parameter in xml (like furniture ones it uses same class), but then you can access those paramters and use them to implement custom settings.  

Note this is all done soft coded, even the 'generic' methods are staying as a core mod they aren't hardcoded into the game, I wanted to keep it this way not only as a good example of how to code for the settings menu but also to push the game as mostly a core mod (an idea that has been held for a while by the community).

Currently only sliders implement it, you can have min/max values and toggle whole numbers on/off.  But also the sound slider has it for the sound channel which removed around 5 different sound classes which only difference was what channel they wrote to.  I will work to add more options :D.  And to move this to performanceHUD (which is as easy as passing to the object a parameter object, which is as easy as doing `Parameter.ReadXml(...)`).  Other than that it should also be a little faster since I fixed up some internal stuff :D.  

Also if you're concerned about the difficulty of implementing a custom option don't be it's super simple here is a single liner for the minimum value (this code is located in the generic slider): ` float minValue = this.parameterData.ContainsKey("MinimumValue") ? this.parameterData["MinimumValue"].ToFloat() : 0;`.  This should make in most cases extending a class not necessary just for small changes, and I'm thinking about having a dirty flag system that could easy be implemented as a generic parameter (aka have a big enum and then each system can just check to see if the flags they want have changed).

Left to do:
- [x] Add more parameter options
- [x] Implement options for performance hud.

All Options:
FOR SETTINGS MENU
- Toggles:
   - Type, can be: `Switch`, `Toggle`, `CircleRadio`, `SquareRadio`
- InputField:
   - Verification, can be: `PositiveInteger`, `Integer`, `Alphanumeric`, `Decimal`, `Pin`, `Password`, `Name`, `EmailAddress`, `Autocorrected`.
- Slider:
   - MinimumValue, can be any float
   - MaximumValue, can be any float
   - WholeNumbers, can be true or false
- Sound Slider (also includes all slider options):
   - SoundChannel, can be any valid sound channel (music, gameSounds, alerts, UI, master)

FOR PERFORMANCE HUD
- FPSPerformanceComponent
   - MeasurePeriod, can be any float
   - DisplayText, can be any text should be in format "Text: "
   - DisplayColor, can be true or false

If you want any more options ask for them in comments (or ping me on discord) :D.

### DIFF
--- a/Assets/Scripts/Models/Functions/Parameter.cs
+++ b/Assets/Scripts/Models/Functions/Parameter.cs
@@ -115,7 +115,7 @@ public class Parameter
             string k = subReader.GetAttribute("name");
 
             // Sometimes the look will get stuck on a Whitespace or an EndElement and error, so continue to next loop if we encounter one
-            if (subReader.NodeType == XmlNodeType.Whitespace || subReader.NodeType == XmlNodeType.EndElement)
+            if (subReader.NodeType == XmlNodeType.Whitespace || subReader.NodeType == XmlNodeType.EndElement || string.IsNullOrEmpty(k))
             {
                 continue;
             }
--- a/Assets/Scripts/Models/Settings/Settings.cs
+++ b/Assets/Scripts/Models/Settings/Settings.cs
@@ -99,7 +99,7 @@ public static class Settings
             }
             catch (Exception exception)
             {
-                UnityDebugger.Debugger.LogErrorFormat("Settings", "Exception {0} whyle trying to convert {1} to type {2}", exception.Message, value, typeof(T));
+                UnityDebugger.Debugger.LogErrorFormat("Settings", "Exception {0} while trying to convert {1} to type {2}", exception.Message, value, typeof(T));
                 return false;
             }
         }
--- a/Assets/Scripts/Models/Time/TimeManager.cs
+++ b/Assets/Scripts/Models/Time/TimeManager.cs
@@ -115,9 +115,9 @@ public class TimeManager
     /// <para>Example: <code>WorldTime = WorldTime.SetHour(8).SetMinute(0)</code> will properly change the time to 8:00, leaving everything else the same.</para>
     /// </summary>
     /// <value>The world time.</value>
-    public WorldTime WorldTime 
+    public WorldTime WorldTime
     {
-        get 
+        get
         {
             return new WorldTime(GameTime * RealTimeToWorldTimeFactor);
         }
@@ -126,7 +126,7 @@ public class TimeManager
         {
             GameTime = value.Seconds / RealTimeToWorldTimeFactor;
         }
-    } 
+    }
 
     // Current position in that array.
     // Public so TimeScaleUpdater can easily get a position appropriate to an image.
@@ -173,7 +173,7 @@ public class TimeManager
     /// </summary>
     /// <param name="time">Time since last frame.</param>
     public void Update(float time)
-    {   
+    {
         float deltaTime = time * TimeScale;
 
         // Systems that update every frame.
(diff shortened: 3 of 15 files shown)


koosemose:
The one *potential* issue with a "dirty" flag I see is if there is a possibility of multiple systems needing to access the same setting. 

To explain I'll provide an example. Let's say we implement a difficulty system. Let's further say that all it does is increase the rate of events and tend them towards negative events. We for, whatever reason don't touch the dirty flag, (perhaps as it is only an occasional check it is deemed not worth the effort). Modder A comes along and decides that harder difficulties should mean crew die more easily (fewer HP), and makes a simple mod that just regularly checks for the dirty flag, and if it is dirty applies a percentage change to HP. All well and good, no conflicts (assuming they clear the dirty flag, and if they don't they'll have their own issues). Now modder B comes along and decides enemies should be tougher at higher difficulties, and similarly when the setting is dirty to apply a percentage change to enemy HP (he's not very creative and may have stolen the idea from modder A). But now there's an issue when Player 1 installs both mods, if they both clear the flag then only one of them will actually discover the dirty flag.

Obviously the example is a bit contrived, and the specific examples I came up with may be able to do it without worrying about dirty flags, but the point still stands, at least on settings implemented by us that modders may also want to attach things to, the flag can only be seen and cleared once, everything after will see a clean flag, and if that's the only way to attach behaviors reacting to a setting change then it will only be useful for Settings implemented by them and only meant to be used by their mod.

Is there another possible way we can implement a way for mods (or our own core) to react to settings changes? Possibly some kind of event system? Of course not for this PR, since this would almost certainly be much more complex than a dirty flag. I can see a way for it to work, for stuff in actual core, we can basically just use actual C# events or actions or whatever it is in C#, and if anything that's modded can be hooked up to the event (I assume the actual C# event will have to be a general event for any setting changed with an object being passed that tells the setting), it can then pass it on to an EventAction (if it does something like `EventActions.Trigger("Setting_Changed_" + <setting key>);` then the modded code will only have to respond to its actual desired setting being changed (since modded code, particularly C# is so much slower, we definitely don't want it responding to any setting change... though technically, I suppose all settings will change at once)... sorry wondered off theorycrafting here...

BraedonWooding:
Sorry when I mean dirty flags I meant event triggers :P, I explained it poorly (just kind of threw it into there since it's not really meant for anything with this PR, but it had a 'slight relevance' so I wanted to mention).  What I was thinking was to steal from other games a little and do an 'InvalidateCache' event, then you would go in and see if your flag has changed (you don't undirty it, it just passes an int which is bitwise aka `BigEnum.SettingsChanged & BigEnum.OtherThingChanged`) which can mean you can check if your flag has changed super fast (literally int checking) and quit out if not.  Then we could be smart with invalidating the cache (aka only do it once upon settings save, and try to do it rarely) and then have very little performance impact and a pretty huge boost to the whole system I think.  Again off a bit to theorycrafting but yeh.

koosemose:
Yeah that was basically the exact sort of thing I was talking about any way... we were on the same page and just didn't realize it. 

And the only reason I suggest having modded stuff flow through to individual EventActions is because the simple act of calling a modded function has more overhead than doing so in core C#, particularly Lua, but still to a degree in modded C#, so it would be a performance hit to call all mod code that needs to respond to a settings change (and lua in particular can't easily respond to C# events (i.e. I think a helper function could be created to do it, but that just adds more overhead, because it would be more layers of going between C# and Lua, and as I understand Moonsharp transitions both ways can have a bit of extra overhead)).

But perhaps we should save further theorycrafting for Discord, a specific Issue, or another PR, because I suspect such a system would be fairly far reaching and vastly balloon this PR.

koosemose:
Nothing seems to break, and no obvious performance degredations... and since this is code side only, that should do it. Merged.

PR #1790, opened 2017-03-30
[Small] Settings menu fix

BraedonWooding:
- Fixes the UI Scale localisation (there wasn't a key of it in the localisation file so I added it)
- Fixes the scaling of the UI menu by changing anchors everything should work nicely now


koosemose:
Haven't had any opportunity to look at this yet, and I'm assuming it'll be a bigger project than this PR, but do you have any thoughts/ideas on how to handle the oversizing of the settings UI on larger scales (since it occupies so much of the screen it doesn't take that large of a UI scale to make it flow off of the screen).

BraedonWooding:
For me this completely fixes all of that, and it scales with the UI scale completely, now if you have a small screen and make the UI scale really large it will probably overscale but that's not really a problem that we should concern ourselves with?  Since if you choose a reasonable size or even more than reasonable for your screen it'll be great.

koosemose:
It seems you (directly or indirectly) adjusted the default size, so this seems to be less of an issue.

However, I still worry about it some because, without live updating of the scale (and maybe even with) it may be possible in some configurations to unknowingly choose a size that is large enough that the UI scale setting can no longer be accessed to be reverted. Perhaps at least the ability to set some property on a setting so that it asks for confirmation if any settings with that property are changed (similar to the prompt when changing resolution in windows).

And there's also the matter of deciding what is really reasonable... as an example on a 720p display 2x scaling seems reasonable for someone with vision problems... Alternatively it should be possible to use the size of the Settings menu (or some other metric) to set a sliding limit to maximum scale (i.e. calculate what size would overflow and prevent it from going that large).

BraedonWooding:
You can always go into the settings to change it, and even on 2x I could still change the UI scale (just barely), I'll adjust it a little by making it about ~15% smaller, which should basically sit it perfectly on 2x for a 720p (I think I'll play around)

BraedonWooding:
Now it has a limit, it should work much nicer now.

koosemose:
Cool. That should work. I'll poke at this more later... see if I can break anything or find any other suggestions to annoy you with ;)

koosemose:
I haven't tested this exhaustively, but as we are moving to a new organization, I'd like to get as much as I can in so PRs don't languish here in the (now old) project.

PR #1791, opened 2017-03-30
[Small] Remove the migrating notice

BraedonWooding:
Self explanatory just removes the migrating notice from the readme.


koosemose:
Merged.

Though I've just begun testing running things on 5.5 again, and it seems to be not having the issues I experienced before, so after I feel comfortable with it I may look into finally migrating. And after that I'll start looking into 5.6, particularly as 5.6 has some features that I think will simplify some of our Sprite stuff (particularly Axial sorting)

Issue #1792, opened 2017-03-31
[Feedback] LUA for UI


BraedonWooding:
I've been scanning through and currently the only feature that my UI system lacks comparative to the modded UI system is LUA.  Now upon inspection it seems that all the LUA that really can be done is just setup stuff that my UI already handles so it's really just on action perform this...  

Which can actually be implemented through custom parameters, something like 'onActionLUAHandler' which will be run alongside the C# one (and if you just are using LUA it'll use the default empty C# one).  

Now this won't give you complete control over the object and you can basically only do normal LUA stuff and won't be as powerful as C# but will 100% work for something like go do this or set this flag.  So a good example would be a modded popup 'magic' window (something not done by base game idk can be anything), you would most likely need to use C# for this since you are adding a ton of custom logic, but if you wanted to do something like a popup slider that changes the time for a ship to spawn then using LUA would be preferable and easy.

This is basically how it sits, for any simple logic LUA can be used (well by that I mean I can add a quick custom parameter header that enables this logic) but if you want anything past 'basic logic' you need to use C#.  This is how it sits for almost every other system and honestly in most cases I think you could use the custom parameter system to even make it so you never even have to use any coding!

Is this acceptable?  For the sake of requiring C# for complex logic we get speed, and flexibility allowing you to add any element also it's a lot simpler to implement then modded dialogs and scales nicer, since it uses layouts and not absolute positions.  Also it's fully soft coded even custom parameter use is soft coded so yah :P.

koosemose:
As long as the object itself is passed to Lua somewhere along the line and made available to Lua's Userdata, it can do anything C# can, except for creating new classes of course.

Since I'm more familiar and comfortable with Lua than you (from discussions on discord) I'll probably be poking my nose into it a lot and making a lot of suggestions. I'll also probably attempt to make some UIs from LUA since that'll give me a lot better idea of what it can do, and what it may need added.

BraedonWooding:
Yep completely fair, I'll add the support later today and we'll see.

PR #1793, opened 2017-04-01
LUA Support for UI

BraedonWooding:
Closes https://github.com/TeamPorcupine/ProjectPorcupine/issues/1792

This adds support for LUA, what it does is add events (which moonsharp can access and add/remove from/to).  The link for how to is [here](http://www.moonsharp.org/objects.html#events).  This has no effect on current code.  

To add lua functions to the events is really easy, you just have to implement an InitializeLUA function by adding the key `LUAInitializeFunction` and the value being your function name as a custom parameter.  It will then call it upon initial load and you can assign yourself to the function just by doing `function.add(myFunction)` you can also fire it just as normal.

This will however not give you access to the initialize object command, this is mainly just to keep the whole thing smooth and simple since already each event has to be done on implementation and not base due to how we implement mods.  So it really is only build for small changes and you can't create custom objects, a good example would be that if you wanted to implement a button to toggle fullscreen you would use LUA where as you would use C# for a dropdown of all resolutions, or something that doesn't fit nicely into the generics.

There is an example in this build for the DevMode.  Note: you have to use `Settings.GetSetting(string key) -> string` to get the setting, the whole caching is really only for C# since we can do a full object implementation rather than a partial function one as in LUA.


koosemose:
Conflicted

BraedonWooding:
I'll get this unconflicted right now!

koosemose:
Remember your title changes ;P

koosemose:
As with the others, merging this without full testing for languish prevention reasons.

